PR #3, opened 2019-11-20
Fixed lifetime issues

aaronalbers:
- Channels are not longer immortal objects
- Fixed (or mitigated) crashes on cleanup

### DIFF
--- a/src/peerconnection.cpp
+++ b/src/peerconnection.cpp
@@ -30,6 +30,7 @@ using namespace std::placeholders;
 
 using std::function;
 using std::shared_ptr;
+using std::weak_ptr;
 
 PeerConnection::PeerConnection() : PeerConnection(Configuration()) {}
 
@@ -97,7 +98,7 @@ shared_ptr<DataChannel> PeerConnection::createDataChannel(const string &label,
 }
 
 void PeerConnection::onDataChannel(
-    std::function<void(std::shared_ptr<DataChannel> dataChannel)> callback) {
+    std::function<void(shared_ptr<DataChannel> dataChannel)> callback) {
 	mDataChannelCallback = callback;
 }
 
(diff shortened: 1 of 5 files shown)


src/peerconnection.cpp
```diff
@@ -120,24 +121,30 @@ void PeerConnection::onGatheringStateChange(std::function<void(GatheringState st
...
+        
 		    switch (state) {
 		    case IceTransport::State::Connecting:
 			    changeState(State::Connecting);
 			    break;
 		    case IceTransport::State::Failed:
 			    changeState(State::Failed);
 			    break;
-		    case IceTransport::State::Ready:
+		    case IceTransport::State::Connected: // At least one working candidate pair
```
paullouisageneau: Good point, but are you sure it works? I chose to wait for ready here because the documentation for `nice_agent_send` explicitly states:
> Component state MUST be NICE_COMPONENT_STATE_READY, or as a special case, in any state if component was in READY state before and was then restarted.

https://nice.freedesktop.org/libnice/NiceAgent.html#nice_agent_send

aaronalbers: Thanks for bringing that up. I will look into it further as I was getting intermittent results when waiting for READY. One day everything would be working fine and then the next (without changes) the state would fail to be reached. I of course was only using the tests to check for functionality but it appeared to be working.

I am wondering how libnice is supposed to handle candidate trickling. I will look into the docs and RFCs but if you know offhand just let me know. For now we can change it back to READY and address that issue in another PR. I will update this PR shortly.

aaronalbers: @paullouisageneau Ok back to using `Ready`. However I was still unable to make it work in this state. So I will be looking into that either today or tomorrow.

paullouisageneau: You are right, I also wonder how you are supposed to do candidate trickling with this requirement, to have the same behavior as on web.
After a quick look at libnice's code, it seems the documentation is off, since the function only checks for a selected pair internally: https://github.com/libnice/libnice/blob/d6087ae5e0c1de2555f64d792c144d49bd816920/agent/agent.c#L4924
Maybe it means something else, like in the general case there is no guarantee if you call it before the candidate selection is final.

PR #14, opened 2019-12-16
TURN Server Support

murat-dogan:
Please check tests for a sample call.

### DIFF
--- a/include/rtc/configuration.hpp
+++ b/include/rtc/configuration.hpp
@@ -27,9 +27,10 @@
 namespace rtc {
 
 struct IceServer {
-	enum class Type { STUN, TURN };
+	enum class Type { Stun, Turn };
 
-	enum class RelayType { RELAY_TYPE_TURN_UDP, RELAY_TYPE_TURN_TCP, RELAY_TYPE_TURN_TLS };
+	// Don' Change It! It should be same order as enum NiceRelayType
+	enum class RelayType { TurnUdp, TurnTcp, TurnTls };
 
 	IceServer(const string &host_);
 	IceServer(const string &hostname_, uint16_t port_);
--- a/include/rtc/configuration.hpp
+++ b/include/rtc/configuration.hpp
@@ -27,12 +27,22 @@
 namespace rtc {
 
 struct IceServer {
+	enum class Type { STUN, TURN };
+
+	enum class RelayType { RELAY_TYPE_TURN_UDP, RELAY_TYPE_TURN_TCP, RELAY_TYPE_TURN_TLS };
+
 	IceServer(const string &host_);
 	IceServer(const string &hostname_, uint16_t port_);
 	IceServer(const string &hostname_, const string &service_);
+	IceServer(const string &hostname_, const string &service_, Type type_, string username_,
+	          string password_, RelayType relayType_);
 
 	string hostname;
 	string service;
+	Type type;
+	string username;
+	string password;
+	RelayType relayType;
 };
 
 struct Configuration {
--- a/src/icetransport.cpp
+++ b/src/icetransport.cpp
@@ -76,7 +76,7 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
 	for (auto &server : servers) {
 		if (server.hostname.empty())
 			continue;
-		if (server.type == IceServer::Type::TURN)
+		if (server.type == IceServer::Type::Turn)
 			continue;
 		if (server.service.empty())
 			server.service = "3478"; // STUN UDP port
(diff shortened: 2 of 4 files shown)


murat-dogan:
If you need a TURN server to test, you can use xirsys.net.
They have a 500MB / month free account.

src/icetransport.cpp
```diff
@@ -128,6 +129,44 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
...
+	for (auto &server : servers) {
+		if (server.hostname.empty())
+			continue;
+		if (server.type == IceServer::Type::STUN)
+			continue;
+		if (server.service.empty())
+			server.service = "3478"; // TURN UDP port
+
+		struct addrinfo hints = {};
+		hints.ai_family = AF_INET; // IPv4
```
paullouisageneau: You should set the family to `AF_UNSPEC` to also handle IPv6.

src/icetransport.cpp
```diff
@@ -128,6 +129,44 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
...
+		hints.ai_socktype = SOCK_DGRAM;
+		hints.ai_protocol = IPPROTO_UDP;
```
paullouisageneau: No big deal, `socktype` and `protocol` should be adapted according to the relay type.

src/icetransport.cpp
```diff
@@ -128,6 +129,44 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
...
+		hints.ai_flags = AI_ADDRCONFIG;
+		struct addrinfo *result = nullptr;
+		if (getaddrinfo(server.hostname.c_str(), server.service.c_str(), &hints, &result) != 0)
+			continue;
+
+		for (auto p = result; p; p = p->ai_next) {
+			if (p->ai_family == AF_INET) {
```
paullouisageneau: You should also allow `AF_INET6` to allow IPv6.

include/rtc/configuration.hpp
```diff
@@ -27,12 +27,22 @@
 namespace rtc {
 
 struct IceServer {
+	enum class Type { STUN, TURN };
```
paullouisageneau: Since I use a format like `enum class Type { Stun, Turn }` everywhere else, could you change it for the sake of consistency ?

include/rtc/configuration.hpp
```diff
@@ -27,12 +27,22 @@
...
+
+	enum class RelayType { RELAY_TYPE_TURN_UDP, RELAY_TYPE_TURN_TCP, RELAY_TYPE_TURN_TLS };
```
paullouisageneau: Same here, also since you cast it directly to pass it to libnice, I think it'd be safer to specify the values explicitely, like:
```
enum class RelayType { 
    TurnUdp = NICE_RELAY_TYPE_TURN_UDP, 
    TurnTcp = NICE_RELAY_TYPE_TURN_TCP, 
    TurnTls = NICE_RELAY_TYPE_TURN_TLS
};
```

src/icetransport.cpp
```diff
@@ -128,6 +129,44 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
...
+				char nodebuffer[MAX_NUMERICNODE_LEN];
+				char servbuffer[MAX_NUMERICSERV_LEN];
+				if (getnameinfo(p->ai_addr, p->ai_addrlen, nodebuffer, MAX_NUMERICNODE_LEN,
+
+				                servbuffer, MAX_NUMERICNODE_LEN,
+				                NI_NUMERICHOST | NI_NUMERICSERV) == 0) {
+					nice_agent_set_relay_info(mNiceAgent.get(), mStreamId, 1, nodebuffer,
+					                          std::stoul(servbuffer), server.username.c_str(),
+					                          server.password.c_str(),
+					                          (NiceRelayType)server.relayType);
```
paullouisageneau: I try to use `static_cast` rather than C-style cast, but this works.

src/icetransport.cpp
murat-dogan: OK

murat-dogan: Actually not OK.
We are resolving IP address of the server right? Why do we need IPV6 address?

src/icetransport.cpp
murat-dogan: Normally Relay servers can communicate with both UDP & TCP.
But you are right we should add it.

include/rtc/configuration.hpp
murat-dogan: Sure

include/rtc/configuration.hpp
murat-dogan: OK

src/icetransport.cpp
murat-dogan: OK

src/icetransport.cpp
murat-dogan: Let's start with IPV4 and see what's coming.

murat-dogan:
If it is OK please merge it.
Next I will add option for ice-tcp.

src/icetransport.cpp
paullouisageneau: I mean that if the server has an IPv6 you should also add it to the relays. However this can be done in a second time.

PR #17, opened 2019-12-17
Fix logic error: If we pass Relay Type then it is a TURN Server

murat-dogan:


include/rtc/configuration.hpp
```diff
@@ -35,8 +35,9 @@ struct IceServer {
 	IceServer(const string &host_);
 	IceServer(const string &hostname_, uint16_t port_);
 	IceServer(const string &hostname_, const string &service_);
-	IceServer(const string &hostname_, const string &service_, Type type_, string username_,
-	          string password_, RelayType relayType_);
+	IceServer(const string &hostname_, const string &service_, string username_, string password_);
+	IceServer(const string &hostname_, const string &service_, string username_, string password_,
```
murat-dogan: Use this Constructor for TURN Server

include/rtc/configuration.hpp
```diff
@@ -35,8 +35,9 @@ struct IceServer {
 	IceServer(const string &host_);
 	IceServer(const string &hostname_, uint16_t port_);
 	IceServer(const string &hostname_, const string &service_);
-	IceServer(const string &hostname_, const string &service_, Type type_, string username_,
-	          string password_, RelayType relayType_);
+	IceServer(const string &hostname_, const string &service_, string username_, string password_);
```
murat-dogan: Use this constructor for adding STUN with credentials

src/icetransport.cpp
```diff
@@ -59,6 +59,10 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
 
 	mMainLoopThread = std::thread(g_main_loop_run, mMainLoop.get());
 
+	mStreamId = nice_agent_add_stream(mNiceAgent.get(), 1);
```
murat-dogan: We need stream to be created before STUN Server config, so I moved this here.
Tests is OK but I am not sure ?
Please check it.

include/rtc/configuration.hpp
paullouisageneau: I might suggest something simpler, why not merge the two types in one, like:
```
enum class Type { Stun, TurnUdp, TurnTcp, TurnTls };
```

src/icetransport.cpp
```diff
@@ -102,6 +106,12 @@ IceTransport::IceTransport(const Configuration &config, Description::Role role,
 					g_object_set(G_OBJECT(mNiceAgent.get()), "stun-server", nodebuffer, nullptr);
 					g_object_set(G_OBJECT(mNiceAgent.get()), "stun-server-port",
 					             std::stoul(servbuffer), nullptr);
+
+					// Do we need to set credentials for STUN server?
+					if (!server.username.empty() && !server.password.empty())
+						nice_agent_set_local_credentials(mNiceAgent.get(), mStreamId,
```
paullouisageneau: I don't think this sets credentials for the STUN request. The function `nice_agent_set_local_credentials` actually forces the local ICE username fragment and password (the ones that are sent in the SDP and allow the remote peer to authenticate). STUN servers with credentials do not seem to be supported, probably because they are hardly ever deployed in practice.

include/rtc/configuration.hpp
murat-dogan: I think we should keep ```RelayType``` as it is. 
It must be same as NiceRelayType.

src/icetransport.cpp
murat-dogan: Actually I made tests before this commit.
But I realized now that also without credentials STUN server gives response. :)

I will delete 2. commit and will forget about STUN credentials for now.

murat-dogan:
I deleted 2. commit.
Now this PR is only for fixing logic error.
If we provide username,password than it is a TURN server.

PR #22, opened 2019-12-18
Integrate plog

murat-dogan:


CMakeLists.txt
```diff
@@ -104,5 +106,6 @@ set_target_properties(tests PROPERTIES
 	VERSION ${PROJECT_VERSION}
 	CXX_STANDARD 17)
 
+target_include_directories(tests PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/deps/plog/include)
```
paullouisageneau: You should instead set the two other `target_include_directories` as `PUBLIC`, the include will then be propagated by CMake with the library.

src/configuration.cpp
```diff
@@ -43,4 +44,10 @@ IceServer::IceServer(const string &hostname_, const string &service_, string use
     : hostname(hostname_), service(service_), type(Type::Turn), username(username_),
       password(password_), relayType(relayType_) {}
 
+Configuration::Configuration(const LogLevel logLevel_) {
+	static plog::ColorConsoleAppender<plog::TxtFormatter> consoleAppender;
+	plog::init(static_cast<plog::Severity>(logLevel_), &consoleAppender);
```
paullouisageneau: This sets the logger for the entire library, correct? What if you create different peer connections with different configurations?
My understanding is that log initialization should be a static function, maybe a function directly under the `rtc` namespace?
Also it could be cool to have an optional log appender to be able to pipe the log the the main program, like shown here: https://github.com/SergiusTheBest/plog#chained-loggers

murat-dogan: Setting logger for each config (or peer) is a complex situation.
I moved init task to rtc namespace.

include/rtc/rtc.h
```diff
@@ -41,6 +41,9 @@ typedef enum {
 	RTC_GATHERING_COMPLETE = 2
 } rtc_gathering_state_t;
 
+typedef enum { none, fatal, error, warning, info, debug, verbose } LogLevel;
+
+void rtcInitLogger(LogLevel logLevel, plog::IAppender *appender = NULL);
```
paullouisageneau: This part is dedicated to the C-compatible API, therefore you'd need another C++ function in the `rtc` namespace called by this one. I'll add it on top of your changes and merge everything.

murat-dogan: I don't see any problem about that. 
Your choice.

Issue #25, opened 2019-12-20
Datachannel Reliability


murat-dogan:
If one side of communication ended or terminated, the other side still sends messages without any error, state is still connected. 
About 5 min later state changes to disconnected.

You can make a test like that;
* Start tests/p2p apps.
* Make a connection
* Send some message
* Terminate one side
* Continue to send messages from other side
* It will send without any error
* Wait until state=disconnected message

paullouisageneau:
It's the expected behavior if you terminate one of the processes without closing the connection. Since the sender can't know the receiver left, the send method can't report an error synchronously.

However, a 5 min timeout seems indeed way too long. Some SCTP default timer might be the cause of this, I'm looking into it.

murat-dogan:
Behavior is  same even if you close connection respectively from one side and then terminate.

paullouisageneau:
OK then something is not working, I'll look into it.

PR #29, opened 2019-12-20
Fix connection shutdown

paullouisageneau:
- Reduce SCTP timeout
- Fix DataChannel close
- Fix SCTP shutdown
- Add some logging

Should fix https://github.com/paullouisageneau/libdatachannel/issues/25

### DIFF
--- a/src/sctptransport.cpp
+++ b/src/sctptransport.cpp
@@ -37,6 +37,14 @@ void SctpTransport::GlobalInit() {
 	if (InstancesCount++ == 0) {
 		usrsctp_init(0, &SctpTransport::WriteCallback, nullptr);
 		usrsctp_sysctl_set_sctp_ecn_enable(0);
+		usrsctp_sysctl_set_sctp_init_rtx_max_default(5);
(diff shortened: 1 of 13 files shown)


src/sctptransport.cpp
```diff
@@ -473,15 +519,15 @@ void SctpTransport::processNotification(const union sctp_notification *notify, s
...
-		if (reset_event.strreset_flags & SCTP_STREAM_RESET_INCOMING_SSN) {
+		if (flags & SCTP_STREAM_RESET_OUTGOING_SSN) {
 			for (int i = 0; i < count; ++i) {
 				uint16_t streamId = reset_event.strreset_stream_list[i];
 				reset(streamId);
 			}
 		}
-
-		if (reset_event.strreset_flags & SCTP_STREAM_RESET_OUTGOING_SSN) {
+		if (flags & SCTP_STREAM_RESET_INCOMING_SSN) {
```
paullouisageneau: Those were reversed, incoming on one side is outgoing on the other...

PR #35, opened 2020-02-20
Compile support on Windows with mingw-w64

murat-dogan:


murat-dogan:
Here is my dev environment;
* Win 10 64 Bit
* VS Code
* cmake 3.16
* msys2(mingw-w64)

PR #47, opened 2020-04-25
Get Selected Candidate Pair Info

murat-dogan:
Use `nice_agent_get_selected_pair` & get pair info.

### DIFF
--- a/include/rtc/candidate.hpp
+++ b/include/rtc/candidate.hpp
@@ -26,8 +26,8 @@
 namespace rtc {
 
 #if not USE_JUICE
-enum class CandidateType { HOST = 0, SERVER_REFLEXIVE, PEER_REFLEXIVE, RELAYED };
-enum class CandidateTransportType { UDP = 0, TCP_ACTIVE, TCP_PASSIVE, TCP_SO };
+enum class CandidateType { Host = 0, ServerReflexive, PeerReflexive, Relayed };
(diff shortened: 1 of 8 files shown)


include/rtc/candidate.hpp
```diff
@@ -25,6 +25,15 @@
 
 namespace rtc {
 
+#if not USE_JUICE
+struct CandidateInfo {
+	string address;
+	int port;
+	string type;
+	string transportType;
```
paullouisageneau, line 33: I'd be more practical if these were enums. You could still define an `operator<<` like for `PeerConnection::State` to print it as string transparently.

include/rtc/candidate.hpp
```diff
@@ -25,6 +25,17 @@
 
 namespace rtc {
 
+#if not USE_JUICE
+enum class CandidateType { HOST = 0, SERVER_REFLEXIVE, PEER_REFLEXIVE, RELAYED };
+enum class CandidateTransportType { UDP = 0, TCP_ACTIVE, TCP_PASSIVE, TCP_SO };
```
paullouisageneau, line 30: Could you please rename them in camel case for consistency with the other enums? Basically `Host`, `ServerReflexive`, `PeerReflexive`,  `Relayed`, `Udp`, `TcpActive`, TcpPassive`, and `TcpSo`.
After that it's good to go!

PR #49, opened 2020-04-26
Do not free candidate memory

murat-dogan:
If we free retrieved candidate memories app crashs in tests.


paullouisageneau:
Indeed, `nice_agent_get_selected_pair` does not seem to allocate a copy actually.

PR #50, opened 2020-04-26
Stats

murat-dogan:


murat-dogan:
bytesReceived gives wrong number. 
As I see it only counts message coming from data-channel.
Could you please check it?

murat-dogan:
SCTP socket is not returning rtt time as expected.
I am working on it.

murat-dogan:
Hi @paullouisageneau ,
```
const std::chrono::milliseconds SctpTransport::rtt() {
	sctp_paddrinfo info = {0};

	struct sockaddr_storage ss;
	struct sockaddr_in si;
	si.sin_family = AF_INET;
	si.sin_port = htons(mPort);
	si.sin_addr.s_addr = htonl(INADDR_ANY);

	memcpy(&ss, &si, sizeof(struct sockaddr_in));
	info.spinfo_address = ss;

	socklen_t optlen = sizeof(info);
	if (usrsctp_getsockopt(mSock, IPPROTO_SCTP, SCTP_GET_PEER_ADDR_INFO, &info, &optlen))
		return std::chrono::milliseconds(0);
	return std::chrono::milliseconds(info.spinfo_srtt);
}
```

According to RFC
```
  spinfo_address:  This is filled by the application and contains the
      peer address of interest.
```

Still can not read rtt. I think port is true but not sure about addr.

paniksystem:
to get the rtt you have to get SCTP_STATUS:

```
struct sctp_status status = {};
socklen_t len= sizeof(status);
if (usrsctp_getsockopt(this->mSock, IPPROTO_SCTP, SCTP_STATUS, &status, &len) == 0) {
		return status.sstat_primary.spinfo_srtt; // here is the current rtt of the sctp transport
}
```

src/sctptransport.cpp
```diff
@@ -445,6 +445,7 @@ int SctpTransport::handleWrite(byte *data, size_t len, uint8_t tos, uint8_t set_
 		mWritten = true;
 		mWrittenOnce = true;
 		mWrittenCondition.notify_all();
+		mBytesSent += len;
```
paullouisageneau, line 448: Actually, this will count everything SCTP outputs, counting control, retransmissions, etc.

I think the incrementation should be done when sent messages are successfully passed to SCTP instead, at the end of `trySendMessage` (when it returns `true`). Of course it should be incremented only if the message type is `String` or `Binary`.

murat-dogan:
@paniksystem Thanks for hint.

src/sctptransport.cpp
murat-dogan: I am trying p2p test app.
offerer output
```
Bytes Sent:1504 Bytes Received:0 Round-Trip Time:30 ms
```

answerer output
```
Bytes Sent:1388 Bytes Received:3 Round-Trip Time:2 ms
```

I sent a message from offerer to answerer ( 3 chars)

murat-dogan: And it counts only messages that I sent (adds message char count)

murat-dogan: OK. My fault. I was looking wrong point. 
I corrected all. Now only counting user messages.

src/peerconnection.cpp
```diff
@@ -579,6 +579,33 @@ bool PeerConnection::getSelectedCandidatePair(CandidateInfo *local, CandidateInf
...
+}
+
+const size_t PeerConnection::bytesReceived() {
+	auto sctpTransport = std::atomic_load(&mSctpTransport);
+	if (sctpTransport)
+		return sctpTransport->bytesReceived();
+	return 0;
+}
+
+const std::chrono::milliseconds PeerConnection::rtt() {
```
paullouisageneau, line 602: It could be more practical to return a `std::optional<std::chrono::milliseconds>` here, this way, we can differenciate a 0 RTT (theoretically it could happen locally) from a non available one which would result in `nullopt`. What do you think?

Also just a quick comment: `const` is superfluous on returned values.

PR #53, opened 2020-04-28
proxy support

murat-dogan:


murat-dogan:
A note from libnice;
`Note that the proxies will only be used with TCP TURN relaying.`

murat-dogan:
What will be the situation for libjuice? Should we check anything?

paniksystem:
i think that libjuice does not support TURN

paullouisageneau:
I might add TURN support to libjuice but I don't plan on adding TURN over TCP anyway to keep it light. People needing it can still use libnice instead.
Therefore you should just ignore it for libjuice, a warning is already displayed if TCP is enabled.

include/rtc/configuration.hpp
```diff
@@ -51,8 +51,19 @@ struct IceServer {
...
+	Type type = Type::None;
+	string ip;
+	uint16_t  port;
+	string username;
+	string password;
+};
+
 struct Configuration {
 	std::vector<IceServer> iceServers;
+	ProxyServer proxyServer;
```
paullouisageneau, line 66: You should make this an optional so you don't need the `None` type anymore.

include/rtc/configuration.hpp
```diff
@@ -51,8 +51,19 @@ struct IceServer {
 	RelayType relayType;
 };
 
+struct ProxyServer {
+	enum class Type { None = 0, Socks5, Http, Last = Http };
+
+	Type type = Type::None;
```
paullouisageneau: Just a suggestion, a constructor like `IceServer` could be handy.

murat-dogan:
@paullouisageneau I deleted my fork accidentally. 
Can you merge this or should I  create it again?

paullouisageneau:
@murat-dogan The merge went fine :+1:

PR #54, opened 2020-04-28
TurnTls as default relayType for turns

murat-dogan:


murat-dogan:
I could not find any document about transport=tls.

paullouisageneau:
This `transport` parameter is not really a standard I think, but since it's the way browsers parse the server URI I did the same here.

PR #55, opened 2020-05-01
define WIN32_LEAN_AND_MEAN in CMakeLists.txt

murat-dogan:
Defining in header file generates so many warnings while compiling


murat-dogan:
* warnings from my own project which includes libdatachannel as lib

PR #62, opened 2020-05-10
Javascript

willm:
Web client to native client example


CMakeLists.txt
```diff
@@ -130,7 +133,7 @@ set_target_properties(datachannel-tests PROPERTIES
 	CXX_STANDARD 17)
 set_target_properties(datachannel-tests PROPERTIES OUTPUT_NAME tests)
 target_include_directories(datachannel-tests PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
-target_link_libraries(datachannel-tests datachannel)
+target_link_libraries(datachannel-tests datachannel nlohmann_json::nlohmann_json)
```
paullouisageneau, line 136: This might not be necessary since `nlohmann/json` is header-only.

include/rtc/peerconnection.hpp
```diff
@@ -87,6 +87,8 @@ class PeerConnection : public std::enable_shared_from_this<PeerConnection> {
 	void onStateChange(std::function<void(State state)> callback);
 	void onGatheringStateChange(std::function<void(GatheringState state)> callback);
 
+	std::string connectionInfo;
```
paullouisageneau, line 90: This should be kept in the test itself.

paullouisageneau:
OK, I'll merge this in `dev` first to do some cleanup and plumbing!

CMakeLists.txt
willm: I don't think this lib is used at all, so it can actually be removed

PR #64, opened 2020-05-12
Add optional WebSocket with the same API

paullouisageneau:
Given how often a WebSocket is used for signaling, it makes sense to have an optional support for it. 
For simplicity, the implementation is abstracted behind a `WebSocket` class implementing the same `Channel` interface as `DataChannel`. WebSocket and DataChannel share the same C API.


stevedanOgochu:
Is the websocket for sending data to the signally server?

PR #69, opened 2020-05-14
check recv cb is valid

### LINKED ISSUE #68: Segmentation Fault from iceTransport
I am testing like that;
* Create peer connection
* Connect (signaling etc...)
* Check connection is successful
* Delete peer object
* Start again

I got error on this line;
https://github.com/paullouisageneau/libdatachannel/blob/8530b20dbea34d27e52ae1e8a4aeda0d51c11725/src/icetransport.cpp#L174

According to my investigation;
We are receiving data from iceTransport before initializing DTLS transport.

murat-dogan:
Fixes #68

### DIFF
--- a/src/icetransport.cpp
+++ b/src/icetransport.cpp
@@ -171,7 +171,10 @@ bool IceTransport::send(message_ptr message) {
 
 void IceTransport::incoming(message_ptr message) {
 	PLOG_VERBOSE << "Incoming size=" << message->size();
-	recv(message);
+	if (mRecvCallback)
+		recv(message);
+	else
+		PLOG_WARNING << "Received data but mRecvCallback is not valid";
 }
 
 void IceTransport::incoming(const byte *data, int size) {
--- a/src/icetransport.cpp
+++ b/src/icetransport.cpp
@@ -171,7 +171,7 @@ bool IceTransport::send(message_ptr message) {
 
 void IceTransport::incoming(message_ptr message) {
 	PLOG_VERBOSE << "Incoming size=" << message->size();
-	if (mRecvCallback)
+	if (recvCallbackValid())
 		recv(message);
 	else
 		PLOG_WARNING << "Received data but mRecvCallback is not valid";
@@ -546,7 +546,10 @@ bool IceTransport::send(message_ptr message) {
 
 void IceTransport::incoming(message_ptr message) {
 	PLOG_VERBOSE << "Incoming size=" << message->size();
-	recv(message);
+	if (recvCallbackValid())
+		recv(message);
+	else
+		PLOG_WARNING << "Received data but mRecvCallback is not valid";
 }
 
 void IceTransport::incoming(const byte *data, int size) {
--- a/src/transport.hpp
+++ b/src/transport.hpp
@@ -47,6 +47,7 @@ public:
 	virtual bool send(message_ptr message) = 0;
 
 	void onRecv(message_callback callback) { mRecvCallback = std::move(callback); }
+	bool recvCallbackValid() { return mRecvCallback; }
 
 protected:
 	void recv(message_ptr message) { mRecvCallback(message); }


src/icetransport.cpp
```diff
@@ -171,7 +171,10 @@ bool IceTransport::send(message_ptr message) {
 
 void IceTransport::incoming(message_ptr message) {
 	PLOG_VERBOSE << "Incoming size=" << message->size();
-	recv(message);
+	if (mRecvCallback)
```
paullouisageneau, line 174: `mRecvCallback` is private in the parent class `Transport`, you shouldn't be able to access it here.

murat-dogan: Yes pasted to wrong place :)

murat-dogan:
Actually I tested it. Worked as expected and I got warning messages that recv cb is not valid. (see my code)

src/transport.hpp
```diff
@@ -47,6 +47,7 @@ class Transport {
 	virtual bool send(message_ptr message) = 0;
 
 	void onRecv(message_callback callback) { mRecvCallback = std::move(callback); }
+	bool recvCallbackValid() { return mRecvCallback; }
 
 protected:
 	void recv(message_ptr message) { mRecvCallback(message); }
```
paullouisageneau, line 53: You should do the check directly here actually rather than introducing a new method.

However, this test is already done there: https://github.com/paullouisageneau/libdatachannel/blob/8530b20dbea34d27e52ae1e8a4aeda0d51c11725/include/rtc/include.hpp#L77

murat-dogan: Actually it was not making its job but somehow now it seems ok. I am investigating.

murat-dogan:
Please see the full stack below.
I am testing without any modification.

Stopped at https://github.com/paullouisageneau/libdatachannel/blob/8530b20dbea34d27e52ae1e8a4aeda0d51c11725/src/icetransport.cpp#L550

```
libc.so.6!__GI_raise(int sig) (/build/glibc-YYA7BZ/glibc-2.31/sysdeps/unix/sysv/linux/raise.c:50)
libc.so.6!__GI_abort() (/build/glibc-YYA7BZ/glibc-2.31/stdlib/abort.c:79)
libstdc++.so.6![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libstdc++.so.6!std::terminate() (Unknown Source:0)
libstdc++.so.6!__cxa_pure_virtual (Unknown Source:0)
rtc::IceTransport::incoming(rtc::IceTransport * const this, const std::byte * data, int size) (/home/murat/c++/test/lib/libdatachannel/src/icetransport.cpp:550)
rtc::IceTransport::RecvCallback(NiceAgent * agent, guint streamId, guint componentId, guint len, gchar * buf, gpointer userData) (/home/murat/c++/test/lib/libdatachannel/src/icetransport.cpp:640)
libnice.so.10![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libglib-2.0.so.0!g_main_context_dispatch (Unknown Source:0)
libglib-2.0.so.0![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libglib-2.0.so.0!g_main_loop_run (Unknown Source:0)
std::__invoke_impl<void, void (*)(_GMainLoop*), _GMainLoop*>(void (*&&)(_GMainLoop *) __f) (/usr/include/c++/9/bits/invoke.h:60)
std::__invoke<void (*)(_GMainLoop*), _GMainLoop*>(void (*&&)(_GMainLoop *) __fn) (/usr/include/c++/9/bits/invoke.h:95)
std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> >::_M_invoke<0ul, 1ul>(std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> > * const this) (/usr/include/c++/9/thread:244)
std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> >::operator()(std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> > * const this) (/usr/include/c++/9/thread:251)
std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> > >::_M_run(std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(_GMainLoop*), _GMainLoop*> > > * const this) (/usr/include/c++/9/thread:195)
libstdc++.so.6![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libpthread.so.0!start_thread(void * arg) (/build/glibc-YYA7BZ/glibc-2.31/nptl/pthread_create.c:477)
libc.so.6!clone() (/build/glibc-YYA7BZ/glibc-2.31/sysdeps/unix/sysv/linux/x86_64/clone.S:95)
```

paullouisageneau:
OK then that's actually what I guessed, `incoming()` is called while `DtlsTransport` is not fully constructed, hence the pure virtual call. This should be solved by https://github.com/paullouisageneau/libdatachannel/pull/70

murat-dogan:
OK then I am closing this PR because it seems wrong.

Issue #67, opened 2020-05-14
Segmentation Fault DtlsTransport


murat-dogan:
I am testing like that;
* Create peer connection
* Connect (signaling etc...)
* Check connection is successful
* Delete peer object
* Start again

I got error on this line;
https://github.com/paullouisageneau/libdatachannel/blob/8530b20dbea34d27e52ae1e8a4aeda0d51c11725/src/dtlstransport.cpp#L395

paullouisageneau:
You mean that you delete the peer connection then create a new one to start again?

murat-dogan:
Yes

paullouisageneau:
I'm trying to run the connectivity test in a loop but I don't reproduce the crash for now. Is there something I'm missing?

murat-dogan:
I think this is also related to issue #68 .

PR #70, opened 2020-05-14
Fix possible race condition on Transport::incoming()

paullouisageneau:
Transports should not register their `incoming()` method before they are fully constructed.

Fixes https://github.com/paullouisageneau/libdatachannel/issues/68


murat-dogan:
Possibly fixe #67 also

PR #107, opened 2020-06-28
Add the right header for IPPROTO_* on platforms such as Termux

Kyrio:
When building libdatachannel using Termux on Android, with Juice and OpenSSL, the IPPROTO_TCP and IPPROTO_UDP constants cannot be found. Adding the `<netinet/in.h>` header fixes the problem.


paullouisageneau:
:+1:

PR #121, opened 2020-07-12
add gcc flag -Wno-format-truncation to usrsctp when using b2

aldenml:
@paullouisageneau I'm in the middle of testing this


aldenml:
@paullouisageneau I can confirm that indeed this fix the build error with b2 and gcc >= 7

paullouisageneau:
Good, thanks.

PR #133, opened 2020-07-24
Extend mitigation for usrsctp send after unregistering

paullouisageneau:
This PR extends the mitigation of https://github.com/paullouisageneau/libdatachannel/pull/116 and https://github.com/paullouisageneau/libdatachannel/pull/120 by enforcing the usrsctp instance pointer check on every callback.

### DIFF
--- a/src/sctptransport.cpp
+++ b/src/sctptransport.cpp
@@ -679,8 +679,10 @@ int SctpTransport::RecvCallback(struct socket *sock, union sctp_sockstore addr, 
 	auto *transport = static_cast<SctpTransport *>(ptr);
 
 	std::shared_lock lock(InstancesMutex);
-	if (Instances.find(transport) == Instances.end())
+	if (Instances.find(transport) == Instances.end()) {
+		free(data);
 		return -1;
+	}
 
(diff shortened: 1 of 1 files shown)


src/sctptransport.cpp
```diff
@@ -676,8 +676,14 @@ std::optional<milliseconds> SctpTransport::rtt() {
 
 int SctpTransport::RecvCallback(struct socket *sock, union sctp_sockstore addr, void *data,
                                 size_t len, struct sctp_rcvinfo recv_info, int flags, void *ptr) {
-	int ret = static_cast<SctpTransport *>(ptr)->handleRecv(
-	    sock, addr, static_cast<const byte *>(data), len, recv_info, flags);
+	auto *transport = static_cast<SctpTransport *>(ptr);
+
+	std::shared_lock lock(InstancesMutex);
+	if (Instances.find(transport) == Instances.end())
+		return -1;
```
arvidn, line 683: won't `data` leak if you return here?

paullouisageneau: Indeed, thanks!

PR #135, opened 2020-07-28
add NO_TESTS option

murat-dogan:


CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ option(USE_NICE "Use libnice instead of libjuice" OFF)
 option(USE_SRTP "Enable SRTP for media support" OFF)
 option(NO_WEBSOCKET "Disable WebSocket support" OFF)
 option(NO_EXAMPLES "Disable examples" OFF)
+option(NO_TESTS "Disable tests build" ON)
```
paullouisageneau, line 13: You need to set it OFF by default, otherwise it breaks the test workflows.

murat-dogan: Sure. I was testing.

PR #142, opened 2020-08-10
Move comon lines before m= lines

stazio:
This references #141 but moves common lines before the m= lines which is necessary to allow for the ice-ufrag and such to be applied globally.


paullouisageneau:
Great, thank you!

PR #143, opened 2020-08-12
Create API for WebRTC Media Streams

stazio:
Hi Paul,
I have been able to receive media using GNU-TLS on Chromium. Something is very much broken with the OpenSSL implementation. The `use_srtp` function returns failure, but the error message that it returns says there was no error.

I had to move the postCreation call elsewhere because virtual methods don't work in constructors though I am not a fan of how this was done.


src/dtlstransport.cpp
```diff
@@ -149,7 +147,7 @@ void DtlsTransport::postHandshake() {
 
 void DtlsTransport::runRecvLoop() {
 	const size_t maxMtu = 4096;
-
+    postCreation();
```
paullouisageneau, line 150: Good catch but the fix seems unsafe. Since it's called in the thread, there is no guarantee it'll be called after `DtlsSrtpTransport` is initialized.

stazio: Absolutely. This is probably why I am currently having a 50% fail rate at starting the transport successfully...

src/dtlssrtptransport.cpp
```diff
@@ -112,8 +112,7 @@ bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 	}
 
 	message->resize(size);
-	outgoing(message);
-	return true;
+	return DtlsTransport::send(message);
```
paullouisageneau, line 115: I think this is wrong, SRTP packets must not be sent over DTLS but on the side.

src/dtlstransport.cpp
paullouisageneau: This should be reliably fixed by #147

stazio:
Just pushed a large number of changes for us to work on @paullouisageneau. This is still very much a WIP, and most definitely an ugly API to work with. Not only that, but this is surely not RFC compliant yet. For some reason this just does not work in Firefox. There is something wrong with the SDP I am offering to the browser.

Here is a list of things I have done:
- Added an RTCPSession class that manages the sending of RR's and REMB packets.
- Brought a lot of changes to description.cpp to allow for us to negotiate media.

What still needs to be done:
- Add the ability to send NACK's, ACK's, and PLI requests.
- Integrate the API more closely with rtc::PeerConnection (I will have to delegate this to you @paullouisageneau).
- Create all of the handles to send media back to the browser.
- Add a RTP demuxer that can send multiple different RTP sessions to the RTCPSession classes. This will allow us to transmit multiple tracks over one peer connection.
- Add more RFCs such as TWCC.
- Implement the specifications regarding relaying.

CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ set(LIBDATACHANNEL_HEADERS
 	${CMAKE_CURRENT_SOURCE_DIR}/include/rtc/configuration.hpp
 	${CMAKE_CURRENT_SOURCE_DIR}/include/rtc/datachannel.hpp
 	${CMAKE_CURRENT_SOURCE_DIR}/include/rtc/description.hpp
+	${CMAKE_CURRENT_SOURCE_DIR}/include/rtc/rtp.hpp
```
paullouisageneau, line 76: The file is already listed below.

Makefile
```diff
@@ -1,116 +1,1180 @@
...
-
-$(NAME).a: $(OBJS)
-	$(AR) crf $@ $(OBJS)
-
-$(NAME).so: $(LOCALLIBS) $(OBJS)
-	$(CXX) $(LDFLAGS) -shared -o $@ $(OBJS) $(LDLIBS)
-
-tests: $(NAME).a $(TEST_OBJS)
-	$(CXX) $(LDFLAGS) -o $@ $(TEST_OBJS) $(NAME).a $(LDLIBS)
+# CMAKE generated file: DO NOT EDIT!
```
paullouisageneau, line 1: I don't know if this is intentional, but please don't replace the standalone Makfile.

include/rtc/description.hpp
```diff
@@ -29,14 +29,23 @@
...
 class Description {
 public:
-	enum class Type { Unspec = 0, Offer = 1, Answer = 2 };
-	enum class Role { ActPass = 0, Passive = 1, Active = 2 };
 
-	Description(const string &sdp, const string &typeString = "");
+    enum class Type { Unspec = 0, Offer = 1, Answer = 2 };
+	enum class Role { ActPass = 0, Passive = 1, Active = 2 };
+    enum Direction {
+        SEND_ONLY,
```
paullouisageneau, line 39: For consistency with the other enums these could be named `SendOnly`, `RecvOnly`, `Both`, and `Unknown`.

src/description.cpp
```diff
@@ -328,8 +376,145 @@ Description::Media::Media(const string &mline) {
...
+            it2++;
+        }
+        if (!rem)
+
+            it++;
+    }
+}
+
+void Description::Media::addVideoCodec(int payloadType, const string &codec) {
+    RTPMap map(std::to_string(payloadType) + " " + codec + "/90000");
```
paullouisageneau, line 438: Just a suggestion, could the bitrate be a parameter here?

src/dtlssrtptransport.cpp
```diff
@@ -84,9 +84,10 @@ bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 
 	int size = message->size();
 	PLOG_VERBOSE << "Send size=" << size;
+//    return outgoing(message);
 
 	// The RTP header has a minimum size of 12 bytes
-	if (size < 12)
+	if (size < 8)
```
paullouisageneau, line 90: Out of curiosity, why is this necessary?

paullouisageneau:
That's great, thanks! I did a first pass but I'll look into it more thoroughly later.

CMakeLists.txt
stazio: This is the only occurrence `rtc.hpp`. Are you referring to `rtp.hpp`?

src/dtlssrtptransport.cpp
stazio: When I was working on this a couple of days ago, I got RR's from the browser with a report count of zero (thus 8 bytes long). I *think* a receiver/sender report with a report count of zero is legal.

CMakeLists.txt
paullouisageneau: Ha ha yes indeed, I misread!

src/dtlssrtptransport.cpp
paullouisageneau: I see! Feel free to add this as a comment.

src/description.cpp
stazio: The `b=AS:` parameter does indeed describe the maximum bitrate in the SDP. I just pushed a change that adds the API calls.
Thanks for the reminder on this, I am shocked that I can get 15mbps out of a browser! 

Though, REMB should be used to change bitrates dynamically, as renegotiating an SDP in-flight sounds terrifying.

paullouisageneau:
I found the dumb mistake that prevented the OpenSSL implementation from working: https://github.com/paullouisageneau/libdatachannel/pull/150

src/description.cpp
```diff
@@ -319,9 +555,33 @@ string Description::roleToString(Role role) {
...
+
+    this->mMedia.insert(std::pair<int, Media>(this->mMedia.size(), media));
+    return this->mMedia.at(mMedia.size()-1);
+}
 
 std::ostream &operator<<(std::ostream &out, const rtc::Description &description) {
 	return out << std::string(description);
 }
 
+}
```
paullouisageneau, line 587: Moving `operator<<` inside `rtc` namespace shouldn't compile as it is declared outside in the header.

paullouisageneau:
Good news, I found the last problem with OpenSSL: keying material order was wrong during key derivation. This is fixed in #151 and video now works as well as with GnuTLS!

src/description.cpp
```diff
@@ -132,11 +132,38 @@ Description::Description(const string &sdp, Type type, Role role)
 			} else if (key == "candidate") {
 				addCandidate(Candidate(attr, currentMedia ? currentMedia->mid : mData.mid));
 			} else if (key == "end-of-candidates") {
-				mEnded = true;
+                mEnded = true;
```
paullouisageneau, line 135: The project uses tabs to align while you used spaces. Not a big deal, but could you please retab the files so it displays nicely?

src/dtlssrtptransport.cpp
```diff
@@ -84,9 +84,11 @@ bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 
 	int size = message->size();
 	PLOG_VERBOSE << "Send size=" << size;
+//    return outgoing(message);
 
 	// The RTP header has a minimum size of 12 bytes
-	if (size < 12)
+	// An RTCP packet can have a minimum size of 8 bytes
+	if (size < 8)
```
paullouisageneau, line 91: You should also modify the equivalent test on incoming RTP.

include/rtc/rtp.hpp
```diff
@@ -0,0 +1,521 @@
...
+    std::function<void(rtc::message_ptr)> txCB;
+    SSRC ssrc;
+    uint32_t greatestSeqNo = 0, greatestTS;
+    uint64_t syncRTPTS, syncNTPTS;
+
+    unsigned int rotationCount = 0;
+
+
+public:
+    std::vector<rtc::message_ptr> sendQueue;
```
paullouisageneau, line 419: There is a dedicated thread-safe message queue that you could use in `include/rtc/queue.hpp`.

paullouisageneau:
It works well, congrats! I think what would be left after addressing my rather cosmetic comments and rebasing is mostly the RTCP implementation, right?

> Integrate the API more closely with rtc::PeerConnection (I will have to delegate this to you @paullouisageneau).
> Create all of the handles to send media back to the browser.
> Add a RTP demuxer that can send multiple different RTP sessions to the RTCPSession classes. This will allow us to transmit multiple tracks over one peer connection.

Something that would be really cool, rather than expose an API, is handling the media streams behind the scene. For instance, you create a `PeerConnection` and pass your media SDP out of your encoder to `setLocalDescription()`. libdatachannel would process it, send the SDP offer with transport and media info, and get the answer. 
However, the remote SDP as shown to the upper layer would reference UDP sockets created locally for media streams. This way, the application could directly pass to the decoder. libdatachannel would send from the UDP sockets to the decoder the way you did in your example, but it would also receive from the encoder and forward the RTP/RTCP traffic! 
On the other side, the process would be exactly the same except the media SDP would be passed as second parameter to `setRemoteDescription()`.
What do you think about that?

> Implement the specifications regarding relaying.

I guess this would be pretty separate and limited to libjuice. I can do it.

stazio:
> It works well, congrats! I think what would be left after addressing my rather cosmetic comments and rebasing is mostly the RTCP implementation, right?

Correct.



> However, the remote SDP as shown to the upper layer would reference UDP sockets created locally for media streams. This way, the application could directly pass to the decoder

That does sound rather interesting, and fun, however it is still necessary to expose the API. I think separating this out into it's own project, such as chubby, would be the right way to go here, which leverages said API. It doesn't have to be too complicated, but something like an equivalent to the Web API `addTrack`, and probably something similar to a modernized version of the `send/recv` socket calls. I would also really love to have deeper handles into RTP/RTCP or even the ability to extend `RTCPSession` with custom behavior.



> > Implement the specifications regarding relaying.
> 
> I guess this would be pretty separate and limited to libjuice. I can do it.

I was referring to the RFC standards on RTP/RTCP relaying. Like the rest of WebRTC, it's a bunch of edge-cases that need to be dealt with.

paullouisageneau:
> > However, the remote SDP as shown to the upper layer would reference UDP sockets created locally for media streams. This way, the application could directly pass to the decoder
> 
> That does sound rather interesting, and fun, however it is still necessary to expose the API. I think separating this out into it's own project, such as chubby, would be the right way to go here, which leverages said API. It doesn't have to be too complicated, but something like an equivalent to the Web API `addTrack`, and probably something similar to a modernized version of the `send/recv` socket calls. I would also really love to have deeper handles into RTP/RTCP or even the ability to extend `RTCPSession` with custom behavior.

Sure, the API could still be exposed. A web-like `addTrack` with a channel-like track object exposing a send method, a receive callback, and a receive method as alternative sounds good and would easily map to the C API. 

> I was referring to the RFC standards on RTP/RTCP relaying. Like the rest of WebRTC, it's a bunch of edge-cases that need to be dealt with.

Oh yes my bad! I've been wanting to add turn relaying for some time that's why :sweat_smile:

paullouisageneau:
If that's OK for you I'll merge this and integrate it with the track API on the `track-api` branch.

stazio:
This all looks good to me! I have to put this on hold for the next couple of days but I'll get back to this soon. That said, I don't think I will be able to have a standard RTX implementation soon...

PR #150, opened 2020-08-18
Fix SSL_set_tlsext_use_srtp() error checking

paullouisageneau:
Error checking on `SSL_set_tlsext_use_srtp()` was wrong so it always followed the error path.


src/dtlssrtptransport.cpp
```diff
@@ -52,7 +52,7 @@ DtlsSrtpTransport::DtlsSrtpTransport(std::shared_ptr<IceTransport> lower,
 #else
 	PLOG_DEBUG << "Setting SRTP profile (OpenSSL)";
 	// returns 0 on success, 1 on error
-	if (SSL_set_tlsext_use_srtp(mSsl, "SRTP_AES128_CM_SHA1_80"), "Failed to set SRTP profile")
+	if (SSL_set_tlsext_use_srtp(mSsl, "SRTP_AES128_CM_SHA1_80"))
```
paullouisageneau, line 55: :facepalm:

PR #173, opened 2020-09-22
Command line options for example server client

ecotarobles:
Example syntax

examples/client/client -s 192.168.0.68 -t 3478 -w 192.168.0.61 -x 8000 -e

STUN server is 192.168.0.68:3478, web-socket server is 192.168.0.61:8000.
-e option echoes recd messages back to sender (useful for ping timing).

Defaults and usage given by -h option and are stun.l.google.com:19302,
localhost:8000 and false (for new echo option).  Version option gives 0.5.


examples/client/cli-change.txt
```diff
@@ -0,0 +1,1918 @@
+diff --git a/examples/client/CMakeLists.txt b/examples/client/CMakeLists.txt
```
paullouisageneau, line 1: This looks like a leftover file you inadvertently committed :wink:

paullouisageneau (commented):
Thanks for the effort, it looks good! There seems to be an issue when compiling with Microsoft Visual C++ because `getopt.c` is in old-style C. You could try putting the include in an `extern "C"` block, maybe it will make it happy.

examples/client/main.cpp
```diff
@@ -43,20 +45,42 @@ unordered_map<string, shared_ptr<PeerConnection>> peerConnectionMap;
...
+void confirmOnStdout(bool echoed, string id, string type, int length);
 string randomId(size_t length);
 
 int main(int argc, char **argv) {
+	Cmdline *params;
+	try {
+		params = new Cmdline(argc, argv);
+	} catch (const std::range_error&e) {
+		std::cout<< e.what() << '\n';
+		delete params;
```
paullouisageneau, line 62: If an exception is thrown in the constructor, the object won't be allocated so you must not delete it.

examples/client/parse_cl.cpp
```diff
@@ -0,0 +1,173 @@
...
+**
+** See http://genparse.sourceforge.net for details and updates
+**
+**
+******************************************************************************/
+
+#include <stdlib.h>
+
+#if defined(_WIN32) || defined(WIN32)
+#include "getopt.c"
```
paullouisageneau, line 23: Including this C file from a C++ file won't compile because the C++ compiler will not recognize C-only syntax like the old-style function parameters list. You should put `extern "C"` around the `#include` line.

Also, you might rename the file something like `getopt_impl.h` instead of `getopt.c` since it's included, not compiled.

ecotarobles:
Thanks for all the feedback and patience.  After some adventures I've managed to get a Windows build environment set up so I can debug this.  The getopt that I initially grabbed is really old and uses totally obsolete C function declaration syntax but I've found a much newer one that's in C++ and is called Xgetopt().  Later this week I'll make sure the license is compatible and then put that in in place of the ancient C language one that I have now.  Hopefully it will both build and work.
Thanks again.
--Erik Cota-Robles, +358 41 748 5774  USA (Google Voice) +1 503 808 0042
P.S. The Windows compiler also picked up on the catch block dereference of an uninitialized pointer.  I will fix that by initializing the pointer to nullptr.  Maybe not optimal but should be sufficient. 

    On Tuesday, September 22, 2020, 04:08:42 PM GMT+3, Paul-Louis Ageneau <notifications@github.com> wrote:  
 
 


@paullouisageneau commented on this pull request.

In examples/client/parse_cl.cpp:
> +**
+** Copyright (c) 2020 Erik Cota-Robles
+**
+** Definition of command line parser class
+**
+** Automatically created by genparse v0.9.3
+**
+** See http://genparse.sourceforge.net for details and updates
+**
+**
+******************************************************************************/
+
+#include <stdlib.h>
+
+#if defined(_WIN32) || defined(WIN32)
+#include "getopt.c"

Including this C file from a C++ file won't compile because the C++ compiler will not recognize C-only syntax like the old-style function parameters list. You should put extern "C" around the #include line.

Also, you might rename the file something like getopt_impl.h instead of getopt.c since it's included, not compiled.

paullouisageneau:
OK great, keep me posted!

examples/client/cli-change.txt
ecotarobles: Sorry.  It's now deleted.

examples/client/main.cpp
ecotarobles: Fixed.

examples/client/parse_cl.cpp
ecotarobles: Will be fixed in next iteration.  I am finding Windows build failures painful to address.

paullouisageneau: Yeah Windows is a real pain for development :sweat:

paullouisageneau (approved):
Great, it looks perfect! Thank you for the contribution to the client.

PR #175, opened 2020-09-22
Added an SFU Copy-Paste Demo

stazio:
I'm back! To get back into the groove with all of your wonderful additions to the Track API (which you have done a great job with), I wrote an SFU demo to send video back to the browser. This demo takes a browser video source, changes the SSRC to [4](https://xkcd.com/221/), and sends it to an unlimited number of receivers.

PR #176, opened 2020-09-23
Changing Behavior of RtcpSession and Many Bug Fixes

stazio:
Hey @paullouisageneau, here is another massive dump of code changes. As I started using `RtcpSession`, I changed how it behaves to be a little bit more useful. I also added audio stuff to `rtc::Description`. Finally, I went through and fixed a few bugs that have managed to make their way back in. That said, I am still seeing a few bugs, currently I am stuck with `RTCP_REMB` misbehaving, but I will fix that shortly. Let's leave this PR open for a little while as I make more changes/bug fixes but I'd like your input on said changes.
This is inline with #175.


src/track.cpp
```diff
@@ -77,25 +77,39 @@ void Track::open(shared_ptr<DtlsSrtpTransport> transport) {
...
+        message->type != Message::Control) {
+        PLOG_WARNING << "Track media direction does not allow reception, dropping";
+    }
+
+	return sendControl(message);
+}
 
-	if (message->size() > maxMessageSize())
-		throw std::runtime_error("Message size exceeds limit");
+bool Track::sendControl(message_ptr message) {
```
stazio, line 97: When I wrote this, it definitely felt like a hack. I need RtcpSession to be able to send data without calling itself again in the process.

include/rtc/rtcp.hpp
```diff
@@ -20,35 +20,52 @@
...
+class Track;
+
 // An RtcpSession can be plugged into a Track to handle the whole RTCP session
-class RtcpSession : public RtcpHandler {
+class RtcpReceivingSession : public RtcpHandler {
+protected:
+    std::shared_ptr<Track> track;
 public:
-	void onOutgoing(std::function<void(rtc::message_ptr)> cb) override;
+    RtcpReceivingSession(std::shared_ptr<Track> track): track(std::move(track)) {}
```
paullouisageneau, line 56: I have actually added the `RtcpSession::onOutgoing` callback to prevent that circular dependency. Here, `Track` has a shared pointer on `RtcpHandler` and `RtcpHandler` has a backward shared pointer on `Track`, so neither will be destructed.

src/track.cpp
paullouisageneau: I think so, actually I have added the outgoing callback precisely for that reason, see below.

stazio:
So I undid the circular dependency on the RtcpHandler. I also had to start using the SSRCs to identify which track to forward to because of the RTCP. RTCP/REMB seems to work correctly now.

paullouisageneau:
Great, congratulations! Indeed, when implementing the dispatching to tracks using mids I was a bit puzzled since it can't account for RTCP.

src/track.cpp
```diff
@@ -128,20 +135,19 @@ void Track::incoming(message_ptr message) {
...
-		mRtcpHandler->onOutgoing([&]([[maybe_unused]] message_ptr message) {
-#if RTC_ENABLE_MEDIA
-			if (auto transport = mDtlsSrtpTransport.lock())
-				transport->sendMedia(message);
-#else
-			PLOG_WARNING << "Ignoring RTCP send (not compiled with SRTP support)";
-#endif
-		});
-	}
+	mRtcpHandler->onOutgoing([&](const rtc::message_ptr& message) {
```
paullouisageneau, line 139: You'll need a `[[maybe_unused]]` attribute on the `message` parameter to prevent an unused variable if media support is disabled.

src/peerconnection.cpp
```diff
@@ -544,46 +544,47 @@ void PeerConnection::forwardMedia(message_ptr message) {
...
-	unsigned int payloadType = message->stream;
+//	if (message->type == Message::Type::Control) {
+//		std::shared_lock lock(mTracksMutex); // read-only
+//		for (auto it = mTracks.begin(); it != mTracks.end(); ++it)
+//			if (auto track = it->second.lock())
+//				return track->incoming(message);
+//
+//		PLOG_WARNING << "No track available to receive control, dropping";
+//		return;
+//	}
```
paullouisageneau, line 555: You may clean up this block, it's now useless thanks to your modification to use SSRC.

include/rtc/description.hpp
```diff
@@ -140,40 +139,55 @@ class Description {
...
 	};
 
 	class Video : public Media {
 	public:
 		Video(string mid = "video", Direction dir = Direction::SendOnly);
+
+        void addVideoCodec(int payloadType, const string &codec);
+        void addH264Codec(int payloadType);
+        void addVP8Codec(int payloadType);
+        void addVP9Codec(int payloadType);
```
paullouisageneau, line 190: Good point, codecs should be separated between audio and video.

src/description.cpp
```diff
@@ -441,6 +442,18 @@ void Description::Entry::parseSdpLine(string_view line) {
 	}
 }
 
+void Description::Media::addSSRC(uint32_t ssrc, std::string name) {
+    mAttributes.emplace_back("ssrc:" + std::to_string(ssrc) + " cname:" + name);
```
paullouisageneau, line 446: Just a suggestion here: instead of comparing the attribute strings, you can store the ssrcs as integers and parse them like it's done for the direction.

stazio: Absolutely! I have plans to rewrite this very soon, just wanted to confirm it works.

stazio:
Can I ask why you pass a `std::vector<std::byte>` in the message callback instead of `rtc::message_ptr`? Forgive me if I am naive, but doesn't that involve a copy? Or at least it would require one should one choose to store that packet for a while?

include/rtc/rtp.hpp
```diff
@@ -0,0 +1,501 @@
...
+
+        // Report Count becomes the format here.
+        header.header.prepareHeader(206, 15, 0);
+
+        // Always zero.
+        header.setMediaSourceSSRC(0);
+
+        header.setPacketSenderSSRC(senderSSRC);
+
+        id[0] = 'R';
```
stazio, line 333: In your example you moved this bit of the code to be part of line 314. I brought it back down here because I am unsure how I would trigger that. If you want to move this back up there could you also fix `RtcpReceivingSession::pushREMB`?

paullouisageneau:
> Can I ask why you pass a `std::vector<std::byte>` in the message callback instead of `rtc::message_ptr`? Forgive me if I am naive, but doesn't that involve a copy? Or at least it would require one should one choose to store that packet for a while?

It doesn't involve a copy thanks to the `move` here https://github.com/paullouisageneau/libdatachannel/blob/master/src/track.cpp#L49. The only case where there is still a copy is when receiving a string message from a datachannel (the conversion to `string` currently involves a copy), however I think this could be optimized given all data is UTF-8 and there is no conversion to make. 

`Message` is not exposed in the API because `Track` has the same `Channel` interface as `Datachannel` and `WebSocket` to help factoring the code and simplifying the C API, and it's cleaner and more idiomatic in C++ for the user to be presented with `variant<binary, string>` rather than a raw message that must be interpreted differently given an enum value. If the user wants to store the packet for a while after reception, she can simply `move` it again to prevent a copy.

include/rtc/rtp.hpp
paullouisageneau: Yes, good point, let's keep it here then :+1:

stazio:
@paullouisageneau These latest commits fix the sfu-media example and thus this PR has superseded #175. There is still a bug in chrome where when we are sending video to it, it sends back RTCP with SSRC=1 and I am unsure why. There are also some missing API's in the description but I believe that major changes to this API set are complete.

src/peerconnection.cpp
```diff
@@ -256,8 +256,8 @@ bool PeerConnection::hasMedia() const {
 }
 
 std::shared_ptr<Track> PeerConnection::addTrack(Description::Media description) {
-	if (localDescription())
-		throw std::logic_error("Tracks must be created before local description");
+//	if (localDescription())
```
stazio, line 259: I did this so that I can add more tracks to an active peer connection. Am I opening a whole can of worms here?

paullouisageneau: Yeah, renegociation is not supported so the new track won't be propagated to the remote peer if you add it after setting the local description.

stazio: Now they are 😎.

src/peerconnection.cpp
```diff
@@ -694,11 +694,23 @@ void PeerConnection::openTracks() {
...
+        for (unsigned int i = 0; i < mTrackLines.size(); i++) {
+            if (auto track = mTrackLines[i].lock()) {
+                if (!track->isOpen()) {
+//                    if (track->description().direction() == rtc::Description::Direction::RecvOnly || track->description().direction() == rtc::Description::Direction::SendRecv)
+//                        srtpTransport->addInboundSSRC(0);
+//                    if (track->description().direction() == rtc::Description::Direction::SendOnly || track->description().direction() == rtc::Description::Direction::SendRecv)
+
+                    for (auto ssrc : track->description().getSSRCs())
+                        srtpTransport->addSSRC(ssrc);
+                    for (auto ssrc : std::get<rtc::Description::Media *>(remoteDescription()->media(i))->getSSRCs())
```
stazio, line 707: We should figure out a way to make these two statements a lot less messy. There needs to be a function called `activeDescription` that returns a mixture of both descriptions.

stazio:
Paul,
I had to make some more changes. One of the changes was to allow for in-flight SDP renegotiations. The other was the realization that SRTP must be aware of all of RTP/RTCP streams in-use or else the browser will slowly start to get more and more garbled data.

src/peerconnection.cpp
paullouisageneau: OK! I'm taking a note to implement renegotiation also for data channels later.

src/peerconnection.cpp
paullouisageneau: The logic looks weird to me. Why is the second for loop in the track loop, since it doesn't even depend on the track?

paullouisageneau:
OK, do you know why the catch-all entry of SRTP did not work properly?

stazio:
I am not really sure. I was sort-of hoping you might know.

src/peerconnection.cpp
stazio: It is indirectly related to the track. In essence it would have the same functionality as `track->remoteDescription()->getSSRCs()` if remoteDescription was a function inside of Track.

paullouisageneau: Yes however currently it sets up all remote tracks and only requires the local description, so why not doing it only once after the track loop?

include/rtc/rtcp.hpp
```diff
@@ -20,35 +20,63 @@
...
+     */
+    void onOutgoing(const std::function<void(rtc::message_ptr)>& cb);
+
 };
 
+class Track;
+
 // An RtcpSession can be plugged into a Track to handle the whole RTCP session
-class RtcpSession : public RtcpHandler {
+class RtcpReceivingSession : public RtcpHandler {
```
paullouisageneau, line 66: I think you need to update the existing media example with the changes.

include/rtc/peerconnection.hpp
```diff
@@ -160,9 +160,10 @@ class PeerConnection final : public std::enable_shared_from_this<PeerConnection>
 
 	std::unordered_map<unsigned int, std::weak_ptr<DataChannel>> mDataChannels; // by stream ID
 	std::unordered_map<string, std::weak_ptr<Track>> mTracks;                   // by mid
+	std::vector<std::weak_ptr<Track>> mTrackLines;                              // by SDP order
 	std::shared_mutex mDataChannelsMutex, mTracksMutex;
 
-	std::unordered_map<unsigned int, string> mMidFromPayloadType; // cache
+	std::unordered_map<unsigned int, string> mMidFromSssrc; // cache
```
paullouisageneau, line 166: The name should be `mMidFromSssrc` I guess ;)

src/dtlssrtptransport.cpp
```diff
@@ -257,41 +262,67 @@ void DtlsSrtpTransport::postHandshake() {
...
+//	srtp_crypto_policy_set_aes_cm_128_hmac_sha1_80(&outbound.rtp);
+//	srtp_crypto_policy_set_aes_cm_128_hmac_sha1_80(&outbound.rtcp);
+//	outbound.ssrc.type = ssrc_any_outbound;
+//	outbound.ssrc.value = 0;
+//	outbound.key = mIsClient ? clientSessionKey : serverSessionKey;
+//	outbound.next = nullptr;
+//
+//	if (srtp_err_status_t err = srtp_add_stream(mSrtpOut, &outbound))
+//		throw std::runtime_error("SRTP add outbound stream failed, status=" +
+//		                         to_string(static_cast<int>(err)));
```
paullouisageneau, line 293: You can remove this if each SSRC is declared instead.

src/dtlssrtptransport.cpp
```diff
@@ -257,41 +262,67 @@ void DtlsSrtpTransport::postHandshake() {
...
 
 	mInitDone = true;
 }
 
+void DtlsSrtpTransport::addSSRC(uint32_t ssrc) {
```
paullouisageneau, line 298: Ideally this should be split into two functions `addInboundSsrc` and `addOutboundSsrc`.

src/dtlssrtptransport.hpp
```diff
@@ -48,6 +49,9 @@ class DtlsSrtpTransport final : public DtlsTransport {
 
 	srtp_t mSrtpIn, mSrtpOut;
 	bool mInitDone = false;
+
+    unsigned char clientSessionKey[SRTP_AES_ICM_128_KEY_LEN_WSALT];
+    unsigned char serverSessionKey[SRTP_AES_ICM_128_KEY_LEN_WSALT];
```
paullouisageneau, line 54: For the sake of consistency, member variables should be prefixed with `m`.

paullouisageneau:
Globally this looks really good! Keep me updated about that Chrome SSRC 1 issue.

stazio:
I fixed the SSRC=1 issue, it was just a typo...

src/dtlssrtptransport.cpp
stazio: I actually did this originally, however, quickly realized that it is rather pointless because the SSRC's are more often than not bi-directional because RTCP is bi-directional and uses the same SSRC's as the RTP.

src/peerconnection.cpp
stazio: Hmm, where would you recommend I move it too then?

paullouisageneau:
> I don't really have an idea as the best way forward here, whether it is just pass that one RTCP to all of the sessions that have an report block (which will add a lot of unnecessary processing), or if we pass just the report block to the session (which will be ugly imo).

I think the first way is fine as it would indeed be way cleaner, I wouldn't worry too much about the unnecessary processing in that case. If it really becomes an issue then it might be optimized later.

stazio:
Latest commit deals with the SSRC=1 issue (rather poorly. I need to look over what I just did) . I think we may want to rethink the role of the RtcpHandler when it comes to reports because I am *slightly* concerned by having to do this much processing for a simple RTCP packet, especially when I plan on having dozens of clients all receiving dozens of tracks at the same time. It would also make sense to merge the report blocks when we send them out.

paullouisageneau:
> I am slightly concerned by having to do this much processing for a simple RTCP packet, especially when I plan on having dozens of clients all receiving dozens of tracks at the same time. It would also make sense to merge the report blocks when we send them out.

These could indeed be good enhancements, however you don't need to do everything at once!

stazio:
So apart, from having to change the behavior RtcpHandler, I think I am done making API breaking changes. I don't think I would be opposed to finally merging this into master! I have been tested the library the past couple of days and it seems to work great! There is still a few API's that need to be exposed regarding renegotiation, but I will add those in as I find the need for them. 

I have made a mistake, and I don't know how much this will impact your library yet as I am circumnavigating this issue by using two RTC sessions. I think, and I haven't tested this, that creating a sendonly and recvonly stream will cause the browser to send back an inactive SDP because that has to be a sendrecv stream. I didn't account for sendrecv streams when changing how the RtcpHandler behaves. 

Also, OpenSSL does not work at all anymore. I am hoping that you might be able to take a look as my understanding of the SSL libraries is about nilch.

src/dtlssrtptransport.cpp
```diff
@@ -90,6 +90,9 @@ bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 	if (size < 8)
 		throw std::runtime_error("RTP/RTCP packet too short");
 
+//    return outgoing(message);
+//    return DtlsTransport::send(message);
```
paullouisageneau, line 94: This should be cleaned up

src/dtlssrtptransport.cpp
```diff
@@ -174,13 +177,16 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
...
+                    PLOG_WARNING << "SRTCP unprotect error, status=" << err << " SSRC="
+                                 << ((RTCP_SR *) message->data())->senderSSRC();
+                }
+                return;
 			}
 			PLOG_VERBOSE << "Unprotected SRTCP packet, size=" << size;
 			message->type = Message::Type::Control;
-			message->stream = to_integer<uint8_t>(*(message->begin() + 1)); // Payload Type
+            auto rtp = (RTCP_SR*) message->data();
+			message->stream = rtp->senderSSRC();
```
paullouisageneau, line 189: You should check the size before accessing. libSRTP should have taken care of it, but you never know.

src/peerconnection.cpp
```diff
@@ -593,6 +586,51 @@ void PeerConnection::forwardMedia(message_ptr message) {
 			track->incoming(message);
 }
 
+std::optional<std::string> PeerConnection::getMidFromSSRC(SSRC ssrc) {
+    if (auto it = mMidFromSssrc.find(ssrc); it != mMidFromSssrc.end()) {
+        return it->second;
+    } else {
+        std::lock_guard lock(mLocalDescriptionMutex);
```
paullouisageneau, line 593: You should actually lock `mRemoteDescription` then `mLocalDescription`.

src/peerconnection.cpp
```diff
@@ -593,6 +586,51 @@ void PeerConnection::forwardMedia(message_ptr message) {
...
+                                        return media->hasSSRC(ssrc)
+                                               ? std::make_optional(media->mid())
+                                               : nullopt;
+                                    }},
+                    mLocalDescription->media(i))) {
+
+                mMidFromSssrc.emplace(ssrc, *found);
+                return *found;
+            }
+        }
```
paullouisageneau, line 629: Just a suggestion: the loop might actually be in a subroutine that you'd call twice with both descriptions, it would prevent duplicating the code.

src/peerconnection.cpp
```diff
@@ -755,8 +814,9 @@ void PeerConnection::processLocalDescription(Description description) {
 		// Add media for local tracks
 		{
 			std::shared_lock lock(mTracksMutex);
-			for (auto it = mTracks.begin(); it != mTracks.end(); ++it) {
-				if (auto track = it->second.lock()) {
+//			for (auto it = mTracks.begin(); it != mTracks.end(); ++it) {
```
paullouisageneau, line 817: You may clean this up.

include/rtc/rtcp.hpp
```diff
@@ -20,35 +20,69 @@
...
 namespace rtc {
 
-typedef uint32_t SSRC;
-
 class RtcpHandler {
+protected:
+    /**
+     * Use this callback when trying to send custom data (such as RTCP) to the client.
+     */
+    synchronized_callback<rtc::message_ptr> outgoingCallback;
```
paullouisageneau, line 37: It should be `mOutgoingCallback` for consistency.

include/rtc/rtp.hpp
```diff
@@ -0,0 +1,506 @@
...
+    }
+
+    [[nodiscard]] static unsigned int size() {
+        return sizeof(RTCP_FB_HEADER);
+    }
+};
+
+struct RTCP_FIR_PART {
+    uint32_t ssrc;
+#if __BYTE_ORDER == __BIG_ENDIAN
```
paullouisageneau, line 391: Hum, what is the idea with byte ordering here? The host byte order should have nothing to do with the order of the two fields in the packet.

include/rtc/rtp.hpp
```diff
@@ -0,0 +1,506 @@
...
+    uint32_t seqNo: 8;
+    uint32_t: 24;
```
paullouisageneau, line 393: The field sizes are actually just a suggestion for the compiler, it's not required to pack the data. It is safer to rely on the size of fields, which are guaranteed without padding since pack is set to 1 byte at the top of the file. Also, it's cleaner as you don't have to mess with byte ordering.

Here, you could use:
```
uint8_t _seqNo;
uint8_t _dummy1;
uint16_t _dummy2;
```

paullouisageneau:
Great, I suggested a few changes. Could you merge the current master an solve the conflicts once you have finished with `RtcpHandler` ?

> I think, and I haven't tested this, that creating a sendonly and recvonly stream will cause the browser to send back an inactive SDP because that has to be a sendrecv stream.

Do you know if the behavior is from the specification or if it is just a limitation of browser implementations ? In the second case, I think it's not a problem.

> Also, OpenSSL does not work at all anymore. I am hoping that you might be able to take a look as my understanding of the SSL libraries is about nilch.

I'll look into that OpenSSL issue.

src/dtlssrtptransport.cpp
```diff
@@ -174,13 +177,16 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
 					PLOG_WARNING << "Incoming SRTCP packet is a replay";
 				else if (err == srtp_err_status_auth_fail)
 					PLOG_WARNING << "Incoming SRTCP packet failed authentication check";
-				else
-					PLOG_WARNING << "SRTCP unprotect error, status=" << err;
-				return;
+				else {
+                    PLOG_WARNING << "SRTCP unprotect error, status=" << err << " SSRC="
+                                 << ((RTCP_SR *) message->data())->senderSSRC();
```
paullouisageneau, line 182: You should definitely check the size here since the packet might not have a valid RTCP header.

src/dtlssrtptransport.cpp
paullouisageneau: My bad, SSRC is from byte 4 to 8, so this is enforced by the check earlier :+1:

paullouisageneau:
The OpenSSL issue comes from `addSSRC()` being called before the handshake is done and SRTP keys are derived, because you changed the order of `postHandshake()` and `changeState(State::Connected)` for GnuTLS but not for OpenSSL. I fixed it for you in https://github.com/stazio/libdatachannel/pull/1

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "deps/json"]
 	path = deps/json
 	url = https://github.com/nlohmann/json.git
+[submodule "deps/libsrtp"]
+	path = deps/libsrtp
+	url = https://github.com/cisco/libsrtp.git
```
paullouisageneau, line 15: What is the rationale for including libsrtp as submodule?

CMakeLists.txt
```diff
@@ -161,25 +161,12 @@ if(WIN32)
...
-	target_compile_definitions(datachannel PUBLIC RTC_ENABLE_MEDIA=1)
-	target_compile_definitions(datachannel-static PUBLIC RTC_ENABLE_MEDIA=1)
-	target_link_libraries(datachannel PRIVATE SRTP::SRTP)
-	target_link_libraries(datachannel-static PRIVATE SRTP::SRTP)
-else()
-	message(STATUS "LibSRTP NOT found, compiling WITHOUT media transport")
-	target_compile_definitions(datachannel PUBLIC RTC_ENABLE_MEDIA=0)
-	target_compile_definitions(datachannel-static PUBLIC RTC_ENABLE_MEDIA=0)
-endif()
+add_subdirectory(deps/libsrtp EXCLUDE_FROM_ALL)
```
paullouisageneau, line 164: This is a new option `USE_SRTP` in `master`, you should have it here too.

stazio:
> Great, I suggested a few changes. Could you merge the current master an solve the conflicts once you have finished with `RtcpHandler` ?

This is blocked by #225.

paullouisageneau:
> > Great, I suggested a few changes. Could you merge the current master an solve the conflicts once you have finished with `RtcpHandler` ?
> 
> This is blocked by #225.

I implemented renegociation support with https://github.com/paullouisageneau/libdatachannel/pull/226

src/peerconnection.cpp
```diff
@@ -676,7 +675,10 @@ std::optional<std::string> PeerConnection::getMidFromSSRC(SSRC ssrc) {
                 return *found;
             }
         }
+	}
 
+        std::lock_guard lockB(mLocalDescriptionMutex);
```
paullouisageneau, line 680: You should unlock `mRemoteDescriptionMutex` before locking this one, as locking multiples mutexes independently creates a risk of deadlock if they are locked in a different order elsewhere.

stazio:
I have merged this branch and master into a new branch called `merge_master`. I will merge that into this PR in a couple of days once I deem it stable as I am currently using this branch in production. It was actually pretty easy to do, which is fantastic! Only issue was that the `m=` lines were out of order during renegotiation's which was an easy fix. 

The only big thing left is that I am running into issues modifying and querying the pt maps which is just some more API calls that need to be made.

src/icetransport.cpp
```diff
@@ -145,7 +145,10 @@ Description IceTransport::getLocalDescription(Description::Type type) const {
 }
 
 void IceTransport::setRemoteDescription(const Description &description) {
-	mRole = description.role() == Description::Role::Active ? Description::Role::Passive
+    if (description.role() == Description::Role::ActPass)
```
stazio, line 148: This almost certainly is not right. I had to add this test because when the browser creates the first offer, and then I renegotiate with a new offer, the browser will reject the offer because the offer's role was active.

src/track.cpp
```diff
@@ -135,21 +145,32 @@ void Track::incoming(message_ptr message) {
...
 	mRtcpHandler = std::move(handler);
 	if (mRtcpHandler) {
-		mRtcpHandler->onOutgoing([&]([[maybe_unused]] message_ptr message) {
-#if RTC_ENABLE_MEDIA
-			if (auto transport = mDtlsSrtpTransport.lock())
-				transport->sendMedia(message);
-#else
-			PLOG_WARNING << "Ignoring RTCP send (not compiled with SRTP support)";
-#endif
+		mRtcpHandler->onOutgoing([&](const rtc::message_ptr& message) {
```
paullouisageneau, line 150: You should keep `[[maybe_unused]] message_ptr message`.

src/icetransport.cpp
paullouisageneau: I see the issue, however this will make the answerer default to passive, which is not recommended as it is inefficient. 
I think it would be better to change the return line of `IceTransport::getLocalDescription` like this:
```
return Description(string(sdp), type, type == Description::Type::Offer ? Description::Role::ActPass : mRole);
```
The modification should also be done for the libnice implementation.

paullouisageneau:
Good job on the merge, this is starting to look finished :+1:

src/icetransport.cpp
stazio: Just to confirm, I am undoing what I have done, and then changing the result of `IceTransport::getLocalDescription` to what you have said.

paullouisageneau: Yes, it should work fine.

include/rtc/rtp.hpp
```diff
@@ -0,0 +1,523 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifndef WEBRTC_SERVER_RTP_HPP
+#define WEBRTC_SERVER_RTP_HPP
+
+#include <cmath>
+#include <netinet/in.h>
```
paullouisageneau, line 24: This include is not available on Windows, where you should instead include `winsock2.h` like in `rtcp.cpp`. Also, I think what you need here is actually `#include <arpa/inet.h>` for the `hton*` functions (the fact that `netinet/in.h` seems to include `arpa/inet.h` itself is more of a side-effect):
```
#ifdef _WIN32
#include <winsock2.h>
#else
#include <arpa/inet.h>
#endif
```

paullouisageneau:
Hello, what's the mergeability status on your side?

stazio:
I don't know how to fix building on MacOS but I think I can merge this at this point.

paullouisageneau:
OK, I'm merging this in a dev branch to fix the builds, then I'll merge everything to master. Thanks for the great work!

stazio:
> OK, do you know why the catch-all entry of SRTP did not work properly?

I figured this one out.
> Note that wildcard SSRC scannot be used to match both inbound and outbound traffic. This restriction is intentional, and it allows libSRTP to ensure that no security lapses result from accidental re-use of SSRC values during key sharing.
From Section 7.3.1 of [libSRTP](http://srtp.sourceforge.net/historical/libsrtp.pdf)

paullouisageneau:
@stazio If I'm not mistaken, there was two wildcards, one for inbound and one for outbound, and it sill didn't work somehow.

PR #185, opened 2020-09-25
example/client has "-n" option to not use STUN server

ecotarobles:
Bonus: removed README file that's not applicable to the updated getopt for Windows now in the client


paullouisageneau (approved):
Looks good, thank you :rocket:

Issue #200, opened 2020-09-29
Adding a C API for removing added loggers


hanseuljun:
May I ask you to add a C API that removes a logger added by rtcInitLogger? The problem is, while I'm currently working on using libdatachannel for my unity project, this leads to accumulating loggers. The reason is due to how scripts get tested inside game engine editors.

Game engines run scripts written by application developers and allows those scripts to access native libraries. However, the native libraries gets loaded only once, while the scripts run repetitively for development. So, in case the script includes adding a logger by rtcInitLogger (i.e., my case), for example, after running the script 5 times, each log appears 5 times.

I'm not sure where you would want to move "static std::optional<plog_appender> appender;" to and whether you would like to include this into rtcCleanup() or make a new function, so submitting an issue instead of a pull request.

Thanks a lot for your nice work! Also, feel free to add https://github.com/hanseuljun/datachannel-unity to your README. I'm not planning to support all platforms but it will at least work for Windows 10 PCs (and Hololens devices).

paullouisageneau:
Datachannel-unity looks great, I'll add it the the readme for sure!

The issue you describe actually comes from the C API `rtcInitLogger` not behaving as the C++ one. Would it be OK for you if calls to `rtcInitLogger` would actually change the log level and the callback (potentially deactivating logging if you set it to NULL)? I think the behavior would make a lot more sense.

paullouisageneau:
OK, could you please try with #201 ? Now the subsequent calls to `rtcInitLogger` will properly change the callback.

paullouisageneau:
I added you project to the Readme in https://github.com/paullouisageneau/libdatachannel/commit/3f084d7527bbb0ddf24cca837dfdfd2a34634f13 !

hanseuljun:
Thanks for the addition to README!

PR #201, opened 2020-09-30
Fix rtcInitLogger to prevent logging multiple times

paullouisageneau:
This PR changes the behavior of `rtcInitLogger()` in C API to allow calling it multiple times to change or unregister the callback.

Should fix https://github.com/paullouisageneau/libdatachannel/issues/200


hanseuljun:
Thanks a lot for the quick update. Just one thing is that the mutex gets locked twice here. I submitted #202 as a fix for that. I guess there needs to be another type of an approach to make the function thread-safe.

Thanks again!

PR #203, opened 2020-10-01
Add CAPI_STDCALL in cmake as an option

hanseuljun:
This option is to turn the calling convention of C API callback functions to __stdcall.

The reason this is needed, in short, is because Unity follows __stdcall even in Windows.

To elaborate this is needed is because of Unity's atypical behavior. In their effort to support many platforms, they decided to stick to __stdcall as their default calling convention for native plugins, not __cdecl, which is the default calling convention of Windows. While this discrepancy can be overcome by providing the calling convention information to C# APIs inside Unity in typical cases of calling native functions, this is not the case when a C# method becomes a callback. To make a C# method as a native function pointer, with Unity's own transpiling backend (C# -> C++), the callback C# method needs to have MonoPInvokeCallback as an attribute. Unfortunately, this attribute lacks ability to pick a calling convention and assumes __stdcall. Since this does not matches with the calling convention of C in Windows (i.e., __cdecl), any call of a callback crashes the application.

As the commit has a large space to improve, feel free to fix it! I won't have any problem as long as there is a way to set the calling convention for the callbacks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,12 @@ option(NO_WEBSOCKET "Disable WebSocket support" OFF)
 option(NO_EXAMPLES "Disable examples" OFF)
 option(NO_TESTS "Disable tests build" OFF)
 option(WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
+option(CAPI_STDCALL "Set calling convention of C API callbacks stdcall" OFF)
+
+message("CAPI_STDCALL: ${CAPI_STDCALL}")
+if(CAPI_STDCALL)
+	add_definitions(-DCAPI_STDCALL)
+endif()
 
 if(USE_NICE)
 	option(USE_JUICE "Use libjuice" OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -11,6 +11,12 @@ option(NO_WEBSOCKET "Disable WebSocket support" OFF)
 option(NO_EXAMPLES "Disable examples" OFF)
 option(NO_TESTS "Disable tests build" OFF)
 option(WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
+option(CAPI_STDCALL "Set calling convention of C API callbacks stdcall" OFF)
+
+message("CAPI_STDCALL: ${CAPI_STDCALL}")
+if(CAPI_STDCALL)
+	add_definitions(-DCAPI_STDCALL)
```
paullouisageneau, line 18: Since you would want that anything importing the library sets the definition too (otherwise it would crash) I think it would be more adapted to replace `add_definitions` with this after the declaration of targets:
```
target_compile_definitions(datachannel PUBLIC CAPI_STDCALL)
target_compile_definitions(datachannel-static PUBLIC CAPI_STDCALL)
```

hanseuljun: It definitely makes sense to have targets instead of making the definition global. I committed applying your suggestion.

paullouisageneau (approved):
Looks good, thanks.

PR #205, opened 2020-10-02
Add rtcGetLocalDescriptionSdp() as a C API

hanseuljun:
Thanks a lot for keep quickly applying the stuff I submit here.

WebRTC requires setting a description before candidates for it. When sending the description and candidates whenever they are found, in a networked scenario, this means that there needs to be a buffer for candidates from the side the receives the description and candidates as there is often no guarantee that packets sent will arrive in the order they were sent.

An approach that I would much more prefer than having a candidate buffer is sending a SDP message that includes a description with all candidates inside it, as demonstrated in /examples/media. This addition of an API allows this approach.

I am not sure whether gathering candidates can take a long time, but from my experience, they tend to happen almost immediately; There is no feasible reason to hurry sending an incomplete description.

### DIFF
--- a/src/capi.cpp
+++ b/src/capi.cpp
@@ -529,6 +529,27 @@ int rtcAddRemoteCandidate(int pc, const char *cand, const char *mid) {
 	});
 }
 
+int rtcGetLocalDescriptionSdp(int pc, char *buffer, int size) {
+	return WRAP({
+		auto peerConnection = getPeerConnection(pc);
+
+		if (size <= 0)
+			return 0;
+
+		if (!buffer)
+			throw std::invalid_argument("Unexpected null pointer for buffer");
+
+		if (auto desc = peerConnection->localDescription()) {
+			auto sdp = string(*desc);
+			const char *data = sdp.data();
+			size = std::min(size - 1, int(sdp.size()));
+			std::copy(data, data + size, buffer);
+			buffer[size] = '\0';
+			return size + 1;
+		}
+	});
+}
+
 int rtcGetLocalAddress(int pc, char *buffer, int size) {
 	return WRAP({
 		auto peerConnection = getPeerConnection(pc);
--- a/src/capi.cpp
+++ b/src/capi.cpp
@@ -547,6 +547,8 @@ int rtcGetLocalDescriptionSdp(int pc, char *buffer, int size) {
 			buffer[size] = '\0';
 			return size + 1;
 		}
+
+		return RTC_ERR_FAILURE;
 	});
 }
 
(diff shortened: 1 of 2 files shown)


hanseuljun:
I don't think adding a function not involved in any test can cause a test failure but in case my commit was the reason for the above failure, please let me know.

paullouisageneau:
> I don't think adding a function not involved in any test can cause a test failure but in case my commit was the reason for the above failure, please let me know.

Don't worry, it's just the Windows tests being capricious, I think because timeouts might be a bit short.

paullouisageneau:
This was indeed missing from the C API. Not doing candidate trickling and just sending the full description is fine, provided you wait for gathering to be completed to get the full local description. Candidates gathering actually can take some time, because even if host candidates are gathered immediately, server reflexive ones depend on the response of the STUN server, which could answer only after a couple retransmissions or not at all.

You still lose some optimization without trickling, as it allows two peers on the same network to connect immediately even if the STUN server is long to answer or unreachable, but depending on the use case this might not be an issue.

paullouisageneau (approved):

src/capi.cpp
```diff
@@ -529,6 +529,27 @@ int rtcAddRemoteCandidate(int pc, const char *cand, const char *mid) {
...
+			throw std::invalid_argument("Unexpected null pointer for buffer");
+
+		if (auto desc = peerConnection->localDescription()) {
+			auto sdp = string(*desc);
+			const char *data = sdp.data();
+			size = std::min(size - 1, int(sdp.size()));
+			std::copy(data, data + size, buffer);
+			buffer[size] = '\0';
+			return size + 1;
+		}
```
paullouisageneau, line 549: I approved a bit too fast, here it would be better to handle the case where `desc` is not present and return `RTC_ERR_FAILURE`, otherwise the wrapper will return `RTC_ERR_SUCCESS` (i.e. 0) by default, which sounds more like the description is there but empty.

paullouisageneau (changes requested):
Just add a `return RTC_ERR_FAILURE` and that looks good to me!

src/capi.cpp
hanseuljun: I added it.

One thing I would like to ask is, does that mean rtcGetLocalAddress and rtcGetRemoteAddress should be modified the same way?

paullouisageneau (approved):
Perfect, thanks!

src/capi.cpp
paullouisageneau: Yes indeed, I'll do it!

Issue #206, opened 2020-10-03
Replacing the last parameter of C callbacks from pointers to indices (the so called user pointers)


hanseuljun:
The C API is designed as it expects access to instances via indices, such as pc and dc. However, the callbacks return the actual pointers of the corresponding instances, not the indices that are used in other functions. This makes finding which instance was the callback function called very difficult.

I propose replacing, for example
typedef void (RTC_API *rtcDescriptionCallbackFunc)(const char *sdp, const char *type, void *ptr);
to
typedef void (RTC_API *rtcDescriptionCallbackFunc)(const char *sdp, const char *type, int pc);

The direct access to instances would be still possible by adding rtcGetUserPointer(int id). 

While I think this won't harm the library, I wanted to check how you think about this before working on its implementation.

paullouisageneau:
I get your point, my only worry here is that it'll break existing wrappers which will need to be updated.

What is your use case exactly? My stance on the user pointer is that it's an opaque type so you can pass whatever you like in it. If you don't need a user structure on your side, couldn't you just pass the indices as user pointer?

hanseuljun:
First, I misunderstood user pointers as somewhat a generalized version of IDs. Thanks for the clarification.

The use case here is a Unity (C#) application with many PeerConnections. Since Unity only allows static methods as callbacks for C code, to avoid having 20 static methods to support 5 PeerConnections, there should be an identifier coming from callback calls.

Definitely I can use void* to contain the IDs as void* is an address, therefore can contain non-negative integers as IDs. But still this involves casting an integer to void* while the integer is not actually an address, which seems dangerous. Another case of sending the addresses of IDs as UserPointers would allow avoid casting a non-address as an address but this would make code written by a language with garbage collection and its own memory model ugly, so I would like to avoid it.

And going one step forward, I would like to propose changing the order of parameters. As changing "void* ptr" to "int pc" implies imitating instance-based callbacks, and since oftentimes people place the parameter to imitate "this" of the OOP languages the first in these cases,
typedef void (RTC_API *rtcDescriptionCallbackFunc)(int pc, const char *sdp, const char *type);
might be more comprehensible.

I think this order thing comes from how people explain and understand the similarity between a.f(b) and g(a, b) when g is the following:
g(A a, B b) { a.f(b) }

hanseuljun:
[...]

PR #208, opened 2020-10-04
Add id as a parameter to C API callbacks.

hanseuljun:
An update based on your suggestion at #206


paullouisageneau:
Great! You also need to update the `copy-paste-capi` example.

hanseuljun:
> Great! You also need to update the `copy-paste-capi` example.

Done!

paullouisageneau (approved):
OK good, thank you.

Issue #216, opened 2020-10-21
Asking opinion UWP support


hanseuljun:
UWP is what Microsoft is pushing as an application framework, and unfortunately, since I work on AR headsets and Microsoft is currently the only competitive manufacturer of them, I have no choice but follow their directions. However, I am not sure you would like to perfectly support UWP.

Currently, libdatachannel works fine for UWP as a library, but examples and tests don't work. I can detour this with the NO_TESTS and NO_EXAMPLES options you thoughtfully provide, but wanted to ask you are interested in the fixes of the tests and examples.

The root cause of failure to build them comes from an old mistake from Microsoft. When everyone started supporting unicode, instead of utf-8, they chose utf-16, which is based on 16-byte characters. This made them end of using wchar_t instead of char for strings. So, when you build applications to support unicode, the Windows APIs by default want const wchar_t* as inputs, instead of const char*.

Also, making this a little bit more complicated, though it is part of Microsoft finally making the right decision, they are coming back to utf8: https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page. If you go this way, this means code such as examples/client/getopt.h becoming obsolete. They handle defining of _UNICODE as a sign to use wchar_t instead of char, which is no longer the recommendation.

I'll stop here for now since this is already quite terrible, but in case you interested in adding UWP support to libdatachannel, let me know. To build as UWP with cmake, you can add
-DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION="10.0"

It means an application framework for Windows Store (Microsoft had many of them) and the one for Windows 10 (UWP, the current one which I hope last longer than its ancestors.)

An example command would be
cmake -S . -B .\build -G "Visual Studio 16 2019" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION="10.0"

paullouisageneau:
I'm a bit familiar with that Microsoft UTF-16 choice to support unicode, it's really a pain.

So if I understand correctly, in UWP applications, you have to use only the wide char version of functions in the Win32 API, you can't use the ANSI version. What is exactly the issue with tests then? Sorry, I don't have a Windows environment to test right now.

hanseuljun:
I'm glad you have interest for UWP support!

[...]

PR #219, opened 2020-10-24
Fix compilation of tests and examples for UWP

paullouisageneau:
See https://github.com/paullouisageneau/libdatachannel/issues/216


paullouisageneau:
Following https://github.com/paullouisageneau/libdatachannel/pull/221 and since there seem to be no feedback from the cmake side for now, I simply disabled setting `OUTPUT_NAME` for UWP as a workaround. 

@hanseuljun please tell me if it solves the compilation issue.

PR #221, opened 2020-10-26
Adding RTC_API to the deletePeer()s.

hanseuljun:
A slight addition to #219.

While this removes most of the error messages from a uwp build, this does not fully fix them all. A problem I found is related to OUTPUT_NAME of set_target_properties in cmake. When creating a uwp application via cmake, it creates a .appxManifest file for the visual studio project.

image

It adds an Application with the exe file of the cmake project in it, and while doing this, the OUTPUT_NAME does not get applied. Due to this cmake bug, setting OUTPUT_NAME leads to errors...

image

The simplest fix would be not renaming applications using OUTPUT_NAME. I'll look further at this cmake bug if I can (e.g., checking whether future versions of cmake fixes this; mine is 3.17).

Though this leaves some error messages, since this including #219 is an apparent improvement that deserves a merge, in case my input helps.


paullouisageneau:
Actually, it's a mistake on my part, `deletePeer` should not have been marked `RTC_API` at declaration here: https://github.com/paullouisageneau/libdatachannel/blob/6ec75d7c700a86c928f6df42d653178cef4a6913/examples/copy-paste-capi/answerer.c#L51

hanseuljun:
Oh I see. I'll do this tomorrow.

paullouisageneau:
> It adds an Application with the exe file of the cmake project in it, and while doing this, the OUTPUT_NAME does not get applied. Due to this cmake bug, setting OUTPUT_NAME leads to errors...
> The simplest fix would be not renaming applications using OUTPUT_NAME. I'll look further at this cmake bug if I can (e.g., checking whether future versions of cmake fixes this; mine is 3.17).

I'd like to keep the output name renaming because it allows for prefixed target names (like `datachannel-tests`) while keeping the short application file names (like `tests`). Maybe there is an obscure other variable name expected by the manifest generation? In the worst case, I think setting the output name could be done only if the platform is not UWP.

Issue #225, opened 2020-10-31
Allow Renegotiation of Media


stazio:
Hi @paullouisageneau, would it be possible for you to add the ability to do renegotiation's? I can't merge #176 with master because it breaks my (admittedly janky) implementation of renegotiation's for media and I would like for renegotiation's to have first-class support.

paullouisageneau:
Yes it's possible, however it needs some rather extensive modifications to be done properly, like the adding the signaling state, so I hope it won't create to much merging difficulties for you.

paullouisageneau:
Renegociation for data channels and tracks is implemented in https://github.com/paullouisageneau/libdatachannel/pull/226

PR #226, opened 2020-11-01
Implement renegotiation

paullouisageneau:
Implements https://github.com/paullouisageneau/libdatachannel/issues/225

This PR implements renegotiation for data channels and tracks.

### DIFF
--- a/include/rtc/description.hpp
+++ b/include/rtc/description.hpp
@@ -34,8 +34,8 @@ namespace rtc {
 
 class Description {
 public:
-	enum class Type { Unspec = 0, Offer = 1, Answer = 2 };
-	enum class Role { ActPass = 0, Passive = 1, Active = 2 };
+	enum class Type { Unspec, Offer, Answer, Pranswer, Rollback };
+	enum class Role { ActPass, Passive, Active };
 	enum class Direction { SendOnly, RecvOnly, SendRecv, Inactive, Unknown };
 
 	Description(const string &sdp, const string &typeString = "");
@@ -45,7 +45,6 @@ public:
 	Type type() const;
 	string typeString() const;
 	Role role() const;
-	string roleString() const;
 	string bundleMid() const;
 	string iceUfrag() const;
 	string icePwd() const;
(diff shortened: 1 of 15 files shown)


stazio:
Wow, thanks for the quick turn around on this! That was impressive. However, I am running into a few issues, and I think they would just be some API's that need to be exposed.

My current implementation does not mark tracks as `Inactive` when I stop using them, I just stop sending traffic to them. This is acceptable, and preferred, because of the rate at which I need to stop and start sending different audio streams. I still renegotiate my video stream by providing a new SSRC for them because the browsers get really fussy about it. I need an API exposed that allows me to modify descriptions of tracks and re-initiate the negotiation.
I have also come to the realization that for some reason, the description passed to `setRemoteDscription` (even when this is not a renegotiation), claims that the Direction is `SendRecv`, when the [SDP answer](https://gist.github.com/stazio/7bf41eb000e1c13123023a88ca15f59c) most definitely does not say that. This causes issues for future negotiations because that is sent in the offer. Also, the SSRC that gets sent by the browser back to us (when the browser is `RecvOnly`) is then sent back to the browser when I add a track, which is also not the correct behavior. It is my decision, as the sender, to pick the SSRC I send to the browser. Do you know what could be causing this?

paullouisageneau:
> My current implementation does not mark tracks as `Inactive` when I stop using them, I just stop sending traffic to them. This is acceptable, and preferred, because of the rate at which I need to stop and start sending different audio streams. I still renegotiate my video stream by providing a new SSRC for them because the browsers get really fussy about it. I need an API exposed that allows me to modify descriptions of tracks and re-initiate the negotiation.

I added the possibility to change a track description in https://github.com/paullouisageneau/libdatachannel/pull/228

> I have also come to the realization that for some reason, the description passed to `setRemoteDscription` (even when this is not a renegotiation), claims that the Direction is `SendRecv`, when the [SDP answer](https://gist.github.com/stazio/7bf41eb000e1c13123023a88ca15f59c) most definitely does not say that. This causes issues for future negotiations because that is sent in the offer. 

This looks like a bug in description parsing, however for now I don't see the issue.

> Also, the SSRC that gets sent by the browser back to us (when the browser is `RecvOnly`) is then sent back to the browser when I add a track, which is also not the correct behavior. It is my decision, as the sender, to pick the SSRC I send to the browser. Do you know what could be causing this?

Yes, for now the `Description::Media::reciprocate()` method is rather dumb, you will want to enhance it. In this case, I think you need to drop the SSRC in the `reciprocated` description if the direction is `RecvOnly`.

stazio:
#228 fixes all of the issues I had!

PR #236, opened 2020-11-07
Fix rtcSetAvailableCallback

paullouisageneau:
This PR fixes `rtcSetAvailableCallback()` as it did not set the right callback.


lerouxrgd:
:+1:

PR #237, opened 2020-11-07
Fix rtcReceiveMessage() and harmonize C API

paullouisageneau:
This PR fixes `rtcReceiveMessage()` and refactors buffer handling in C API.

It introduces the new error codes `RTC_ERR_NOT_AVAIL` and `RTC_ERR_TOO_SMALL`:
-  `RTC_ERR_NOT_AVAIL` is now returned when the fetched element is not available instead of `RTC_ERR_FAILURE`, including `rtcReceiveMessage()` when no message is available.
-  `RTC_ERR_TOO_SMALL` is now returned by every function taking a buffer when data won't fit instead of clamping the data.
- A `NULL` buffer now makes the function return the size without changing the state, allowing the user to dimension the buffer.

### DIFF
--- a/include/rtc/channel.hpp
+++ b/include/rtc/channel.hpp
@@ -54,7 +54,8 @@ public:
 
 	// Extended API
 	virtual std::optional<message_variant> receive() = 0; // only if onMessage unset
-	virtual size_t availableAmount() const; // total size available to receive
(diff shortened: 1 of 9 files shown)


lerouxrgd:
:+1:

PR #240, opened 2020-11-10
Mac compatibility

LeoNavel:
1. Added generate XCode project to README
2. Fixed build error on macOS (missing clang reinterpret_pointer_cast implementation)

### DIFF
--- a/src/peerconnection.cpp
+++ b/src/peerconnection.cpp
@@ -35,7 +35,7 @@
 #include <thread>
 
 #if __clang__
-namespace std {
+namespace {
 
 template <typename To, typename From>
 inline std::shared_ptr<To> reinterpret_pointer_cast(std::shared_ptr<From> const & ptr) noexcept {
@@ -43,6 +43,8 @@ inline std::shared_ptr<To> reinterpret_pointer_cast(std::shared_ptr<From> const 
 }
 
 }
+#else
+using std::reinterpret_pointer_cast;
 #endif
 
 namespace rtc {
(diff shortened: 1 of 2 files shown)


src/peerconnection.cpp
```diff
@@ -33,6 +34,17 @@
...
+namespace std {
+
+template <typename To, typename From>
+inline std::shared_ptr<To> reinterpret_pointer_cast(std::shared_ptr<From> const & ptr) noexcept {
+    return std::shared_ptr<To>(ptr, reinterpret_cast<To *>(ptr.get()));
+}
+
+}
+#endif
+
```
paullouisageneau, line 47: Indeed, `reinterpret_pointer_cast` was missing from clang's libc++ and has been [added only recently](https://reviews.llvm.org/rGce195fb22b542c20745e3ed46d1516e8b4e3a76f).

I think adding the function directly in the `std` namespace may cause a conflict when it is available in the local libc++. Why not definining it in an empty namespace instead?

```suggestion
#if __clang__
namespace {
template <typename To, typename From>
inline std::shared_ptr<To> reinterpret_pointer_cast(std::shared_ptr<From> const & ptr) noexcept {
    return std::shared_ptr<To>(ptr, reinterpret_cast<To *>(ptr.get()));
}
}
#else
using std::reinterpret_pointer_cast;
#endif
```
And then replace `std::reinterpret_pointer_cast` calls with `reinterpret_pointer_cast`.

LeoNavel: You are right. I have removed reinterpret_pointer_cast implementation from std namespace.

paullouisageneau (approved):
Looks good, thanks!

PR #248, opened 2020-11-14
Refactor RTCP dispatching

paullouisageneau:
This PR refactors the RTCP dispatching "monstruosity". It is way more concise but it should behave the same @stazio

### DIFF
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -139,8 +139,7 @@ private:
 	void forwardMessage(message_ptr message);
 	void forwardMedia(message_ptr message);
 	void forwardBufferedAmount(uint16_t stream, size_t amount);
-	std::optional<std::string> getMidFromSSRC(SSRC ssrc);
-	std::optional<uint32_t> getMLineFromSSRC(SSRC ssrc);
+	std::optional<std::string> getMidFromSsrc(uint32_t ssrc);
 
 	std::shared_ptr<DataChannel> emplaceDataChannel(Description::Role role, string label,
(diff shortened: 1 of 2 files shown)


include/rtc/peerconnection.hpp
```diff
@@ -139,8 +139,7 @@ class PeerConnection final : public std::enable_shared_from_this<PeerConnection>
 	void forwardMessage(message_ptr message);
 	void forwardMedia(message_ptr message);
 	void forwardBufferedAmount(uint16_t stream, size_t amount);
-	std::optional<std::string> getMidFromSSRC(SSRC ssrc);
-	std::optional<uint32_t> getMLineFromSSRC(SSRC ssrc);
```
stazio, line 143: So I added this in as an optimization. Is it a good idea to be performing so many string compares in the real-time code?

paullouisageneau: I think it's OK since the strings are guaranteed to be very small. The comparison is indeed a bit slower but it looks like an over-optimization since you will never have thousands of tracks.

stazio: That's not necessarily true. I, or my peer, can name the mids whatever I wish. The last time I looked, `rtc::Description` defaults to setting the mid to `video` or `audio`.

paullouisageneau: Yes, this is short, and the RFC actually recommends 4 characters or less. If there is a performance issue in some extreme use case, it would still be quick to add a SSRC to track cache, without re-implementing the mapping for line indexes.

Issue #253, opened 2020-11-15
Renegotiation does not set the SSL role correctly


stazio:
It seems this bug is still present. The browser is failing to accept the remote answer with the warning `Failed to apply the description for m= section with mid='1': Failed to set SSL role for the transport.` where `mid:1` is the first mline in the description. Here is how to reproduce the issue:

1. libdatachannel produces an offer (setup:actpass)
2. The browser accepts and produces an answer (setup:active)
3. libdatachannel accepts the answer
4. The browser produces an offer (setup:actpass)
5. libdatachannel accepts and produces an answer (setup:active)
6. The browser rejects the offer.

paullouisageneau:
OK, I'm pushing a fix.

stazio:
There is still something wrong here. libdatachannel now throws a "Incompatible roles with remote description" exception. I have attached a gist [here](https://gist.github.com/stazio/ec72d0607a2d3d4be3a66799ab47d634) of both the offer and the answer from the perspective of the browser.

stazio:
Commenting out the offending exception produces no issues as far as I can tell.

paullouisageneau:
This indicates a deeper problem as it means the local setup and the remote setup are the same. How do you end up in this situation?

stazio:
> How do you end up in this situation?

So I went to sleep, and woke up this morning and I can no longer reproduce the issue. This certainly has to do with the fact that both the browser and libdatachannel can produce an offer, though the signaling guarantees that only one offer is ever in progress at any time.

paullouisageneau:
OK, I see. The API should behave a bit like the browser one so you have to either offer always from the same side or handle conflicts yourself using the signaling state. Please re-open if you encounter such an issue again.

PR #255, opened 2020-11-15
Fix roles during renegociation

paullouisageneau:
Fixes https://github.com/paullouisageneau/libdatachannel/issues/253


stazio:
Thank you for the fix! it works beautifully.

PR #264, opened 2020-11-18
Exporting C++ API for Windows

hanseuljun:
MSVC requires adding __declspec(dllexport) in front of functions or in between class declarations, otherwise they don't allow access to them through shared libraries. So, I propose start adding these declarations. While one selfish reason of this pull request is to make functions I need accesible, the other is to start the discussion of how to do this.

In terms of implementation, I tried following the approach using "include.hpp" as a global header file. The problem was that it is not a perfectly global header file. For example it is not contained in log.hpp. Also, it is an .hpp file, not a .h file. So I added a new macro RTC_CPP_EXPORT instead of reusing RTC_EXPORT from rtc.h.

In my opinion, there are two decisions to make.
1. One is where to place the definition of RTC_CPP_EXPORT. Would you like to keep it in include.hpp and add include.hpp to log.hpp, or do you want to find a new approach?
2. Do you want to merge RTC_CPP_EXPORT (for c++) and RTC_EXPORT (for c) into one? They follow the same logic, so it is definitely possible (for example by adding something like export.h that only contains RTC_EXPORT). This would also log.hpp not including include.hpp but it means adding another file (i.e., export.h).

Let me know how you think about this. If you pick an approach, I can add __declspec(dllexport) to all classes!


paullouisageneau:
>     1. One is where to place the definition of RTC_CPP_EXPORT. Would you like to keep it in include.hpp and add include.hpp to log.hpp, or do you want to find a new approach?

Yes I think including include.hpp from log.hpp is the way to go.

>     2. Do you want to merge RTC_CPP_EXPORT (for c++) and RTC_EXPORT (for c) into one? They follow the same logic, so it is definitely possible (for example by adding something like export.h that only contains RTC_EXPORT). This would also log.hpp not including include.hpp but it means adding another file (i.e., export.h).

You may keep both `RTC_CPP_EXPORT` and `RTC_EXPORT`, it's fine by me!

include/rtc/candidate.hpp
```diff
@@ -25,7 +25,7 @@
 
 namespace rtc {
 
-class Candidate {
+class RTC_CPP_EXPORT Candidate {
```
paullouisageneau, line 28: You also need to export the << operators at the end of the file.

include/rtc/description.hpp
```diff
@@ -32,7 +32,7 @@
 
 namespace rtc {
 
-class Description {
+class RTC_CPP_EXPORT Description {
```
paullouisageneau, line 35: Same here, you also need the << operators at the end.

hanseuljun: I only added RTC_CPP_EXPORT to the ones that caused an error to my project. I'll update based on your choices soon!

hanseuljun:
I have added a commit to this branch but it does not appear here... I'm not sure whether I am doing something wrong or is this a github bug... The commit is below:
https://github.com/hanseuljun/libdatachannel/commit/679c0ccd2e4c084a4836d774e3a0471a80f85e0b

If this does not resolve until tomorrow, I'll do something like closing this pull request and making another one...

paullouisageneau:
The commit appears fine for me!

> 2. Placed the definition of RTC_CPP_EXPORT in log.hpp. I chose log.hpp instead of include.hpp as include.hpp includes log.hpp. I felt you would prefer having it in include.hpp but that seemed like have to be preceded by changing the order in the header chain...

Since headers have a guard you could include include.hpp from log.hpp while keeping the log.hpp include from include.hpp, just be sure to set `RTC_CPP_EXPORT` in include.hpp before including log.hpp.

> 3. Added ignoring C4251 for MSVC in cmake. What it is trying to say is exporting a class that contains another non-exported class can cause a problem. There are two reasons I decided to ignore the warning is that, though it is not wrong. First, the fix requires a huge amount of work like not using STL in any public class. Second, MSVC warns this 2500 times resulting in almost a DDOS attack to consoles.

Ha ha good one, typical MSVC :laughing: 

> The thing I was not sure about your opinion was about how much classes do you want to make public. Do you plan to make a subset of your C++ library as public interface and others private? Then, I can limit dllexports only to them.

Actually only what is in `include` is public, headers in `src` are private, so you should remove the export for them (these headers can't be reached externally anyway). Apart from that, I think you only miss the ostream operators in `description.hpp`.

hanseuljun:
Thanks, I have updated based on your guidance! Sorry for messing up your git history with the classes in /src 😅

paullouisageneau (approved):
It looks good, thanks!

PR #266, opened 2020-11-19
fix build warning on windows when built with MSVC /W4.

supercairos:
Some arguments where conflicting with data member names.


paullouisageneau (approved):
Thanks for the fix!

PR #273, opened 2020-11-23
Fixed a typo during the protection of RTP packets

stazio:
There is a typo in the protection of SRTP packets.


paullouisageneau (approved):
OMG so obvious, thank you. This makes me think that the current media test do not attempt to pass RTP traffic for now, whereas it would allow to catch this kind of regression.

PR #283, opened 2020-12-03
Fix first packet dropped

iturrioz:
In DtlsSrtpTransport, when the first packet is received, the context is not set yet. After adding the SSRC, the code tries to unprotect the packet, and returns without checking the result of that call.

### DIFF
--- a/src/dtlssrtptransport.cpp
+++ b/src/dtlssrtptransport.cpp
@@ -206,7 +206,8 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
 					PLOG_WARNING << "SRTCP unprotect error, status=" << err
 					             << " SSRC=" << ((RTCP_SR *)message->data())->senderSSRC();
 				}
-				return;
+				if (err)
+					return;
 			}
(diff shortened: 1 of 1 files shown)


src/dtlssrtptransport.cpp
```diff
@@ -228,7 +228,8 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
 				} else
 					PLOG_WARNING << "SRTP unprotect error, status=" << err
 					             << " SSRC=" << reinterpret_cast<RTP *>(message->data())->ssrc();
-				return;
+					if (err)
+						return;
```
paullouisageneau, line 232: Good catch, could you please add brackets around the `PLOG_WARNING` line for clarity and fix the similar issue with `srtp_unprotect_rtcp()` a few lines higher?

iturrioz: I guess we could add the same check in the "Range 64-95" block above

paullouisageneau: Yes sounds good!

paullouisageneau (approved):
Looks good, thanks!

PR #285, opened 2020-12-08
Fixes a few potential seg-faults in the sending of RTP and fixes the building of NACK packets

stazio:
Goodness gracious! I have been struggling to identify this bug for weeks now! I added a `std::lock_guard` at `DtlsSrtpTransport::sendMedia` because I can call `RtcpHandler::outgoingCallback` at the same time as `Track::send` and cause race conditions that shows visible data corruption and can even potentially cause a segfault in SRTP! I also removed that double decode of SRTP and I have opened #286 to discuss how to fix it. It can't be done because the documentation doesn't guarantee the safety of those parameters (although the code shows no reason why it can't be done).

### DIFF
--- a/include/rtc/rtp.hpp
+++ b/include/rtc/rtp.hpp
@@ -459,8 +459,9 @@ public:
 			return true;
 		} else {
 			// TODO SPEEED!
-			parts[(*fciCount) - 1].blp = htons(ntohs(parts[(*fciCount) - 1].blp) |
-			                                   (1u << (unsigned int)(missingPacket - *fciPID)));
+			auto blp = ntohs(parts[(*fciCount) - 1].blp);
+			auto newBit = 1u << (unsigned int)(missingPacket - (1 + *fciPID));
+			parts[(*fciCount) - 1].blp = htons(blp | newBit);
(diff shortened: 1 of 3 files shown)


include/rtc/rtp.hpp
```diff
@@ -459,8 +459,9 @@ class RTCP_NACK {
 			return true;
 		} else {
 			// TODO SPEEED!
-			parts[(*fciCount) - 1].blp = htons(ntohs(parts[(*fciCount) - 1].blp) |
-			                                   (1u << (unsigned int)(missingPacket - *fciPID)));
+			auto blp = ntohs(parts[(*fciCount) - 1].blp);
```
paullouisageneau, line 462: Is there a reason why `fciCount` and `fciPID` are passed by pointers and not by references in this function, since they may not be null? Also, why is `missingPacket` a const reference.

stazio: The first two are personal preference as I prefer passing by pointer when that value is going to be modified. The last one just happened.

paullouisageneau (approved):

Issue #281, opened 2020-12-01
Adding libjuice-based TURN support


hanseuljun:
Hi again!

I am back with another edge case with UWP. So, this time, I am looking for adding TURN support to a UWP application. The problem is, my colleague at another state, who should use this application seemingly lives at a place where TURN support is required. If I don't add this, I may need to ask him to go to his lab for research at least once a week :(

While this is the case, I figured out that I cannot easily build libnice for UWP since it depends on glib, which does not support UWP and is gigantic. Of course, there is nothing impossible in terms of porting libraries, but seeing the below slides describing how porting GStreamer for UWP was makes me feel like running away from this track.
https://gstreamer.freedesktop.org/data/events/gstreamer-conference/2019/Nirbheek%20Chauhan%20-%20GStreamer,%20Windows%20UWP,%20and%20Firefox%20on%20the%20HoloLens%202.pdf

So, I thought of asking you a probably a dumb question. Would adding TURN to libdatachannel and libjuice be hard for me to do? Would you recommend me to try implementing it? I thought it might be a nice practice to better understand networking under the hood if you find it doable (and would be incredibly enjoyable compared to porting glib to UWP).

paullouisageneau:
I perfectly understand the issue, actually that's why I wrote libjuice in the first place!

First, note that even if libjuice does not support gathering relay candidates, it can still connect to relay candidates from the other peer, therefore you should be able to have a relayed connection between a browser and libdatachannel+libjuice provided a TURN server is set on the browser.

To answer your question, there is basically nothing to do in libdatachannel, only the support to add in libjuice by extending the STUN protocol implementation. It's quite a big task actually because TURN needs a session to be maintained, and I don't know if you are familiar with plain C. This is also something I have been wanting to do for some time, so if you are not in a hurry I might find time to do it soon-ish.

hanseuljun:
Thanks again for your kind response. Feels like you are living in my future!

Thanks for introducing me the bypass, though unfortunately that would not be my case since I am using libdatachannel from both ends...

About the TURN implementation, as I would not call myself familiar with plain C and is very likely to contaminate your repository, I can definitely wait! I'll tell my colleague to do something with his router for now haha

paullouisageneau:
PR https://github.com/paullouisageneau/libdatachannel/pull/287 adds TURN support with libjuice.

This is not completely finished and requires further testing, but it works good enough if you want to test.

hanseuljun:
[...]

PR #287, opened 2020-12-08
TURN support for libjuice

paullouisageneau:
This PR introduces TURN support for libjuice.

Implements https://github.com/paullouisageneau/libdatachannel/issues/281

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ set(LIBDATACHANNEL_HEADERS
 set(TESTS_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/test/main.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/test/connectivity.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/test/turn_connectivity.cpp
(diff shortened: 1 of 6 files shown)


hanseuljun:
This basically works! With providing a turn server as an ICE server, I got a relay candidate back!

One thing I am not sure I not doing wrong is, when providing a STUN server and a TURN server at the same time (2 ICE servers), libdatachannel did not return a relay candidate. But maybe providing a TURN server is the standard practice? (The STUN and TURN server were not from different machines, it was like stun:coturn.hanseul.com and turn:user:password@coturn.hanseul.com with a towards a ubuntu server running coturn.)

paullouisageneau:
> One thing I am not sure I not doing wrong is, when providing a STUN server and a TURN server at the same time (2 ICE servers), libdatachannel did not return a relay candidate. But maybe providing a TURN server is the standard practice? (The STUN and TURN server were not from different machines, it was like stun:coturn.hanseul.com and turn:user:[password@coturn.hanseul.com](mailto:password@coturn.hanseul.com) with a towards a ubuntu server running coturn.)

Good catch, it came from a mistake checking whether gathering is done. It should be good now!

hanseuljun:
Awesome! The thing I mentioned is fixed as you said!

Issue #296, opened 2020-12-19
Too Much Logging


stazio:
When entering failure states, there is too much logging happening and causing a further degredation of the application. Even if ideally the developer was to create a perfect application, the peer, whether malicious or not, can still trigger a lot of logs at reasonable log levels such as WARNING. Ideally, there should be no logging performed on the processing thread.
I was thinking about going through and changing all of logs messages that are present in the real-time paths and making them at least `PLOG_DEBUG`. I wanted to get your opinion first if you had a better solution.

paullouisageneau:
What failure states are you speaking about? I guess you are refering to situations where there is an SRTP issue and each message triggers a warning, right?

I think a good number of them can be changed to PLOG_DEBUG, another possibility for the most critical warnings is to have a local define that would call PLOG_WARNING only for the first call for a given instance, then PLOG_DEBUG (this can be achieved easily with `std::call_once`).

stazio:
I like the way Janus does it where they count the number of errors and report it at an interval.

paullouisageneau:
It sounds good too. How is this achieved, I guess the first line is printed, and the following identical lines are cached for a given duration, right?

stazio:
Just looked at the code: `JANUS_LOG(LOG_ERR, "[%"SCNu64"] Got %d SRTP/SRTCP errors in the last few seconds (last error: %s)\n", handle->handle_id, handle->srtp_errors_count, janus_srtp_error_str(handle->last_srtp_error));`. It's been a while since I used Janus, must have forgotten how those logs look. Apparently when debugging it prints every error.

I'd rather have those print on their own thread. Does libdatachannel have a free thread that this could be used.

paullouisageneau:
That could be good. 

You can delegate a task to libdatachannel's threadpool with `ThreadPool::Instance().enqueue()`, but for now it has no support for scheduling a task at a given time and I think that's what you'd want here.

paullouisageneau:
I quickly added a `schedule()` method in https://github.com/paullouisageneau/libdatachannel/pull/300

PR #302, opened 2021-01-04
Refactored runtime errors to be logged periodically

stazio:
I added a new class called `LogCounter` using the scheduling changes in #300. This now causes all common errors (such as replay attacks) to be logged periodically. I used a lot of static globals here which I will argue are okay in this context, but I fully understand how my approach may be discouraged. I concluded that it was rather pointless to have these initialized in the class because you can't differentiate between multiple peer connections anyways.

### DIFF
--- a/src/datachannel.cpp
+++ b/src/datachannel.cpp
@@ -20,6 +20,7 @@
 #include "include.hpp"
 #include "peerconnection.hpp"
 #include "sctptransport.hpp"
+#include "logcounter.hpp"
 
 #ifdef _WIN32
 #include <winsock2.h>
@@ -27,6 +28,7 @@
 #include <arpa/inet.h>
 #endif
 
+rtc::LogCounter COUNTER_USERNEG_OPEN_MESSAGE(plog::warning, "Number of open messages for a user-negotiated DataChannel received");
 namespace rtc {
 
 using std::shared_ptr;
@@ -170,7 +172,7 @@ void DataChannel::open(shared_ptr<SctpTransport> transport) {
 }
 
 void DataChannel::processOpenMessage(message_ptr) {
-	PLOG_WARNING << "Received an open message for a user-negotiated DataChannel, ignoring";
+    COUNTER_USERNEG_OPEN_MESSAGE++;
 }
 
 bool DataChannel::outgoing(message_ptr message) {
--- a/src/datachannel.cpp
+++ b/src/datachannel.cpp
@@ -172,6 +172,7 @@ void DataChannel::open(shared_ptr<SctpTransport> transport) {
 }
 
 void DataChannel::processOpenMessage(message_ptr) {
+    PLOG_DEBUG << "Received an open message for a user-negotiated DataChannel, ignoring";
     COUNTER_USERNEG_OPEN_MESSAGE++;
 }
 
--- a/src/dtlssrtptransport.cpp
+++ b/src/dtlssrtptransport.cpp
@@ -18,6 +18,7 @@
 
 #include "dtlssrtptransport.hpp"
 #include "tls.hpp"
+#include "logcounter.hpp"
 
 #if RTC_ENABLE_MEDIA
 
@@ -27,6 +28,16 @@
 using std::shared_ptr;
 using std::to_integer;
 using std::to_string;
+
+static rtc::LogCounter COUNTER_MEDIA_SENT_EARLY(plog::warning, "Number of RTP packets sent before keys have been derived");
+static rtc::LogCounter COUNTER_MEDIA_TRUNCATED(plog::warning, "Number of truncated SRT(C)P packets received");
+static rtc::LogCounter COUNTER_UNKNOWN_PACKET_TYPE(plog::warning, "Number of RTP packets received with an unknown packet type");
+static rtc::LogCounter COUNTER_SRTCP_REPLAY(plog::warning, "Number of SRTCP replay packets received");
+static rtc::LogCounter COUNTER_SRTCP_AUTH_FAIL(plog::warning, "Number of SRTCP packets received that failed authentication checks");
+static rtc::LogCounter COUNTER_SRTCP_FAIL(plog::warning, "Number of SRTCP packets received that had an unknown libSRTP failure");
+static rtc::LogCounter COUNTER_SRTP_REPLAY(plog::warning, "Number of SRTP replay packets received");
+static rtc::LogCounter COUNTER_SRTP_AUTH_FAIL(plog::warning, "Number of SRTP packets received that failed authentication checks");
+static rtc::LogCounter COUNTER_SRTP_FAIL(plog::warning, "Number of SRTP packets received that had an unknown libSRTP failure");
 
 namespace rtc {
 
@@ -73,13 +84,14 @@ DtlsSrtpTransport::~DtlsSrtpTransport() {
 	srtp_dealloc(mSrtpOut);
 }
 
+
 bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 	std::lock_guard lock(sendMutex);
 	if (!message)
 		return false;
 
 	if (!mInitDone) {
-		PLOG_WARNING << "SRTP media sent before keys are derived";
+		COUNTER_MEDIA_SENT_EARLY++;
 		return false;
 	}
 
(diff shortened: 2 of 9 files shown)


src/logcounter.cpp
```diff
@@ -0,0 +1,46 @@
...
+            }
+            PLOG(severity) << text << ": " << countCopy << " (over " << std::chrono::duration_cast<std::chrono::seconds>(duration).count() << " seconds)";
+        });
+    }
+    return *this;
+}
+
+rtc::LogCounter::~LogCounter() {
+    if (future) {
+        future->wait();
```
stazio, line 44: I know this isn't correct but I don't know how this should be done.

stazio:
This is the solution for #296.

src/datachannel.cpp
```diff
@@ -170,7 +172,7 @@ void DataChannel::open(shared_ptr<SctpTransport> transport) {
 }
 
 void DataChannel::processOpenMessage(message_ptr) {
-	PLOG_WARNING << "Received an open message for a user-negotiated DataChannel, ignoring";
```
paullouisageneau, line 173: I think it would be good to keep the log line as debug.

stazio: Instead of a counter?

src/dtlssrtptransport.cpp
```diff
@@ -73,13 +84,14 @@ DtlsSrtpTransport::~DtlsSrtpTransport() {
...
 }
 
+
 bool DtlsSrtpTransport::sendMedia(message_ptr message) {
 	std::lock_guard lock(sendMutex);
 	if (!message)
 		return false;
 
 	if (!mInitDone) {
-		PLOG_WARNING << "SRTP media sent before keys are derived";
```
paullouisageneau, line 82: This warning doesn't need a counter, and could actually be promoted to an error.

It shouldn't happen as tracks should not have a transport before they are open by `PeerConnection`, which happens after the init here.

src/dtlssrtptransport.cpp
```diff
@@ -178,11 +192,13 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
 			PLOG_VERBOSE << "Incoming SRTCP packet, size=" << size;
 			if (srtp_err_status_t err = srtp_unprotect_rtcp(mSrtpIn, message->data(), &size)) {
 				if (err == srtp_err_status_replay_fail)
-					PLOG_WARNING << "Incoming SRTCP packet is a replay";
+				    COUNTER_SRTCP_REPLAY++;
 				else if (err == srtp_err_status_auth_fail)
-					PLOG_WARNING << "Incoming SRTCP packet failed authentication check";
-				else
-					PLOG_WARNING << "SRTCP unprotect error, status=" << err;
+                    COUNTER_SRTCP_AUTH_FAIL++;
```
paullouisageneau, line 197: Why not keeping the log line as verbose?

src/dtlssrtptransport.cpp
```diff
@@ -193,13 +209,14 @@ void DtlsSrtpTransport::incoming(message_ptr message) {
...
-					PLOG_WARNING << "Incoming SRTP packet is a replay";
-				else if (err == srtp_err_status_auth_fail)
-					PLOG_WARNING << "Incoming SRTP packet failed authentication check";
-				else
-					PLOG_WARNING << "SRTP unprotect error, status=" << err;
-
+                if (err == srtp_err_status_replay_fail)
+                    COUNTER_SRTP_REPLAY++;
+                else if (err == srtp_err_status_auth_fail)
+                    COUNTER_SRTP_AUTH_FAIL++;
```
paullouisageneau, line 215: Same here.

src/logcounter.cpp
paullouisageneau: You should not wait on the future and actually do nothing, since you don't know whether the thread pool will have run the task. If it has not this will deadlock.

src/datachannel.cpp
paullouisageneau: No, in addition to it. This message can only be sent over DTLS by a trusted peer anyway.

src/logcounter.cpp
stazio: Is there a possibility for a race condition here where the counter would be destroyed before the thread pool finishes doing it's thing?

paullouisageneau: Definitely, because both are global objects, so you have no control over the destruction order.

stazio: So what exactly should be happening here? Isn't there the potential for a segfault if the ThreadPool triggers the log counter after the log counter was destroyed?

src/logcounter.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace rtc {
+class LogCounter {
+private:
+    plog::Severity severity;
+    std::string text;
+    std::chrono::steady_clock::duration duration;
+
+    int count = 0;
+    std::mutex mutex;
+    std::optional<invoke_future_t<void (*)()>> future;
```
paullouisageneau, line 34: You probably don't need to keep the future anymore.

src/logcounter.cpp
```diff
@@ -0,0 +1,40 @@
...
+rtc::LogCounter& rtc::LogCounter::operator++(int) {
+    std::lock_guard lock(mutex);
+    count++;
+    if (!future) {
+        future = ThreadPool::Instance().schedule(duration, [this]() {
+            int countCopy;
+            {
+                std::lock_guard lock(mutex);
+                countCopy = count;
+                count = 0;
```
paullouisageneau, line 33: If you make `count` a `<std::atomic<int>`, you can replace this by an atomic exchange and you don't need the mutex anymore.

src/logcounter.cpp
paullouisageneau: I think the easiest way to keep the `LogCounter` instance globals is to store the state of `LogCounter` in a structure held by a `shared_ptr`, and pass a `weak_ptr` to the task.

stazio: What about using `std::enable_shared_from_this` like I just did?

src/logcounter.cpp
stazio: Wouldn't it be possible for there to be a race between incrementing and testing the count? Based on your previous recommendation, and depending on the implementation, I could skip an enqueue or queue multiple times at the same time.

src/logcounter.cpp
paullouisageneau: You could do that, however `enable_shared_from_this` would require to create every instance of `LogCounter` with a `shared_ptr`, which is a bit cumbersome.

src/logcounter.cpp
paullouisageneau: If `count` is atomic, the increment operators are also atomic, so `if(count++ == 0)` increments the count and tests it without a possible race condition.

stazio: Thanks for teaching my modern C++ haha

src/logcounter.cpp
stazio: So I did it, but I feel like it might be hacky.

src/logcounter.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include "logcounter.hpp"
+
+rtc::LogCounter::LogCounter(plog::Severity severity, const std::string &text, std::chrono::seconds duration) :
+        mSeverity(severity), mText(text), mDuration(duration), mIsValidMutex(std::make_shared<std::mutex>()), mIsValid(std::make_shared<bool>(true)) {}
+
+rtc::LogCounter& rtc::LogCounter::operator++(int) {
+    if (mCount++ == 1) {
```
paullouisageneau, line 25: `mCount++` returns the previous value, so I think it should be `mCount++ == 0`.

src/logcounter.hpp
```diff
@@ -0,0 +1,47 @@
...
+class LogCounter {
+private:
+    plog::Severity mSeverity;
+    std::string mText;
+    std::chrono::steady_clock::duration mDuration;
+
+    std::atomic<int> mCount = 0;
+
+    std::shared_ptr<std::mutex> mIsValidMutex;
+    std::shared_ptr<bool> mIsValid;
```
paullouisageneau, line 35: I think this works, however you don't actually need the mutex since `mCount` is atomic.

Just a suggestion: you could simplify by putting `mSeverity`, `mText`, `mDuration`, and `mCount` directly in a structure held by a `shared_ptr`. Then you only need to pass a `weak_ptr` to the task, the validity will be given by whether the `weak_ptr` can be locked or not.

stazio: :+1:

paullouisageneau (approved):
It looks great, thanks!

PR #305, opened 2021-01-06
Added support for filtering candidates

stazio:
This PR compliments  [libjuice#70](https://github.com/paullouisageneau/libjuice/pull/70) to add support for filtering callbacks.

### DIFF
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -114,6 +114,7 @@ public:
 	void onDataChannel(std::function<void(std::shared_ptr<DataChannel> dataChannel)> callback);
 	void onLocalDescription(std::function<void(Description description)> callback);
 	void onLocalCandidate(std::function<void(Candidate candidate)> callback);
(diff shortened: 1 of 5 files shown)


stazio:
Where would you put the libnice guard? I was gonna put it around the constructor in `icetransport.hpp`.

paullouisageneau:
Sorry but I don't really get the point of this for now. How is this better than filtering local candidates directly before sending them?

For instance by simply doing:
```
pc->onLocalCandidate([](Candidate candidate) {
    if(candidate.type() == Candidate::Type::Host)
        return;
    [...]
});
```
or 
```
auto desc = pc.localDescription();
auto cands = desc.extractCandidates();
candidates.erase(std::remove_if(cands.begin(), cands.end(),
    [](const Candidate &c) { return c.type() == Candidate::Type::Host; }), cands.end());
desc.addCandidates(std::move(cands));
```

I especially don't see the point of filtering the candidates down to libjuice internal representation. Local candidates being added to the local description internally should not slow down the discovery process with libjuice (it might with libnice).

PR #291, opened 2020-12-10
Streaming H264 and opus samples

LeoNavel:
This feature adds API for sending H264 and opus samples over RTP and example project.

New API for:

- setting all RTP properties,
- setting all RTCP_SR properties,
- creating RTCP SDES and its chunks,
- RTP packetization,
- RTP packetization of H264 samples (supports single nal unit and nal unit fragmentation),
- RTP packetization of opus samples,
- recording RTP stats and sending RTCP sender report,
- handlers for automatic H264 and opus packetization and for sending RTCP sender reports.

Example project contains:

- source codes for server app to stream video and audio to multiple clients,
- H264 samples (including original video track),
- opus samples (including original audio track),
- python script to generate H264 samples from video,
- python script to generate opus samples from audio,
- html page for client.

PR #306, opened 2021-01-13
C api for h264/opus streaming

LeoNavel:
This feature adds C API for sending H264 and opus samples over RTP using API added in #291.


include/rtc/rtc.h
```diff
@@ -77,6 +77,28 @@ typedef enum { // Don't change, it must match plog severity
 	RTC_LOG_VERBOSE = 6
 } rtcLogLevel;
 
+#if RTC_ENABLE_MEDIA
+
+typedef enum {
+    // video
+    RTC_H264,
```
paullouisageneau, line 84: Other enums have an explicit type in their name so this should be `RTC_CODEC_H264` for consistency.

include/rtc/rtc.h
```diff
@@ -77,6 +77,28 @@ typedef enum { // Don't change, it must match plog severity
...
+    RTC_VP8,
+    RTC_VP9,
+
+    // audio
+    RTC_OPUS
+} rtcCodec;
+
+typedef enum {
+    RTC_SENDONLY,
```
paullouisageneau, line 93: Same, this should be `RTC_DIRECTION_SENDONLY` for consistency.

include/rtc/rtc.h
```diff
@@ -77,6 +77,28 @@ typedef enum { // Don't change, it must match plog severity
...
+    RTC_RECVONLY,
+    RTC_SENDRECV,
+    RTC_INACTIVE,
+    RTC_UNKNOWN
```
paullouisageneau, line 97: You could put `UNKNOWN` as the first value so it'll map to zero.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+/// @param pc Peer connection id
+/// @param codec Codec
+/// @param payloadType Payload type
+/// @param ssrc SSRC
+/// @param _mid MID (default: video|audio)
+/// @param _direction Direction (default RTC_SENDONLY)
+/// @param _name Name (optional)
+/// @param _msid MSID (optional)
+/// @returns Track id
+RTC_EXPORT int rtcAddTrackEx(int pc, rtcCodec codec, int payloadType, uint32_t ssrc, const char *_mid, rtcDirection * _direction, const char *_name, const char *_msid);
```
paullouisageneau, line 215: I don't really like introducing a pointer for an optional value, it's confusing and rather cumbersome to call as you need to introduce local variables. `rtcDirection * _direction` should be `rtcDirection _direction` and the user could pass `RTC_DIRECTION_UNKNOWN` i.e; `0` to get the default behavior.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Set H264PacketizationHandler for track
+/// @param tr Track id
+/// @param ssrc SSRC
+/// @param cname CName
+/// @param payloadType Payload Type
+/// @param clockRate Clock rate
+/// @param _sequenceNumber Sequence number (default: pseudo random value)
+/// @param _timestamp Timestamp (default: pseudo random value)
+RTC_EXPORT int rtcSetH264PacketizationHandler(int tr, uint32_t ssrc, const char * cname, uint8_t payloadType, uint32_t clockRate, uint16_t * _sequenceNumber, uint32_t * _timestamp);
```
paullouisageneau, line 225: Same here for `uint16_t * _sequenceNumber` and `uint32_t * _timestamp`, could `(uint16_t)-1` be interpreted as a random value? If so, you could define `RTC_RANDOM_SEQUENCE_NUMBER` and `RTC_RANDOM_TIMESTAMP`.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Set OpusPacketizationHandler for track
+/// @param tr Track id
+/// @param ssrc SSRC
+/// @param cname CName
+/// @param payloadType Payload Type
+/// @param clockRate Clock rate
+/// @param _sequenceNumber Sequence number (default: pseudo random value)
+/// @param _timestamp Timestamp (default: pseudo random value)
+RTC_EXPORT int rtcSetOpusPacketizationHandler(int tr, uint32_t ssrc, const char * cname, uint8_t payloadType, uint32_t clockRate, uint16_t * _sequenceNumber, uint32_t * _timestamp);
```
paullouisageneau, line 235: Same here for `uint16_t * _sequenceNumber` and `uint32_t * _timestamp`.

paullouisageneau (commented):
Thank you for adding thinks!

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+
+/// Set onOpen callback for track with given id
+/// @param id Track identifier
+/// @param cb OnOpen callback
+int rtcSetTrackOpenCallback(int id, rtcOpenCallbackFunc cb);
```
paullouisageneau, line 241: This dedicated function is redundant as `rtcSetOpenCallback()` should work for tracks.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Set onClosed callback for track with given id
+/// @param id Track identifier
+/// @param cb OnClosed callback
+int rtcSetTrackClosedCallback(int id, rtcClosedCallbackFunc cb);
```
paullouisageneau, line 246: Same, it is redundant as `rtcSetClosedCallback()` should work for tracks.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Set start time for RTP stream
+/// @param startTime_s Start time in seconds
+/// @param timeIntervalSince1970 Set true if `startTime_s` is time interval since 1970, false if `startTime_s` is time interval since 1900
+/// @param _timestamp Start timestamp (default: current timestamp)
+int rtcSetRTPConfigurationStartTime(int id, double startTime_s, bool timeIntervalSince1970, const uint32_t * _timestamp);
```
paullouisageneau, line 252: It should be written `Rtp` instead of `RTP` for consistency with the naming convention. Also, `id` should be `tr` to reflect that it must be a track.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Start stats recording for RTCP Sender Reporter
+/// @param id Track identifier
+int rtcStartRTCPSenderReporterRecording(int id);
```
paullouisageneau, line 256: Same, it should be written `Rtcp` instead of `RTCP` for consistency with the naming convention.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+/// Set `NeedsToReport` flag in RTCPSenderReportable handler identified by given track id
+/// @param id Track id
+int rtcSetNeedsToSendRTCPSR(int id);
+
+/// Sends data
+/// @param id Track id
+/// @param data Data buffer
+/// @param size Size
+/// @returns Data size
+int rtcTrackSend(int id, const uint8_t *data, int size);
```
paullouisageneau, line 300: `rtcSend()` already does the same.

include/rtc/rtc.h
```diff
@@ -176,6 +199,113 @@ RTC_EXPORT int rtcDeleteTrack(int tr);
...
+
+/// Add callback for received binary data (string data is ignored)
+/// @param tr Track id
+/// @param cb callback
+RTC_EXPORT int rtcSetTrackOnReceiveCallback(int tr, rtcMessageCallbackFunc cb);
```
paullouisageneau, line 305: Same with `rtcSetReceiveCallback()`

src/capi.cpp
```diff
@@ -276,6 +357,15 @@ void rtcInitLogger(rtcLogLevel level, rtcLogCallbackFunc cb) {
 
 void rtcSetUserPointer(int i, void *ptr) { setUserPointer(i, ptr); }
 
+void * rtcGetUserPointer(int i) {
+    auto userPointer = getUserPointer(i);
+    if (userPointer.has_value()) {
+        return userPointer.value();
+    } else {
+        return NULL;
+    }
```
paullouisageneau, line 366: Just a note: this should be shortened to `return getUserPointer(i).value_or(nullptr);`

include/rtc/rtc.h
LeoNavel: Done

include/rtc/rtc.h
LeoNavel: Done

include/rtc/rtc.h
LeoNavel: Done

include/rtc/rtc.h
LeoNavel: In that case I would prefer to remove "optional" and let users to choose value for sequence number and timestamp.

paullouisageneau: OK, works for me!

LeoNavel:
Is there anything else to fix?

paullouisageneau:
No, it looks good, thank you for the fixes!

PR #314, opened 2021-01-25
Added a mutex for RtcpHandler

### LINKED ISSUE #312: mRtcpHandler needs to be atomic
Ran into a segfault today and I think (though can't reproduce) it occurred because I was setting `mRtcpHandler` as it was being used to process an RTCP packet. Am I correct that this is a potential race condition, and how would you want to solve this?

stazio:
Fixes #312. The purpose behind unlocking the lock before using the handler is to prevent a deadlock should the `RtcpHandler` end up sending something again.

### DIFF
--- a/src/track.cpp
+++ b/src/track.cpp
@@ -136,7 +136,7 @@ void Track::incoming(message_ptr message) {
 
 	// Tail drop if queue is full
 	if (mRecvQueue.full()) {
-        COUNTER_QUEUE_FULL++;
+		COUNTER_QUEUE_FULL++;
 		return;
 	}
 
--- a/src/track.cpp
+++ b/src/track.cpp
@@ -62,7 +62,7 @@ bool Track::send(message_variant data) {
 
 	auto message = make_message(std::move(data));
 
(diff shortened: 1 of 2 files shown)


src/track.cpp
```diff
@@ -62,8 +62,11 @@ bool Track::send(message_variant data) {
 
 	auto message = make_message(std::move(data));
 
-	if (mRtcpHandler) {
-		message = mRtcpHandler->outgoing(message);
+    std::shared_lock lock(mRtcpHandlerMutex);
+    if (mRtcpHandler) {
+        auto copy = mRtcpHandler;
+        lock.unlock();
```
paullouisageneau, line 68: This whole repeated block could be made shorter with
'''
if (auto copy = getRtcpHandler()) copy->...
'''

paullouisageneau (approved):
It looks good, but could benefict from better factoring by using `getRtcpHandler` internally. Could you please make the change before I merge?

paullouisageneau (approved):

PR #318, opened 2021-01-28
Chaining of multiple RTP/RTCP handlers

LeoNavel:
This feature adds funcitonality to easily implement handling of RTP/RTCP as standalone parts and chain them together. H264/opus packetization is separated from sending RTCP SR and can be easily chained together. This feature contains RTCP NACK handling too. 

This may help with code reusablity and better scalability of RTP/RTCP handlers. Feature has minimal impact on public API (but still has some, check *examples/streamer*).


include/rtc/h264rtppacketizer.hpp
```diff
@@ -21,22 +21,43 @@
...
 	/// Constructs h264 payload packetizer with given RTP configuration.
 	/// @note RTP configuration is used in packetization process which may change some configuration
 	/// properties such as sequence number.
 	/// @param rtpConfig  RTP configuration
-	H264RtpPacketizer(std::shared_ptr<RtpPacketizationConfig> rtpConfig);
+	/// @param maximumFragmentSize maximum size of one NALU fragment
+	H264RtpPacketizer(std::shared_ptr<RtpPacketizationConfig> rtpConfig,
+					  uint16_t maximumFragmentSize = NalUnits::defaultMaximumFragmentSize);
+
+	ChainedOutgoingProduct processOutgoingBinaryMessage(ChainedMessagesProduct messages, std::optional<message_ptr> control) override;
```
paullouisageneau, line 58: Is there a particular reason for wrapping pointers in optionals? Wouldn't `nullptr` behave just as `nullopt`?

include/rtc/messagehandlerelement.hpp
```diff
@@ -0,0 +1,118 @@
...
+public:
+	MessageHandlerElement();
+
+	/// Creates response to incoming message
+	/// @param messages Current repsonse
+	/// @returns New response
+	std::optional<ChainedOutgoingResponseProduct> processOutgoingResponse(ChainedOutgoingResponseProduct messages);
+
+	// Process incoming and ougoing messages
+	std::optional<message_ptr> formIncomingControlMessage(message_ptr message, std::function<bool (ChainedOutgoingResponseProduct)> send);
```
paullouisageneau, line 79: Same here for pointers in optionals?

include/rtc/include.hpp
```diff
@@ -56,6 +56,7 @@ using std::byte;
 using std::string;
 using std::string_view;
 using binary = std::vector<byte>;
+using binary_ptr = std::shared_ptr<binary>;
```
paullouisageneau, line 59: Just a question: what is the difference of meaning between `message_ptr` and `binary_ptr`?

paullouisageneau (commented):
Thank you, this is great. I mostly have a few remarks and renaming suggestions.

include/rtc/messagehandlerelement.hpp
```diff
@@ -0,0 +1,118 @@
...
+struct RTC_CPP_EXPORT ChainedIncomingControlProduct {
+	ChainedIncomingControlProduct(message_ptr incoming, std::optional<ChainedOutgoingResponseProduct> outgoing = nullopt);
+	const std::optional<message_ptr> incoming;
+	const std::optional<ChainedOutgoingResponseProduct> outgoing;
+};
+
+/// Chainable handler
+class RTC_CPP_EXPORT MessageHandlerElement: public std::enable_shared_from_this<MessageHandlerElement> {
+	std::optional<std::shared_ptr<MessageHandlerElement>> upstream = nullopt;
+	std::optional<std::shared_ptr<MessageHandlerElement>> downstream = nullopt;
```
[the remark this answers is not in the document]
LeoNavel: This is just my personal preference (I can change it if you want). By wrapping pointer in optional I declare that value of pointer won't be `nullptr` so there should not be need for `nullptr` check (I don't like idea where I work with shared pointers and need to think if it can be `nullptr` or not).

include/rtc/include.hpp
LeoNavel: `binary_ptr` is shared pointer to `vector<byte>`, message_ptr is shared pointer to `Message` (subclass of `vector<byte>`), I added `binary_ptr` because I didn't need subclass of `vector<byte> (`Message`).

include/rtc/messagehandlerelement.hpp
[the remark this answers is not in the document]
paullouisageneau: I see! Since in other places `message_ptr` may be null, I think it would be great if you can change to do the same for the sake of consistency.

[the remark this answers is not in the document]
LeoNavel: done

paullouisageneau (approved):
Looking good, thanks!

Issue #310, opened 2021-01-22
Fixing the UWP crashes (continuing #216)


hanseuljun:
Finally I had time to come back! So, which seemingly was a crash (see #216) was neither an actual crash or something caused by a missing dll. It was UWP by default expecting all applications to be GUI and complaining about the libdatachannel applications not having an "App" class with methods like OnLaunched, OnSuspending, and OnNavigationFailed in it!

So, the fix is stating the libdatachannel applications are not GUI but console. When running CMake to create a Visual Studio Solution, CMake auto-generates a package.appxManifest file per executable. To claim the executables are console, these appxManifest files need to be manipulated as below:

1. Add two attributes to the <Package> tag
xmlns:desktop4="http://schemas.microsoft.com/appx/manifest/desktop/windows10/4" 
xmlns:iot2="http://schemas.microsoft.com/appx/manifest/iot/windows10/2"

2. Add four attributes to <Application> tags inside the <Applications> tag.
desktop4:Subsystem="console"
desktop4:SupportsMultipleInstances="true"
iot2:Subsystem="console"
iot2:SupportsMultipleInstances="true"

This information comes from this website (https://docs.microsoft.com/en-us/windows/uwp/launch-resume/console-uwp) and attached is an appxManifest file modified for datachannel-benchmark provided as an example.
[package.appxManifest.txt](https://github.com/paullouisageneau/libdatachannel/files/5858615/package.appxManifest.txt)

The problem here is that I don't think CMake currently supports adding those attributes by a flag or something. So, to fix this UWP crashing issue, an approach would be adding the above attributes via XML manipulation. While I am not even sure XML manipulation can be done inside a CMake script, I will try it soon...

p.s. One positive thing I have learned is that actually your applications were working fine. The crashes happen after completely running main() and does not cause any actual problems. It was just having an error alerting window, which is terribly unappealing, at the end of execution.

paullouisageneau:
Great to have you back on the UWP quest!

> The problem here is that I don't think CMake currently supports adding those attributes by a flag or something. So, to fix this UWP crashing issue, an approach would be adding the above attributes via XML manipulation. While I am not even sure XML manipulation can be done inside a CMake script, I will try it soon...

You can't do XML manipulation, however you could try using [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) to do variable substitution on an XML template.

hanseuljun:
[...]

PR #324, opened 2021-02-03
Uwp fix (addressing #310 with the error dialog popping up)

hanseuljun:
The way to add project.appxManifest was to just add them as files to add_executable...
Learned it from here: https://github.com/Kitware/CMake/blob/master/Tests/VSXaml/CMakeLists.txt

Besides finally implementing this shockingly easy solution, I removed the if guard for not setting OUTPUT_NAME in UWP.

One more thing is, currently, the resource files are at test/uwp-benchmark and test/uwp-tests. While there may be a better location to place them, what I originally wanted to do was putting them in test/uwp/benchmark and test/uwp/tests. This was not possible because tests is inside .gitignore. Is having "tests" inside .gitignore currently needed?


paullouisageneau:
> The way to add project.appxManifest was to just add them as files to add_executable...
> Learned it from here: https://github.com/Kitware/CMake/blob/master/Tests/VSXaml/CMakeLists.txt

> Besides finally implementing this shockingly easy solution, I removed the if guard for not setting OUTPUT_NAME in UWP.

Great news!

> One more thing is, currently, the resource files are at test/uwp-benchmark and test/uwp-tests. While there may be a better location to place them, what I originally wanted to do was putting them in test/uwp/benchmark and test/uwp/tests. This was not possible because tests is inside .gitignore. Is having "tests" inside .gitignore currently needed?

Actually `tests`is actually a typo, it should be `/tests` to ignore only the output of the Makefile in the top-level directory. Please fix it so you can put them in `test/uwp/benchmark` and `test/uwp/tests`.

hanseuljun:
Thanks and done!

paullouisageneau (approved):

Issue #325, opened 2021-02-04
Can not send media via media track


dshurbin:
First of all - I understand, that it is my failure, because example/stream works like charm.
I trying to adopt the code, but can't understand, what is going wrong.
I am opening PeerConnection, creating videotrack, then I am sending videoframes. On the chrome://webrtc-internals I can clearly see the growing graphs showing "frames received", "frames decoded". But at the same level grows graph "frames dropped".
When I tried to send raw h264 packets with "00 00 00 01"  header, I was receiving assertion failure at H264PacketizationHandler::splitMessage. Looking at the sample .h264 files i figured out, that four first bytes are uint32 representing frame length (though I also found description of Separator, where it was clearly said about different starting sequences like (0 0 0 1), (0 0 1) but in the streamer example i didn't found where they are used), so I just replaced starting sequence of the NALU 00 00 00 01 received from camera on uint 32 representing remaining length. Thus I avoid the assertion failure. Frames are flowing, but are dropped.
Can You please direct me where to dig to?

dshurbin:
I tried with the camera: 
Codec - H264-MPEG4-AVC (part 10)(h264) 704x576, fps:10
used payload type 102:
- local SDP:
a=rtpmap:102 H264/90000\r\n
a=rtcp-fb:102 goog-remb\r\n
a=rtcp-fb:102 ccm fir\r\n
a=rtcp-fb:102 nack\r\n
a=rtcp-fb:102 nack pli\r\n
a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f\r\n

- remote (chromium) SDP:
a=rtpmap:102 H264/90000\r\n
a=rtcp-fb:102 goog-remb\r\n
a=rtcp-fb:102transport-cc\r\n
a=rtcp-fb:102 ccm fir\r\n
a=rtcp-fb:102 nack\r\n
a=rtcp-fb:102 nack pli\r\n
a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f\r\n

paullouisageneau:
[...]

PR #328, opened 2021-02-05
Fix: splitting of h264 NALU by start sequence

LeoNavel:
This may fix issue in #325


paullouisageneau (commented):
Thank you for the reactivity on this one!

paullouisageneau (approved):

Issue #334, opened 2021-02-16
Assigning integers to rtcCodec


hanseuljun:
Currently, rtcCodec and rtcDirection at include/rtc/rtc.h do not have integers assigned. Especially for rtcCodec, it seems like this is happening since there are more codecs to come, so it may not ideal to firmly tie them to integers.

This causes an issue when calling C APIs from other languages (e.g., C#). Of course it is possible to call based on inferred integers instead of enums, but this would be fragile to future updates.

How do you think about assigning integers as RTC_CODEC_H264 = 0, RTC_CODEC_VP8 = 1, RTC_CODEC_VP9 = 2, RTC_CODEC_OPUS = 64? I chose 64 for RTC_CODEC_OPUS since it will leave enough space for future video codecs while it is still a small good-looking (for programmers) number. Maybe H265 and AVI can be added to rtcCodec before any implementation of them, but I'm not sure about that.

Finally, though it is not my first time leaving something here, I just wanted to thank again for the clarity of this library. It is really easy to use and has all the things when I am looking for them!

paullouisageneau:
> Currently, rtcCodec and rtcDirection at include/rtc/rtc.h do not have integers assigned. Especially for rtcCodec, it seems like this is happening since there are more codecs to come, so it may not ideal to firmly tie them to integers.

Not exactly, it's actually an oversight because integers values were not needed for them.

> This causes an issue when calling C APIs from other languages (e.g., C#). Of course it is possible to call based on inferred integers instead of enums, but this would be fragile to future updates.
> 
> How do you think about assigning integers as RTC_CODEC_H264 = 0, RTC_CODEC_VP8 = 1, RTC_CODEC_VP9 = 2, RTC_CODEC_OPUS = 64? I chose 64 for RTC_CODEC_OPUS since it will leave enough space for future video codecs while it is still a small good-looking (for programmers) number. 

Sure, it works for me. I think it's even better to choose `RTC_CODEC_OPUS = 128` as it would allow to recognize the audio codecs by the most significant bit when represented on a byte, which could be handy.

> Maybe H265 and AVI can be added to rtcCodec before any implementation of them, but I'm not sure about that.

Don't worry, I think it's better to add codecs as H265 when supported, otherwise it gives the wrong impression when you look at the API definition. However, AVI is not a codec, it's actually a container like MKV (but a quite old one). 

[...]

PR #336, opened 2021-02-16
Assign integers to enums rtcCodec and rtcDirection.

hanseuljun:
Addressing #334. Needed for calling functions with these enums as parameters from other languages.


paullouisageneau (approved):
:rocket:

Issue #331, opened 2021-02-13
datachannel.dll missing for examples on Windows


hanseuljun:
As captured in the following screenshot, by default, the examples are placed separately from datachannel.dll when built, and this leads to an error as below.
![image](https://user-images.githubusercontent.com/5001447/107834543-38ed9080-6d4b-11eb-94d5-ae2d5e725c85.png)

While I can make them work by manually copy pasting the datachannel.dll into the example folders when needed, would you like to make CMake do that automatically? It would require adding something like below to the CMakeLists.txt files of the examples. Also, I'm not sure whether this is happening only on windows.

add_custom_command(TARGET Answerer POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy_if_different
  "$<TARGET_FILE_DIR:datachannel>/datachannel.dll"
  $<TARGET_FILE_DIR:Answerer>
)

paullouisageneau:
That's because ELF executables on Linux have a feature called rpath which allows the linker to set an additional path where to look up for dynamic libraries.

For instance for the streamer example on my machine:
```
paulo@zeus:~/src/libdatachannel/build/examples/streamer(master)$ readelf -d streamer | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [/home/paulo/src/libdatachannel/build]
```

I don't think Windows has such a mechanism so it looks for DLLs only in the current directory, which is a problem if you run an example from a subdirectory. 

Feel free to add Windows-only CMake commands to copy the DLL with each example as you suggest.

hanseuljun:
As far as I know, there is no such functionality of the superior operating systems in Windows... So, I guess I'll have to add that ugly copy command to your cmake files...

PR #337, opened 2021-02-16
Add copy commands that copies datachannel.dll to examples when examples built 

hanseuljun:
This addresses #331. Besides that, I trimmed steamer/CMakeLists.txt.


paullouisageneau (approved):

hanseuljun:
Thanks for the quick merges!

PR #339, opened 2021-02-17
Proper fix for possible deadlock at exit

paullouisageneau:
This PR reverts https://github.com/paullouisageneau/libdatachannel/pull/326 and introduces a proper fix where tasks are guaranteed to be executed.


lerouxrgd:
:rocket: :ship: :100:

PR #350, opened 2021-03-01
Candidate == operator compares also mNode and mService

ilsalvopss:
This takes into account some libertine implementations that don't vary foundation when they should


paullouisageneau (approved):
Thanks!

PR #360, opened 2021-03-07
Change SCTP congestion control to HSTCP

paullouisageneau:


paullouisageneau:
Switching to the default congestion control algorithm instead, see https://github.com/paullouisageneau/libdatachannel/pull/361.

PR #365, opened 2021-03-11
Added a move to the MediaHandler and a null test

stazio:


src/impl/track.cpp
```diff
@@ -182,10 +182,11 @@ bool Track::transportSend([[maybe_unused]] message_ptr message) {
 void Track::setMediaHandler(shared_ptr<MediaHandler> handler) {
 	{
 		std::unique_lock lock(mMutex);
```
stazio, line 184: Should this function `mRtcpHandeler->onOutgoing(nullptr)` before setting the new callback?

paullouisageneau: Yes, I think it could be good to ensure they is only one handler outputting something.

paullouisageneau (approved):

src/impl/track.cpp
```diff
@@ -182,10 +182,11 @@ bool Track::transportSend([[maybe_unused]] message_ptr message) {
...
-		mMediaHandler = handler;
+		mMediaHandler = std::move(handler);
 	}
 
-	handler->onOutgoing(std::bind(&Track::transportSend, this, std::placeholders::_1));
+	if (mMediaHandler)
+		mMediaHandler->onOutgoing(std::bind(&Track::transportSend, this, std::placeholders::_1));
```
paullouisageneau, line 189: This operates on `mMediaHandler` outside the lock and therefore breaks synchonization. Fixed in https://github.com/paullouisageneau/libdatachannel/pull/369.

src/impl/track.cpp
paullouisageneau: Added in https://github.com/paullouisageneau/libdatachannel/pull/369.

PR #367, opened 2021-03-12
Separate node.js signaling server

murat-dogan:
To have signaling server in the web client directory could be confusing.
I just separated the files.


murat-dogan:
Actually we don't need respond code, so I deleted it totally. (tested & working)
But I can add it if you prefer otherwise no problem.

paullouisageneau (approved):
OK thanks!

PR #368, opened 2021-03-12
Client for benchmark purposes

murat-dogan:


.vscode/settings.json
```diff
@@ -0,0 +1,74 @@
+{
```
paullouisageneau, line 1: Is this file really needed?

examples/client-benchmark/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+	dataChannelMap.emplace(id, dc);
+
+	const int duration = params.durationInSec() > 0 ? params.durationInSec() : INT32_MAX;
+	cout << "Becnhmark will run for " << duration << " seconds" << endl;
+
+	int printStatCounter = 0;
+
+	for (int i = 1; i <= duration; ++i) {
+		this_thread::sleep_for(1000ms);
+		cout << "#" << i << " Received: " << receivedSize.load() / 1024 << " KB/s"
```
paullouisageneau, line 230: I would like a benchmarking client to properly compute throughputs. Here, the code relies on `sleep_for` being exact, which is not guaranteed, especially on a CPU under heavy load. Sleep calls are only guaranteed to wait at least for the requested duration, there is no guarantee on the upper bound. You should therefore compute the actual elapsed time like in`test/benchmark.cpp`.

examples/client-benchmark/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+		     << "   Sent: " << sentSize.load() / 1024 << " KB/s"
+		     << "   BufferSize: " << dc->bufferedAmount() << endl;
+		receivedSize = sentSize = 0;
```
paullouisageneau, line 233: It would be better to exchange the values rather than load then set to zero to make the operation atomic. Here some bytes could not be counted if outputting the line is a bit slow.

paullouisageneau (commented):
This is basically a copy-paste of the existing client with modifications. In don't really like maintaining two clients just for this. Do you think it could be possible to integrate the benchmarking mode in the existing client instead?

.vscode/settings.json
murat-dogan: No need for this.

murat-dogan:
I also don't like code repetition. First I thought to combine them but then I decided to operate them.
Because for new comers, for who wants to understand library basics these examples should be a simple as they could be.
Adding all functionalities to the same client could get more complicated than we expected.

I also want to add constant throughput functionality to the new client example.

examples/client-benchmark/main.cpp
murat-dogan: That's right. Fixing it.

examples/client-benchmark/main.cpp
murat-dogan: I know that but it is a not a promised constant throughput. It is just calculating approximate throughput.

murat-sx:
For me it looks good. 
Could you please check it so than we can decide how to continue

murat-dogan:
For me it looks good. 
If you can check it than we can decide how to continue.

murat-dogan:
Also we can move this app to tests folder. Because I think it would be used for benchmark & testing purposes mostly.

examples/client-benchmark/main.cpp
```diff
@@ -0,0 +1,460 @@
...
+				sentSize += tempMessageData.size();
+			}
+		}
+
+		if (printCounter >= STEP_COUNT_FOR_1_SEC) {
+			unsigned long _receivedSize = receivedSize.exchange(0);
+			unsigned long _sentSize = sentSize.exchange(0);
+			float dataCountMultiplier =
+			    duration_cast<milliseconds>((steady_clock::now() - printTime)).count() /
+			    (1000.0); // sec
```
paullouisageneau, line 287: It looks quite convoluted to compute this `dataCountMultiplier`. Why not compute an actual `elaspedTime` of type `seconds` and just divide the size by it? You could do the same in the sending thread.

murat-dogan: Thanks for the hint.

examples/client-benchmark/main.cpp
```diff
@@ -388,14 +385,14 @@ shared_ptr<PeerConnection> createPeerConnection(const Configuration &config,
 						break;
 
 					try {
-						float dataCountMultiplier =
-						    duration_cast<microseconds>((steady_clock::now() - stepTime)).count() /
-						    (1000.0 * stepDurationInMs);
+						const double elapsedTimeInMs =
+						    std::chrono::duration<double>(steady_clock::now() - stepTime).count() *
```
paullouisageneau, line 389: Here it should be `std::chrono::duration<milliseconds>`, otherwise you are not sure about the unit.

murat-dogan: Here I need duration as double. `std::chrono::duration<double>` is in seconds so I am converting to ms by multipling with 1000. Am I missing something?

paullouisageneau: Oh OK, my bad. Could you just change the variable name to `elapsedTimeInSecs` to reflect it's now in seconds? After that it looks good.

murat-dogan: Actually it is ms. I moved multiplication with 1000 to lower line for you :)
So now it is `elapsedTimeInSecs`.

paullouisageneau (approved):

PR #379, opened 2021-03-23
Supporting Examples in UWP 

hanseuljun:
This PR applies the solution described in #310 to the examples, removing the if guard for examples in UWP!


paullouisageneau (approved):
Excellent, thanks!

PR #398, opened 2021-04-10
Add Multiple Data Channel Option

murat-dogan:
It will be nice to be able to create multiple data-channel on app. 

I am also investigating https://github.com/streamr-dev/node-datachannel-performance-diagnostics/issues/1#issuecomment-816651025

### DIFF
--- a/examples/client-benchmark/main.cpp
+++ b/examples/client-benchmark/main.cpp
@@ -212,7 +212,7 @@ int main(int argc, char **argv) try {
 				try {
 					while (dcLocked->bufferedAmount() <= bufferSize) {
 						dcLocked->send(messageData);
-						sentSizeMap[label] = sentSizeMap.at(label) + messageData.size();
+						sentSizeMap.at(label) += messageData.size();
 					}
 				} catch (const std::exception &e) {
 					std::cout << "Send failed: " << e.what() << std::endl;
@@ -235,7 +235,7 @@ int main(int argc, char **argv) try {
 			try {
 				while (dcLocked->isOpen() && dcLocked->bufferedAmount() <= bufferSize) {
 					dcLocked->send(messageData);
-					sentSizeMap[label] = sentSizeMap.at(label) + messageData.size();
+					sentSizeMap.at(label) += messageData.size();
 				}
 			} catch (const std::exception &e) {
 				std::cout << "Send failed: " << e.what() << std::endl;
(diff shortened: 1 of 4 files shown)


examples/client-benchmark/main.cpp
```diff
@@ -61,7 +62,9 @@ string randomId(size_t length);
 // Benchmark
 const size_t messageSize = 65535;
 binary messageData(messageSize);
-atomic<size_t> receivedSize = 0, sentSize = 0;
+unordered_map<string, atomic<size_t>> receivedSizeMap;
+unordered_map<string, atomic<size_t>> sentSizeMap;
+// atomic<size_t> receivedSize = 0, sentSize = 0;
```
paullouisageneau, line 67: I think you can clean this up now.

examples/client-benchmark/main.cpp
```diff
@@ -188,63 +191,67 @@ int main(int argc, char **argv) try {
...
-		if (auto dcLocked = wdc.lock()) {
-			try {
-				while (dcLocked->bufferedAmount() <= bufferSize) {
-					dcLocked->send(messageData);
-					sentSize += messageData.size();
+			if (auto dcLocked = wdc.lock()) {
+				try {
+					while (dcLocked->bufferedAmount() <= bufferSize) {
+						dcLocked->send(messageData);
+						sentSizeMap[label] += messageData.size();
```
paullouisageneau, line 216: Operator [] is basically an insert, which is not thread-safe. Sure, here you took care of emplacing elements before, so it should work in practice, however I think it would still be cleaner to access the unordered map in a read-only with `at`.

murat-dogan: I am not sure If I got you correctly. Normally online one thread changes one map value, and this should be safe.
Anyway I changed the logic to;
`sentSizeMap[label] = sentSizeMap.at(label) + messageData.size();`

paullouisageneau: My point is the operator [] is technically an insert, so it would be more elegant to use the read-only method `at()` to retrieve the reference on the atomic like this:
```
sentMapSize.at(label)+= value;
```

murat-dogan: Ok

paullouisageneau (approved):
Great, thanks.

PR #407, opened 2021-04-14
support finding openssl from homebrew on M1 Macs

arvidn:
homebrew changed its default installation paths for M1 Macs


paullouisageneau:
Thank you, I'll target the v0.12 branch and bump the version in libtorrent once merged.

Issue #388, opened 2021-04-04
Little CMake project improvements suggestion.


Nemirtingas:
Hi, would you like to add some CMake tweaks to make it easier to port to vcpkg ?
Here is the current patch that needs to be applied to the project to make it work: https://github.com/microsoft/vcpkg/blob/master/ports/libdatachannel/fix-for-vcpkg.patch

Right now, I think those changes might be nice for everyone using CMake and not only vcpkg:
This adds an include interface, when building you can set an absolute path, but when installing, cmake requires it to be relative.
https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/libdatachannel/fix-for-vcpkg.patch#L43

This add an export target, so you don't need to search for the library with cmake as its now aware of what include/library you export
https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/libdatachannel/fix-for-vcpkg.patch#L81-L85
https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/libdatachannel/fix-for-vcpkg.patch#L93-L97

Tho the exported library will be LibDatachannel::datachannel, instead of your current alias LibDatachannel::LibDatachannel., this might be configurable somewhere.

paullouisageneau:
Sure, how does this work with plog through? I see you removed plog as a library to replace it with a manual include directory. Does it mean it must be available as an external library?

Nemirtingas:
Yup, I removed plog and usrsctp as this can be built by vcpkg and I add plog and usrsctp as dependencies, so you don't have to dowload em separatly anymore, they are used if they are already built or built if missing.

paullouisageneau:
> Thats because plog is header only (https://github.com/SergiusTheBest/plog) and doesn't export its configuration to cmake like I did when porting your libs to vcpkg (the export target), and to find it automatically, I needed to tell cmake where to look.

OK, then how would it work with the export target? plog is a public dependency so the headers need to be installed. I guess it would require adding an export target for plog too, right?

Nemirtingas:
Best would be that plog exports its interface library yes, like nlohmann's json did here: https://github.com/nlohmann/json/blob/develop/CMakeLists.txt#L164-L182 but since plog doesn't, I needed to find the include path.

paullouisageneau:
[...]

PR #420, opened 2021-05-13
Fix cmake export file name

wasphin:
Set the export file name explicitly.  See discussion at https://github.com/paullouisageneau/libdatachannel/issues/388#issuecomment-840221093.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -327,7 +327,7 @@ if(WARNINGS_AS_ERRORS)
 	endif()
 endif()
 
-install(TARGETS datachannel EXPORT datachannel-export
+install(TARGETS datachannel EXPORT libdatachannel-config
 	RUNTIME DESTINATION bin
 	LIBRARY DESTINATION lib
 	ARCHIVE DESTINATION lib
@@ -338,7 +338,7 @@ install(FILES ${LIBDATACHANNEL_HEADERS}
 )
 
 install(
-  EXPORT datachannel-export
+  EXPORT libdatachannel-config
   NAMESPACE LibDatachannel::
   DESTINATION share/cmake/libdatachannel
 )


CMakeLists.txt
```diff
@@ -340,6 +340,7 @@ install(FILES ${LIBDATACHANNEL_HEADERS}
 install(
   EXPORT datachannel-export
   NAMESPACE LibDatachannel::
+  FILE libdatachannel-config.cmake
```
paullouisageneau, line 343: It should be `datachannel-config.cmake`, right? Rather than adding a `FILE` line, you could simply change `EXPORT datachannel-export` to `EXPORT datachannel-config`.

wasphin: > It should be `datachannel-config.cmake`, right?

No, it should be ``libdatachannel-config.cmake``, as we use:
```
find_package(libdatachannel CONFIG)
```
rather than
```
find_package(datachannel CONFIG)
```
to find the ``libdatachannel`` package.

> Rather than adding a `FILE` line, you could simply change the export name from `datachannel-export` to ` datachannel-config`.

Done.

wasphin: Do not merge, ``datachannel-config`` does not work, let me check again.

wasphin: Works now, the export name should indeed be ``libdatachannel-config`` rather than ``datachannel-config``.
Also see patch at [vcpkg](https://github.com/microsoft/vcpkg/blob/7b1016e10ef0434fe7045d320a9ee05b63e0efe9/ports/libdatachannel/fix-for-vcpkg.patch#L73).

paullouisageneau: OK thanks :+1:

paullouisageneau (approved):

PR #431, opened 2021-05-31
Improve capi for ssrc manipulation

LeoNavel:


src/description.cpp
```diff
@@ -875,6 +879,18 @@ void Description::Media::addRTPMap(const Description::Media::RTPMap &map) {
 
 std::vector<uint32_t> Description::Media::getSSRCs() { return mSsrcs; }
 
+std::optional<std::string> Description::Media::getCNameForSsrc(uint32_t ssrc) {
+	for (auto &val : mAttributes) {
+		if (val.find("ssrc:") == 0 && val.find("cname:") != std::string::npos) {
```
paullouisageneau, line 884: Just a note, in a future refactoring, it would be worth properly parsing SSRCs and their attributes and storing them in descriptions rather than relying and string lookups every time.

src/capi.cpp
```diff
@@ -633,6 +677,85 @@ int rtcSetNeedsToSendRtcpSr(int id) {
...
+			auto element = *it;
+			if (lowercased(element.second.format) == codec) {
+				payloadTypes.push_back(element.first);
+			}
+		}
+		return copyAndReturn(payloadTypes, buffer, size);
+	});
+}
+
+int rtcGetSsrcsForTrack(int tr, uint32_t * buffer, int bufferSize) {
```
paullouisageneau, line 697: `bufferSize` is rather ambiguous here. Could you please change `bufferSize` to `count` and make it the number of elements in the array?

paullouisageneau (commented):
Thanks for adding, I've just a few change requests for the C API.

src/description.cpp
LeoNavel: Agree.

src/description.cpp
```diff
@@ -858,7 +860,16 @@ void Description::Media::parseSdpLine(string_view line) {
...
 			// always added
 		} else if (key == "ssrc") {
-			mSsrcs.emplace_back(to_integer<uint32_t>(value));
+			auto ssrc = to_integer<uint32_t>(value);
+			if (!hasSSRC(ssrc)) {
+				mSsrcs.emplace_back(ssrc);
+			}
+            auto cnamePos = value.find("cname:");
+            if (cnamePos != std::string::npos) {
+                auto cname = value.substr(cnamePos);
```
paullouisageneau, line 869: Shouldn't this be `auto cname = value.substr(cnamePos + 6)`?

src/capi.cpp
```diff
@@ -268,6 +268,24 @@ int copyAndReturn(binary b, char *buffer, int size) {
...
 }
 
+template<typename T>
+int copyAndReturn(std::vector<T> b, T *buffer, int size) {
+	if (!buffer)
+		return int(b.size());
+
+	if (size < int(b.size()))
+		return RTC_ERR_TOO_SMALL;
+	memcpy(buffer, b.data(), b.size() * sizeof(*buffer));
```
paullouisageneau, line 278: You could replace this by a more clear `std::copy(b.begin(), b.end(), buffer)`.

paullouisageneau:
Thank you, it looks good to me!

PR #442, opened 2021-06-16
BugFixed: to perfer the specified openssl root path over the system v…

industriousonesoft:


CMakeLists.txt
```diff
@@ -283,6 +283,13 @@ if (USE_GNUTLS)
 	target_link_libraries(datachannel PRIVATE GnuTLS::GnuTLS)
 	target_link_libraries(datachannel-static PRIVATE GnuTLS::GnuTLS)
 else()
+	if(APPLE)
+		# This is a bug in CMake that causes it to prefer the system version over 
+		# the one in the specified ROOT folder
+		set(OPENSSL_ROOT_DIR ${OPENSSL_ROOT_DIR} "/usr/local/Cellar/openssl@1.1/1.1.1g/")
```
paullouisageneau, line 289: I don't really like hardcoding the homebrew version here (additionally, the last version is 1.1.1k). Would there be a way to bypass the issue without setting the version manually?

industriousonesoft: You are right! Let me try it!

PR #443, opened 2021-06-17
MacOS: prefer the specified openssl root folder over the system version

industriousonesoft:


paullouisageneau:
Thanks for the fix.

Issue #474, opened 2021-09-08
Why change export name?


KangLin:
Changed the export name will change the forward compatibility.

https://github.com/paullouisageneau/libdatachannel/blob/a17eea83ab93991c04568c6e9653dfb9e71f92ed/CMakeLists.txt#L354

paullouisageneau:
I changed the export name because it was inconsistent: local export name was `LibDataChannel::LibDataChannel` while after installing it became `LibDataChannel::datachannel`. It's true that it breaks compatibility, that's why it won't be released in 0.14.x.

paullouisageneau:
Thinking about it, it could be exported as an alias with a manually-generated config file.

paullouisageneau:
I added an alias to allow forward compatibility in #475.

paullouisageneau:
I added an alias for backward compatibility in #475.

PR #475, opened 2021-09-08
CMake: Add retro-compatibility target alias and fix config file destination

paullouisageneau:
This PR introduces a CMake config file to add the former installed target name `LibDataChannel::datachannel` as an alias for retro-compatibility (The installed target name was fixed to be `LibDataChannel::LibDataChannel` in  https://github.com/paullouisageneau/libdatachannel/pull/464). It also changes the CMake config file destination to `lib` instead of `share`, as `share` should be reserved to platform-independent macros.

Fixes https://github.com/paullouisageneau/libdatachannel/issues/474

### DIFF
--- a/cmake/LibDataChannelConfig.cmake
+++ b/cmake/LibDataChannelConfig.cmake
@@ -1,0 +1,5 @@
+include("${CMAKE_CURRENT_LIST_DIR}/LibDataChannelTargets.cmake")
+
+# For forward compatibility
+add_library(LibDataChannel::datachannel ALIAS LibDataChannel::LibDataChannel)
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,13 +387,13 @@ install(
 	EXPORT LibDataChannelTargets
 	FILE LibDataChannelTargets.cmake
 	NAMESPACE LibDataChannel::
-	DESTINATION share/cmake/LibDataChannel
+	DESTINATION lib/cmake/LibDataChannel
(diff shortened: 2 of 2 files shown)


cmake/LibDataChannelConfig.cmake
```diff
@@ -0,0 +1,5 @@
+include("${CMAKE_CURRENT_LIST_DIR}/LibDataChannelTargets.cmake")
+
+# For forward compatibility
+add_library(LibDataChannel::datachannel ALIAS LibDataChannel::LibDataChannel)
```
KangLin, line 4: The alias target don't export. so that, the application use find_package() that don't import it!

KangLin (commented):
The alias target don't export. so that, the application use find_package() that don't import it!

cmake/LibDataChannelConfig.cmake
paullouisageneau: Could you describe the failing setup? It works fine on my side after installation with:
```
cmake_minimum_required(VERSION 3.7)
project(test-libdatachannel)

find_package(Threads REQUIRED)
find_package(LibDataChannel REQUIRED)

add_executable(test main.cpp)
set_target_properties(test PROPERTIES CXX_STANDARD 17)
target_link_libraries(test Threads::Threads LibDataChannel::datachannel)
```

KangLin: It is ok!

PR #480, opened 2021-09-16
Is it necessary to set a higher priority for DTLS handshake packet?

industriousonesoft:


paullouisageneau:
Good point, however I would argue CS6 is not the right class here, as it should be reserved to network control traffic. I think it should be considered as low-Latency low-drop traffic and marked AF21 (18), which is also consistent with the recommendation for high-priority data in [RFC 8837](https://datatracker.ietf.org/doc/html/rfc8837#section-5).

industriousonesoft:
@paullouisageneau Thx, Paul, I really appreciate that you have solved the problem confusing me quite a while.

paullouisageneau (approved):

PR #486, opened 2021-09-26
Is it necessary to clear all SSRC attributes when creating reciprocated SDP of Media?

industriousonesoft:
In my options, all the SSRC attributes are individual, so we should reset them when creating reciprocated SDP.

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -690,6 +690,17 @@ Description::Media Description::Media::reciprocate() const {
 		break;
 	}
 
+	// Clear all ssrc attributes as them are individual
+	auto it = mAttributes.begin();
+	while (it != mAttributes.end()) {
+		if (match_prefix(*it, "ssrc:"))
+			it = reciprocated.mAttributes.erase(it);
+		else
+			++it;
+	}
+	reciprocated.mSsrcs.clear();
+	reciprocated.mCNameMap.clear();
+
 	return reciprocated;
 }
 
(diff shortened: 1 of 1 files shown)


paullouisageneau:
Indeed, SSRC attributes are local and shouldn't be reciprocated. Attributes for remote SSRCs must be specified with the `remote-ssrc` SDP attribute.

src/description.cpp
```diff
@@ -690,6 +690,17 @@ Description::Media Description::Media::reciprocate() const {
 		break;
 	}
 
+	// Clear all ssrc attributes as them are individual
+	auto it = mAttributes.begin();
+	while (it != mAttributes.end()) {
+		if (match_prefix(*it, "ssrc:"))
+			it = reciprocated.mAttributes.erase(it);
```
paullouisageneau, line 697: This will probably crash as `it` is an iterator on `mAttributes`, not on `reciprocated.mAttributes`.

src/description.cpp
```diff
@@ -690,6 +690,17 @@ Description::Media Description::Media::reciprocate() const {
 		break;
 	}
 
+	// Clear all ssrc attributes as them are individual
```
paullouisageneau, line 693: Just a typo: should be "they".

src/description.cpp
industriousonesoft: Sorry, I made a stupid mistake.

paullouisageneau: It happens :smile:

paullouisageneau:
Looks good, thanks!

PR #488, opened 2021-09-28
Support usage of openssl as static library files in ios.

hanseuljun:
Thanks a lot for the recent update supporting iOS! As the update allows me to directly use the upstream version again, which I am very happy about, I wanted to suggest this PR.

The current CMakeLists.txt assumes using .dylib files in its APPLE support, while some people (e.g., me) prefer using .a files. This PR adds a way to support the use of .a files.
OPENSSL_USE_STATIC_LIBS as a flag comes from CMake's find_package.
https://cmake.org/cmake/help/v3.6/module/FindOpenSSL.html
Using CMake's flag here, so one day the whole detour with OPENSSL_ROOT_DIR can be gone without causing additional changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,8 +312,16 @@ else()
 		# This is a bug in CMake that causes it to prefer the system version over
 		# the one in the specified ROOT folder
 		if(EXISTS ${OPENSSL_ROOT_DIR})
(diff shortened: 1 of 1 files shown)


paullouisageneau (approved):
LGTM!

PR #490, opened 2021-09-30
Correct the bit mask of 'RC/FMT' field in RTCP header

industriousonesoft:


paullouisageneau (approved):
Oh, good catch, thank you for the fix!

PR #548, opened 2022-02-02
fix position bit for isEnd & reservedBit6

evaldemar:


paullouisageneau:
Thank you for the fix! :eyes: 

For reference: https://www.ietf.org/rfc/rfc3984.html#section-5.8

PR #580, opened 2022-03-24
Fix rtcReceiveMessage() in C api consuming messages when buffer is NULL

DavidEichmann:


paullouisageneau (approved):
Looks good, thank you for the fix and doc update!

szabolcsdombi:
Hi,

This still does not work. I have `libdatachannel[core,ws]:x64-windows -> 0.17.11` installed with vcpkg and it is the "latest" at the moment.

```
int size = 0;
int q1 = rtcReceiveMessage(self->dc, NULL, &size);
int q2 = rtcReceiveMessage(self->dc, temp, &size);
 ```

Both `rtcReceiveMessage` calls return zero, the message is not copied to temp.
Size changes to 8 on the first call and it is changed back to 0 by the second call.

If I set the size to 8 and remove the first `rtcReceiveMessage` the message is received.

paullouisageneau:
@szabolcsdombi Indeed, after testing, the function still does not behave as expected when the buffer is `NULL`. The underlying issue is with the method `DataChannel::peek()` and is fixed by https://github.com/paullouisageneau/libdatachannel/pull/799.

PR #601, opened 2022-04-18
add qt signaling-server

cheungxiongwei:


paullouisageneau:
Thank you for your contribution! 

Could you please format the code with the same style as the rest of the repository (basically LLVM style). There is a `.clang-format` configuration file, just running `clang-format` would suffice.

paullouisageneau (approved):
Looking good, thanks!

Issue #602, opened 2022-04-18
High memory usage in usrsctp


paullouisageneau:
In accordance with the recommendation in [RFC 8831](https://www.rfc-editor.org/rfc/rfc8831.html#section-6.2), libdatachannel currently sets the number of streams negotiated during SCTP association setup to 65535, the maximum number:
https://github.com/paullouisageneau/libdatachannel/blob/cb2d43b091ff6b68761cd67d94008eb158b3afdf/src/impl/sctptransport.cpp#L285

However, usrsctp internally allocates tables to maintain the stream states, and setting the number of stream to 65536 results in huge tables with each peer connection allocating a few MB. The allocations are done at [sctputil.c:1325](https://github.com/sctplab/usrsctp/blob/08c09d3d1d42eeaa6d7fd1ba0f04cb25c615b183/usrsctplib/netinet/sctputil.c#L1325) and [sctp_input.c:376](https://github.com/sctplab/usrsctp/blob/08c09d3d1d42eeaa6d7fd1ba0f04cb25c615b183/usrsctplib/netinet/sctp_input.c#L376).

This is an issue for servers handling hundreds or thousands of connections as it results in a very large memory footprint.

PR #603, opened 2022-04-19
Limit number of streams to prevent high memory usage

paullouisageneau:
This PR limits the number of SCTP streams to 1024 instead of the recommended 65535, as a mitigation to reduce the high memory usage from usrsctp. Since this is already done by browsers ([Chrome historically limits it to 1024](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/media/sctp/sctp_transport_internal.h;drc=cfea2182f8e41c3d8293d4e26b291ef2abf23257;l=46) for instance), it shouldn't be a problem.

**This change reduces the memory usage of PeerConnection instances with Data Channels by a whopping few MBs.**

Fixes https://github.com/paullouisageneau/libdatachannel/issues/602

### DIFF
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -85,6 +85,7 @@ public:
 	optional<Description> remoteDescription() const;
 	optional<string> localAddress() const;
 	optional<string> remoteAddress() const;
+	uint16_t maxDataChannelId() const;
 	bool getSelectedCandidatePair(Candidate *local, Candidate *remote);
 
 	void setLocalDescription(Description::Type type = Description::Type::Unspec);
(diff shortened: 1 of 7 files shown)


Nemirtingas:
Firefox even seems to use max 256 streams: https://github.com/w3c/webrtc-pc/issues/2158#issuecomment-577574316

paullouisageneau:
@Nemirtingas Indeed, however it seems to be a bit more complex as Firefox also supports negotiating more streams up to 2048 if supported by the remote peer, [however the mechanism was broken at some point](https://bugzilla.mozilla.org/show_bug.cgi?id=1562341). Anyway, I think 1024 is a good number for compatibility with both browsers.

Nemirtingas:
Sure, I wasn't saying that you should lower it even more, just put it there for future reference in case someone wants to lower it even more.

PR #612, opened 2022-04-29
URL decode username/password in STUN/TURN URL. Fixes #611.

### LINKED ISSUE #611: No way to have special characters in the TURN server URL username/password in C API
`rtcCreatePeerConnection` only takes in `rtcConfiguration` which in turn takes in URLs for `iceServers`, but it's not possible currently to have characters such as `:` or `@` in the username or password of that URL, as they'll get interpreted as delimiters. The credentials from https://subspace.com/ have `:` in the username so it's currently not possible to use them with libdatachannel.

Would it be possible to URL decode the parsed parts of the URL such that `:` and `@` can be safely supplied? Thanks!

bigfarts:


src/configuration.cpp
```diff
@@ -44,6 +45,23 @@ bool parse_url(const std::string &url, std::vector<std::optional<std::string>> &
...
 }
 
+std::string url_decode(const std::string &str) {
+	static const std::regex r(R"(%[0-9A-Fa-f]{2})", std::regex::extended);
+
+	std::stringstream ss;
+	std::smatch m;
+	for (int i = 0; i < str.length(); ++i) {
+		std::string substr = str.substr(i, 3);
+		if (std::regex_match(substr, m, r)) {
```
paullouisageneau, line 55: A regex seems overkill here and this approach will silently ignore invalid percent-encoded characters, but it works for valid inputs.

paullouisageneau (approved):
LGTM

src/configuration.cpp
```diff
@@ -44,6 +45,23 @@ bool parse_url(const std::string &url, std::vector<std::optional<std::string>> &
 	return true;
 }
 
+std::string url_decode(const std::string &str) {
+	static const std::regex r(R"(%[0-9A-Fa-f]{2})", std::regex::extended);
+
+	std::stringstream ss;
+	std::smatch m;
+	for (int i = 0; i < str.length(); ++i) {
```
paullouisageneau, line 53: `i` should be a `size_t` to prevent a signed/unsigned comparison.

paullouisageneau: Thanks!

PR #627, opened 2022-05-12
Verify pong replies on ping messages

deblauwetom:
Allow the ping interval to be configured for a websocket.
Allow to fail the connection when a configured amount of
pings are not answered with pongs.

### DIFF
--- a/include/rtc/websocket.hpp
+++ b/include/rtc/websocket.hpp
@@ -46,6 +46,8 @@ public:
 		bool disableTlsVerification = false; // if true, don't verify the TLS certificate
 		optional<ProxyServer> proxyServer;   // unsupported for now
 		std::vector<string> protocols;
+		int maxMissedPongsAllowed = 2;                   // -1 is no check
+		std::chrono::milliseconds sendPingInterval = std::chrono::seconds(10); // interval at which to send pings
 	};
 
 	WebSocket();
--- a/src/impl/wstransport.cpp
+++ b/src/impl/wstransport.cpp
@@ -54,14 +54,15 @@ using random_bytes_engine =
 
 WsTransport::WsTransport(variant<shared_ptr<TcpTransport>, shared_ptr<TlsTransport>> lower,
                          shared_ptr<WsHandshake> handshake, message_callback recvCallback,
-                         state_callback stateCallback)
+                         state_callback stateCallback, int maxPongsMissed)
     : Transport(std::visit([](auto l) { return std::static_pointer_cast<Transport>(l); }, lower),
                 std::move(stateCallback)),
       mHandshake(std::move(handshake)),
       mIsClient(
           std::visit(rtc::overloaded{[](shared_ptr<TcpTransport> l) { return l->isActive(); },
                                      [](shared_ptr<TlsTransport> l) { return l->isClient(); }},
-                     lower)) {
+                     lower)),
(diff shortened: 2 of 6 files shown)


include/rtc/websocket.hpp
```diff
@@ -46,6 +46,8 @@ class RTC_CPP_EXPORT WebSocket final : private CheshireCat<impl::WebSocket>, pub
 		bool disableTlsVerification = false; // if true, don't verify the TLS certificate
 		optional<ProxyServer> proxyServer;   // unsupported for now
 		std::vector<string> protocols;
+		int maxMissedPongsAllowed = 2;                   // -1 is no check
```
paullouisageneau, line 49: I prefer using an `optional<int>` to allow disabling, which should be unset by default to keep the current behavior.

src/impl/wstransport.cpp
```diff
@@ -312,11 +313,14 @@ void WsTransport::recvFrame(const Frame &frame) {
 	}
 	case PING: {
 		PLOG_DEBUG << "WebSocket received ping, sending pong";
-		sendFrame({PONG, frame.payload, frame.length, true, mIsClient});
+		if (!sendFrame({PONG, frame.payload, frame.length, true, mIsClient})) {
```
paullouisageneau, line 316: Not need to check here, it will actually throw if the transport fails.

include/rtc/websocket.hpp
```diff
@@ -46,6 +46,8 @@ class RTC_CPP_EXPORT WebSocket final : private CheshireCat<impl::WebSocket>, pub
...
+		std::chrono::milliseconds sendPingInterval = std::chrono::seconds(10); // interval at which to send pings
```
paullouisageneau, line 50: Just a nitpick, but this could be called just `pingInterval`.

src/impl/wstransport.cpp
paullouisageneau, line 316: No need to check here, it should throw if the transport fails.

src/impl/wstransport.hpp
```diff
@@ -71,12 +71,16 @@ class WsTransport final : public Transport {
 	void recvFrame(const Frame &frame);
 	bool sendFrame(const Frame &frame);
 
+	void addOpenPing();
+
 	const shared_ptr<WsHandshake> mHandshake;
 	const bool mIsClient;
+	const int mMaxPongsMissed;
```
paullouisageneau, line 78: It should be `mMaxMissedPongs` for consistency.

include/rtc/websocket.hpp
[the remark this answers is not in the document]
paullouisageneau: Actually, it could be good to rename the setting to something with 0 as the default value in order to expose the same setting in the C API (the default value has to be 0 there). 

I suggest `maxOutstandingPings`, with 1 meaning 0 missed pongs, so 0 can mean disabled in the C API (the phrasing also sounds more common). What do you think?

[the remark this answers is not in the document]
deblauwetom: ok, great idea

so just to be clear, you still want the optional, right? so it becomse `optional<int> maxOutstandingPings`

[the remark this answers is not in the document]
paullouisageneau: Yes, I'd prefer making it optional for clarity and consistency with other configuration settings in (`rtc::Configuration` for instance).

PR #666, opened 2022-07-01
Mark external dependency submodules as shallow

dtzxporter:
This will help speed up cloning the repository. When making libdatachannel a sub module of a rust project, we can't change these dynamically.


paullouisageneau:
Thank you!

PR #668, opened 2022-07-01
Remove Extra Lookup in PeerConnection::forwardMedia()

SE2Dev:
Optimizes media forwarding somewhat.

### DIFF
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -546,57 +546,6 @@ void PeerConnection::forwardMedia(message_ptr message) {
 	}
 }
 
-optional<std::string> PeerConnection::getMidFromSsrc(uint32_t ssrc) {
-	if (auto it = mMidFromSsrc.find(ssrc); it != mMidFromSsrc.end())
-		return it->second;
-
-	{
-		std::lock_guard lock(mRemoteDescriptionMutex);
-		if (!mRemoteDescription)
-			return nullopt;
-
-		for (unsigned int i = 0; i < mRemoteDescription->mediaCount(); ++i) {
-			if (auto found =
-			        std::visit(rtc::overloaded{[&](Description::Application *) -> optional<string> {
-				                                   return std::nullopt;
-			                                   },
-			                                   [&](Description::Media *media) -> optional<string> {
-				                                   return media->hasSSRC(ssrc)
-				                                              ? std::make_optional(media->mid())
-				                                              : nullopt;
-			                                   }},
-			                   mRemoteDescription->media(i))) {
-
-				mMidFromSsrc.emplace(ssrc, *found);
-				return *found;
-			}
-		}
-	}
-	{
-		std::lock_guard lock(mLocalDescriptionMutex);
-		if (!mLocalDescription)
-			return nullopt;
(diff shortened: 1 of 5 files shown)


paullouisageneau (commented):
While it's a good idea to optimize the dispatching by SSRC, I'm pretty sure the proposed changes won't work as-is for all use cases: since `incomingTrack()` is only called in `processLocalDescription()` when generating a local answer description, SSRCs present in a remote answer after a local offer won't be inserted in `mTracksBySsrc`. Have you checked the `media-receiver` example still work with the changes?

SE2Dev:
> While it's a good idea to optimize the dispatching by SSRC, I'm pretty sure the proposed changes won't work as-is for all use cases: since `incomingTrack()` is only called in `processLocalDescription()` when generating a local answer description, SSRCs present in a remote answer after a local offer won't be inserted in `mTracksBySsrc`. Have you checked the `media-receiver` example still work with the changes?

You're right, I checked against `media-receiver` and it doesn't work for that use-case. Do you think populating the map in `PeerConnection::processLocalDescription` and `PeerConnection::processRemoteDescription` would be a better method?

paullouisageneau:
> You're right, I checked against `media-receiver` and it doesn't work for that use-case. Do you think populating the map in `PeerConnection::processLocalDescription` and `PeerConnection::processRemoteDescription` would be a better method? I'm not entirely sure at what point the tracks are actually populated so I'm not really sure what a safe time to set up the mapping would be.

Actually, this is quite independant from tracks creation: for instance, an SSRC will commonly be added in the remote description for an already-existing track. I think updating the map in `PeerConnection::processLocalDescription` (after local tracks are added to the description) and `PeerConnection::processRemoteDescription` would be good.

SE2Dev:
I've re-implemented the changes to take place in `PeerConnection::processLocalDescription` and `PeerConnection::processRemoteDescription`, and checked that it does still work with the `media-receiver` example.

src/impl/peerconnection.cpp
```diff
@@ -1005,6 +953,8 @@ void PeerConnection::processLocalDescription(Description description) {
 
 		mLocalDescription.emplace(description);
 		mLocalDescription->addCandidates(std::move(existingCandidates));
+
+		updateSsrcTrackCache(mLocalDescription.value());
```
paullouisageneau, line 957: You could call it on `description` before the lock, as it would handle potential exceptions more gracefully.

paullouisageneau:
Thank for the changes, I still have a few comments but now the approach looks good.

src/impl/peerconnection.cpp
```diff
@@ -1218,4 +1170,39 @@ void PeerConnection::resetCallbacks() {
 	gatheringStateChangeCallback = nullptr;
 }
 
+void PeerConnection::updateTrackSsrcCache(const Description &description) {
+	std::lock_guard lock(mTracksMutex); // read-only (writing to mTracksBySsrc)
```
paullouisageneau, line 1174: You should remove `read-only` from the comment. Also, you could use `unique_lock` instead of `lock_guard` for clarity.

paullouisageneau (approved):
Looks good, thank you!

PR #690, opened 2022-07-27
Fix race condition during destruction / deletion of PeerConnections.

dtzxporter:
Add mProcessor.join() to impl/PeerConnection::close(). Reset impl/PeerConnection dtor to default.

This fixes a race condition that will occur when the shared_pointer of the impl() PeerConnection has more use_count than 1 at the time of the destruction of the wrapping PeerConnection, causing callbacks to still fire after the wrapper is dropped, which in turn is due to the fact that the impl() PeerConnection's destructor is never called. Since the wrapping PeerConnection is the only way to initialize the impl() PeerConnection, it is safe to only join in close(), which is always called upon wrapping PeerConnection dtor.

### DIFF
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -73,11 +73,6 @@ PeerConnection::PeerConnection(Configuration config_)
 	}
 }
 
-PeerConnection::~PeerConnection() {
-	PLOG_VERBOSE << "Destroying PeerConnection";
-	mProcessor.join();
-}
-
 void PeerConnection::close() {
 	PLOG_VERBOSE << "Closing PeerConnection";
 
@@ -88,6 +83,11 @@ void PeerConnection::close() {
 	mProcessor.enqueue(&PeerConnection::closeTracks, shared_from_this());
 
 	closeTransports();
+
+	// We must join all tasks here, in order to ensure all references to 'this' are freed.
+	// otherwise the destructor will never be called, causing async callbacks to fire after the base
+	// PeerConnection was freed.
+	mProcessor.join();
 }
 
 optional<Description> PeerConnection::localDescription() const {
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -83,11 +83,6 @@ void PeerConnection::close() {
 	mProcessor.enqueue(&PeerConnection::closeTracks, shared_from_this());
 
 	closeTransports();
-
-	// We must join all tasks here, in order to ensure all references to 'this' are freed.
-	// otherwise the destructor will never be called, causing async callbacks to fire after the base
-	// PeerConnection was freed.
-	mProcessor.join();
 }
 
 optional<Description> PeerConnection::localDescription() const {
@@ -1165,6 +1160,11 @@ bool PeerConnection::changeSignalingState(SignalingState newState) {
 	                   signalingStateChangeCallback, newState);
 
 	return true;
+}
+
+void PeerConnection::join() {
+	// Allows the wrapping PC to wait for callbacks to finish in it's dtor.
+	mProcessor.join();
 }
 
 void PeerConnection::resetCallbacks() {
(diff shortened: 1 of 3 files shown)


src/impl/peerconnection.cpp
```diff
@@ -88,6 +83,11 @@ void PeerConnection::close() {
 	mProcessor.enqueue(&PeerConnection::closeTracks, shared_from_this());
 
 	closeTransports();
+
+	// We must join all tasks here, in order to ensure all references to 'this' are freed.
+	// otherwise the destructor will never be called, causing async callbacks to fire after the base
+	// PeerConnection was freed.
+	mProcessor.join();
```
paullouisageneau, line 90: Joining here will make the processor wait for itself and therefore create a deadlock when the user calls `close()` from a callback (for instance on `State::Failed`).

paullouisageneau (changes requested):
Thank you for the fix attempt, however I don't think joining in close is a good idea as it introduces a deadlock if the user calls `close()` from a callback.

> This fixes a race condition that will occur when the shared_pointer of the impl() PeerConnection has more use_count than 1 at the time of the destruction of the wrapping PeerConnection, causing callbacks to still fire after the wrapper is dropped, which in turn is due to the fact that the impl() PeerConnection's destructor is never called. 

This is actually expected, in that case the destructor should be called when the last task finishes. However callbacks can't be called after `close()` as they are reset on close in `resetCallbacks()`.

paullouisageneau:
I'm just seeing now that the signaling state and track callbacks are not reset properly, I fixed it in https://github.com/paullouisageneau/libdatachannel/pull/691. I don't know if it's related.

dtzxporter:
> I'm just seeing now that the signaling state and track callbacks are not reset properly, I fixed it in #691. I don't know if it's related.

These callbacks weren't even being called.

Long running callbacks like onLocalDescription triggered the race.

src/impl/peerconnection.cpp
```diff
@@ -73,11 +73,6 @@ PeerConnection::PeerConnection(Configuration config_)
 	}
 }
 
-PeerConnection::~PeerConnection() {
-	PLOG_VERBOSE << "Destroying PeerConnection";
-	mProcessor.join();
```
paullouisageneau, line 78: I'm afraid joining here may also create a deadlock situation if the user destroys the `PeerConnection` from a callback, either by releasing the last shared_ptr or by calling `rtcDeletePeerConnection` if using the C API.

I think what your are trying to do might not be feasible: you can't wait for the following tasks to finish from one of the tasks as doing so you block the task processor and therefore the following tasks.

src/peerconnection.cpp
```diff
@@ -53,6 +53,10 @@ PeerConnection::~PeerConnection() {
 	} catch (const std::exception &e) {
 		PLOG_ERROR << e.what();
 	}
+
+	// Force callbacks to run in impl() so that the shared references to it will free
+	// before our dtor ends, otherwise it won't be freed until after this dtor runs.
+	impl()->join();
```
paullouisageneau, line 59: I'm afraid joining here may also create a deadlock situation if the user destroys the `PeerConnection` from a callback, either by releasing the last shared_ptr or by calling `rtcDeletePeerConnection` if using the C API.

I think what your are trying to do might not be feasible: you can't wait for the following tasks to finish from one of the tasks as doing so you block the task processor and therefore the following tasks.

dtzxporter: I think reasonably, no one should be calling delete while inside a callback no?

Close makes more sense, but for this library to be even usable, the user must have a shared_ptr or similar wrapping the original PeerConnection which wouldn't be safe to free in a callback regardless.

paullouisageneau: It's actually common to delete the `PeerConnection` from a callback if the user doesn't have their own event loop, irrelevant of wrapping: typically, the user would have a map of connections by some kind of id from which a connection is removed when its state changes to disconnected or closed. It's also necessary for the C API because calling `rtcDeletePeerConnection` is explicitly allowed in the API reference.

dtzxporter:
I am unable to trigger this ATM with the fuzzer. So I will close unless this pops up again.

PR #703, opened 2022-08-15
Ensure PeerConnection::mTracksBySsrc Always Updates

SE2Dev:
The `mTracksBySsrc.emplace()` call in `PeerConnection::updateTrackSsrcCache()` will not overwrite (no longer used) SSRCs if the new description contains tracks with the same SSRCs due to the fact that `std::unordered_map::emplace()` does not overwrite by default. Instead, C++17's `std::unordered_map::insert_or_assign()` can be used with minimal performance impact.


paullouisageneau:
Thanks for the fix!

Issue #730, opened 2022-10-26
`track.isOpen()` is `false` in the track.onOpen() callback


JonasVautherin:
Note: I had asked on [StackOverflow](https://stackoverflow.com/questions/74211320/webrtc-track-onopen-is-called-but-track-is-not-open), but I was told to ask here instead.

Using libdatachannel, I establish a `PeerConnection` between two partners using some out-of-band signaling. I can create `DataChannel`s and send data between them successfully (with `peer_connection->createDataChannel()`).

But I am struggling to do the same with a video track. Here is what I do:

* I create a track from one partner:

```c++
rtc::Description::Video media("myvideo", rtc::Description::Direction::RecvOnly);
media.addH264Codec(96);
media.setBitrate(3000);

auto track = peer_connection->addTrack(media);
peer_connection->setLocalDescription();
```

Note how I call `setLocalDescription()` after `addTrack`, so that libdatachannel will negotiate the track and I don't need to send an SDP out of band (at least that's my understanding).

* From the other partner, I check the `onTrack()` callback:

```c++
peer_connection->onTrack([this](const std::shared_ptr<rtc::Track>& track) {
    track->onClosed([]() {
        std::cout << "onClosed" << std::endl;
    });
    track->onOpen([]() {
        std::cout << "onOpen" << std::endl;

        if (track->isOpen()) {
            std::cout << "track is open" << std::endl;
        } else {
            std::cout << "track is not open" << std::endl;
        }
    });
}
```

What I observe is that `onTrack` is called, and the `track` has the `mid` I set from the sending side ("myvideo"). But right in the `onOpen()` callback, the call to `track->isOpen()` says that "track is __not__ open".

If I try to use the track later (with e.g. `track->send()`), if fails with a SIGABRT:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Track is closed
Signal: SIGABRT (Aborted)
```

So somehow it feels like my track was never properly open in the first place. Though I would expect `track->onOpen()` to be called... when the track is open.

Am I missing something?

Nemirtingas:
This issue looks like this one: #596

JonasVautherin:
Hmm interesting. Though I am using v0.17.10, which should have the fix, shouldn't it? :thinking:

JonasVautherin:
[...]

PR #733, opened 2022-11-01
Add a configuration option to force track allocation

JonasVautherin:
When creating a `PeerConnection` without a track, libdatachannel will not initialize the SRTP transport (as an optimization, since tracks are apparently not going to be used).

But that's a problem if a track is added dynamically later, with `peer_connection->addTrack()`. At the moment, it is just ignored. This PR fixes this behavior in two ways:

* It adds a `Configuration` option to force the SRTP initialization (as a way to say "Don't optimize, I _will_ use tracks").
* It throws an exception if `addTrack()` is called in the situation where it would be ignored.

Note that `Configuration.forceTrackAllocation` defaults to `false`, such that it does not change the current libdatachannel behavior.

This is the first part discussed in https://github.com/paullouisageneau/libdatachannel/issues/730.

### DIFF
--- a/include/rtc/configuration.hpp
+++ b/include/rtc/configuration.hpp
@@ -85,6 +85,7 @@ struct RTC_CPP_EXPORT Configuration {
 	bool enableIceTcp = false;    // libnice only
 	bool enableIceUdpMux = false; // libjuice only
 	bool disableAutoNegotiation = false;
+	bool forceMediaTransport = false;
 
 	// Port range
 	uint16_t portRangeBegin = 1024;
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -758,9 +758,17 @@ void PeerConnection::openTracks() {
 #if RTC_ENABLE_MEDIA
 	if (auto transport = std::atomic_load(&mDtlsTransport)) {
 		auto srtpTransport = std::dynamic_pointer_cast<DtlsSrtpTransport>(transport);
-		iterateTracks([&](shared_ptr<Track> track) {
-			if (!track->isOpen())
-				track->open(srtpTransport);
+
+		iterateTracks([&](const shared_ptr<Track>& track) {
+			if (!track->isOpen()) {
+				if (srtpTransport) {
+					track->open(srtpTransport);
+				} else {
+					auto errorMsg = "addTrack() was called, but srtp transport was not initialized. This is an optimization for use of the library with data channels only. Set config.forceMediaTransport to 'true' to get the transport initialized before dynamically adding tracks.";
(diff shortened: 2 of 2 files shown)


test/track.cpp
```diff
@@ -33,12 +33,14 @@ void test_track() {
 	InitLogger(LogLevel::Debug);
 
 	Configuration config1;
+	config1.forceTrackAllocation = true;
```
JonasVautherin, line 36: @paullouisageneau: with my condition above (in src/peerconnection.cpp), these tests now need `config.forceTrackAllocation`. Which suggests that either my condition above is wrong, or those tests were not really adding the track, right?

Now they fail, but I don't understand why. Would be nice to have your opinion :innocent:.

src/peerconnection.cpp
```diff
@@ -283,6 +283,11 @@ void PeerConnection::onDataChannel(
 }
 
 std::shared_ptr<Track> PeerConnection::addTrack(Description::Media description) {
+	auto local = impl()->localDescription();
+	if (!impl()->config.forceTrackAllocation && !(local && local->hasAudioOrVideo())) {
+		throw std::runtime_error("addTrack() called, but srtp transport was not initialized. This is an optimization for use of the library with data channels only. Set config.forceTrackAllocation to 'true' to get the transport initialized before dynamically adding tracks.");
+	}
```
paullouisageneau, line 289: You can't do such a check here as most of the time `addTrack()` is called before generating the local offer, so it will throw when `forceTrackAllocation` is false. That's why existing tests for tracks fail.

You could instead prevent opening the track, logging an error, and maybe calling `track->triggerError()` here: https://github.com/paullouisageneau/libdatachannel/blob/99be4b7451fc68a9d52dc307b766c503fb153f8f/src/impl/peerconnection.cpp#L759

test/track.cpp
paullouisageneau: I think this is caused by the check in `addTrack()`, see https://github.com/paullouisageneau/libdatachannel/pull/733#discussion_r1011520309

src/peerconnection.cpp
JonasVautherin: Something like this?

paullouisageneau: Yes it looks good!

paullouisageneau (commented):
It looks good, however just a nitpick: the setting name might not be very clear. It's not about forcing track allocation but about forcing media transport allocation, as tracks refer to the per-stream interface to media transport. What about `forceMediaTransport`?

JonasVautherin:
> What about forceMediaTransport?

Yep that's better! Changed it :+1:

paullouisageneau:
Perfect, thank you!

PR #752, opened 2022-11-23
Support for nal start codes preceded by a zero

web2098:
This PR address an issue in the StartSequenceMatchSucc which didn't correctly identify long/short nal start sequences that were preceded by a zero.

With a byte sequence of 0x00, 0x00, 0x00, 0x0, 0x01:
The unfixed code returns : NUSM_firstZero, NUSM_secondZero, NUSM_thirdZero, NUSM_noMatch

This change will result in the sequence:
NUSM_firstZero, NUSM_secondZero, NUSM_thirdZero, NUSM_thirdZero, NUSM_longMatch

Thus allowing the 3/4 byte sequence detection to continue to shift with any number of preceding zeroes.


paullouisageneau (approved):
Thank you for the fix!

paullouisageneau:
I released the fix in v0.17.12

PR #759, opened 2022-11-26
Enhance seeding for random number generation

paullouisageneau:
Follow-up on https://github.com/paullouisageneau/libdatachannel/pull/757

This PR enhances seeding for random number generation, mostly for the sake of good practice as we are not dealing with cryptographically secure generators here.

### DIFF
--- a/src/impl/utils.cpp
+++ b/src/impl/utils.cpp
@@ -20,11 +20,13 @@
 
 #include "impl/internals.hpp"
 
+#include <algorithm>
 #include <cctype>
 #include <chrono>
 #include <functional>
 #include <iterator>
 #include <sstream>
+#include <thread>
 
 namespace rtc::impl::utils {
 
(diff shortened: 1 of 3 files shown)


src/impl/utils.cpp
```diff
@@ -118,14 +120,20 @@ std::seed_seq random_seed() {
...
 		// Ignore
 	}
 
-	// Seed with current time
-	using std::chrono::system_clock;
-	seed.push_back(static_cast<unsigned int>(system_clock::now().time_since_epoch().count()));
+	// Seed with high-resolution clock
+	using std::chrono::high_resolution_clock;
+	seed.push_back(
+	    static_cast<unsigned int>(high_resolution_clock::now().time_since_epoch().count()));
```
paullouisageneau, line 132: @Nemirtingas Thanks for the suggestion!

PR #765, opened 2022-11-30
Fix export define for Windows

paullouisageneau:
This PR fixes the export define for Windows, as previously it would lead to incorrect behavior, for instance re-exported symbols when used as a static library in a dynamic library.


Nemirtingas:
Should apply the same to https://github.com/paullouisageneau/libdatachannel/blob/master/include/rtc/common.hpp#L30-L40

paullouisageneau:
@Nemirtingas I fixed the C++ exports too.

Nemirtingas:
LGTM.

Issue #728, opened 2022-10-07
Re-licensing under MPL 2.0


paullouisageneau:
After different discussions (in particular in https://github.com/paullouisageneau/libdatachannel/issues/557), there seems to be a common blocking point with the difficulty to bundle the library into closed-source apps due to the LGPL license, even the impossibility to do so for apps in Apple's App store.

**Therefore, I'd like to re-license libdatachannel under [MPL 2.0](https://www.mozilla.org/en-US/MPL/2.0/).**

The MPL 2.0 is also a weak copyleft license like the LGPL and is compatible with the GPL. Basically, modifications must still be redistributed under the same license, but the license applies on file level which allows combining the code with code under any other license. In particular, it would allow static linking the library with closed-source applications without restrictions (see more information in the [MPL 2.0 FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)). It should not change anything for current libdatachannel users.

A preliminary requirement is also re-licensing the dependency libjuice under MPL 2.0: https://github.com/paullouisageneau/libjuice/issues/183

@LeoNavel @murat-dogan @hanseuljun @industriousonesoft @stevedanomodolor @ecotarobles @laramackey @SE2Dev @asgeir @DavidEichmann @willm @nicochatzi @hhgyu @cheungxiongwei @aaronalbers @bigfarts @Faless You have made non-trivial contributions to the library, tests, or example code. Thank you again! Could you please confirm here whether you agree with the re-licensing under MPL 2.0?

murat-dogan:
I will follow your decision.

I guess no need to change `node-datachannel` license, right?

paullouisageneau:
@murat-dogan Thank you. You don't need to change the license, but it could be good to make usage easier. Currently node-datachannel has the same limitations, for instance you can't easily bundle it in a closed-source electron app.

stevedanomodolor:
> After different discussions (in particular in #557), there seems to be a common blocking point with the difficulty to bundle the library into closed-source apps due to the LGPL license, even the impossibility to do so for apps in Apple's App store.
> 
> **Therefore, I'd like to re-license libdatachannel under [MPL 2.0](https://www.mozilla.org/en-US/MPL/2.0/).**
> 
[...]

PR #766, opened 2022-12-01
Add PCMA/PCMU codec to description

Dw9:
This pr supports setting the PCMA/PCMU encoder for webrtc.
test case is adapted from [media_sender](https://github.com/paullouisageneau/libdatachannel/blob/b0a5d99059fe6482ab85400ced92d92bc973bc41/examples/media-sender/main.cpp#L76)

1.    ` rtc::Description::Audio media("audio", rtc::Description::Direction::SendOnly);media.addPCMUCodec(123);`

2.    ffmpeg -stream_loop -1  -re -i test.mp4 -vn  -acodec pcm_mulaw  -ar 8000 -ac 1 -f rtp rtp://127.0.0.1:6000

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -1086,8 +1086,13 @@ Description::Audio::Audio(string mid, Direction dir)
     : Media("audio 9 UDP/TLS/RTP/SAVPF", std::move(mid), dir) {}
 
 void Description::Audio::addAudioCodec(int payloadType, string codec, optional<string> profile) {
-	if (codec.find('/') == string::npos)
-		codec += "/48000/2";
+	if (codec.find('/') == string::npos){
+        if(0 == codec.compare("OPUS"))
+            codec += "/48000/2";
+        else if(0 == codec.compare("PCMA") || 0 == codec.compare("PCMU"))
+            codec += "/8000/1";
+    }
+
 
 	RtpMap map(std::to_string(payloadType) + ' ' + codec);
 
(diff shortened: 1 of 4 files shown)


src/description.cpp
```diff
@@ -1086,8 +1086,13 @@ Description::Audio::Audio(string mid, Direction dir)
     : Media("audio 9 UDP/TLS/RTP/SAVPF", std::move(mid), dir) {}
 
 void Description::Audio::addAudioCodec(int payloadType, string codec, optional<string> profile) {
-	if (codec.find('/') == string::npos)
-		codec += "/48000/2";
+	if (codec.find('/') == string::npos){
+        if(0 == codec.compare("OPUS"))
+            codec += "/48000/2";
+        else if(0 == codec.compare("PCMA") || 0 == codec.compare("PCMU"))
+            codec += "/8000/1";
```
paullouisageneau, line 1093: You could check for "PCMA" or "PCMU" or fallback to default to keep the former behavior. Also, calling `compare` seems convoluted here:
```suggestion
        if(codec == "PCMA") || codec == "PCMU")
            codec += "/8000/1";
        else
            codec += "/48000/2";
```

paullouisageneau:
Thank for the PR. Before merging, could you please confirm you agree with the upcoming re-licensing to MPL 2.0 ? (see https://github.com/paullouisageneau/libdatachannel/issues/728)

hassan379:
it does not work on chrome or Edge but works perfectly on firefox
have u tested it on chrome or edge ?

PR #770, opened 2022-12-08
about the use of RTCP

Dw9:
I use libdatachannel  to play a webrtc stream from a [webrtc server](https://github.com/ZLMediaKit/ZLMediaKit/blob/04422b31b6d96106385e8e64b429b56828a751ac/webrtc/WebRtcTransport.cpp#L779) which got RR will send SR。 
They will continuously receive and transmit RR and SR。 
I think it is better to control the sending of RR and SR at the application layer to prevent the peer from automatically sending.

### DIFF
--- a/include/rtc/rtcpreceivingsession.hpp
+++ b/include/rtc/rtcpreceivingsession.hpp
@@ -40,11 +40,13 @@ public:
 
 	bool requestKeyframe() override;
 
-protected:
-	void pushREMB(unsigned int bitrate);
-	void pushRR(unsigned int lastSR_delay);
+    void pushREMB(unsigned int bitrate);
+    void pushRR(unsigned int lastSR_delay);
 
-	void pushPLI();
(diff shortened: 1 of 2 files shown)


paullouisageneau:
While letting the application layer would indeed solve the issue, it can't be done by simply exposing these methods: the application does not know when the last SR was received so it can't compute `lastSR_delay` to call `pushRR()` (the RTCP receiving session should compute it internally), and `pushREMB` should actually use `mRequestedBitrate`.

Dw9:
I got it, thanks for your reply. :smile: 
When playing a WebRTC stream over UDP, the client should send RTCP  packets to keep the connection alive。 otherwise, the client terminates the connection, the server will continue to transmit data.
Can we set up some callbacks for the user to handle RR?
btw, Is there a plan to implement WebRTC over TCP. as some Webrtc servers like zlm / mediasoup support this feature .

PR #774, opened 2022-12-22
libnice: close agent before releasing it

DarthBo:
Fixes the following warning:

> nice: Agent 0xaaaab3ae7690 : We still have alive TURN refreshes. Consider using
> nice_agent_close_async() to prune them before releasing the agent.


paullouisageneau (approved):
Thank you for the fix!

paullouisageneau:
Thank you for the fix!

PR #784, opened 2023-01-20
Added example signaling server example using libdatachannel ws capi

Tim-S:
Added another signaling-server using libdatachannels websocket implementation + its c-api


Tim-S:
Hi @paullouisageneau ,
I'm not quite sure why the "Build with OpenSSL / build-windows" check failed.
(tests.exe failed, however neither the libdatachannel library-code nor the tests were touched with the changes proposed in this PR).

Can we somehow re-run the tests?

paullouisageneau:
Sorry, I forgot to look into this PR. Thank you for the contribution, but I'm a bit puzzled by the use case, as it is C++ code but is mixed with C-style code and relies on libdatachannel's C API. Could you elaborate please?

Tim-S:
Intended to be just an example on the Websocket C-API provided by libdatachannel.
Moreover thought it would be useful to a signalling-server example given only libdatachannel itself.
Wanted to keep within the libs provided or used by libdatachannel, so no additional dependecies would be required; that is I used to given json.hpp and switched to cpp-style for that part.

Feel free to decline if you think this could should not be merged or at least not in its current form (Mixed C / C++ style).

Just wanted to share, in case someone is interested ;)

PR #785, opened 2023-01-20
Provide ability to bind WebSocketServer to specifc address

### LINKED ISSUE #686: Add some way to specify the interface in rtc::WebSocketServer::Configuration
It is currently not possible to limit or specify the interface that will be used by the WebSocketServer. It would be really nice to have a IP field inside the configuration for that.

Tim-S:
Added 'bindAddress' option to structs rtc::WebSocketServer::Configuration  and rtcWsServerConfiguration (c-api)

Think this resolves #686


src/impl/websocketserver.cpp
```diff
@@ -40,9 +40,13 @@ WebSocketServer::WebSocketServer(Configuration config_)
 			    "Either none or both certificate and key PEM files must be specified");
 		}
 	}
-
+	
+	const char* bindAddress = nullptr;
+	if(config.bindAddress){
+		bindAddress = config.bindAddress->data();
```
paullouisageneau, line 46: While this is valid since C++11 as `data()` now guarantees that a null character terminates the character sequence, I'd prefer a call to `c_str()` when getting a C-style string for clarity.

paullouisageneau:
Thank you for the contribution!

Tim-S:
> Thank you for the contribution!

I welcome the opportunity to make a brief contribution to this great project. Keep going!

PR #795, opened 2023-02-01
feat: add GCM256 support

rafiw:


src/impl/dtlssrtptransport.cpp
```diff
@@ -218,33 +218,51 @@ bool DtlsSrtpTransport::demuxMessage(message_ptr message) {
...
+	if (strcmp(suite, "SRTP_AES128_CM_SHA1_80")==0)
+		return {SRTP_AES_128_KEY_LEN, SRTP_SALT_LEN, srtp_profile_aes128_cm_sha1_80};
+	if (strcmp(suite,"SRTP_AES128_CM_SHA1_32")==0)
+		return {SRTP_AES_128_KEY_LEN, SRTP_SALT_LEN, srtp_profile_aes128_cm_sha1_32};
+	if (strcmp(suite,"SRTP_AEAD_AES_128_GCM")==0)
+		return {SRTP_AES_128_KEY_LEN, SRTP_AEAD_SALT_LEN, srtp_profile_aead_aes_128_gcm};
+	if (strcmp(suite,"SRTP_AEAD_AES_256_GCM")==0)
+		return {SRTP_AES_256_KEY_LEN, SRTP_AEAD_SALT_LEN, srtp_profile_aead_aes_256_gcm};
+	throw std::logic_error("Unexpected SRTP suite name: " + std::string(suite));
+}
```
paullouisageneau, line 231: You should pass a `string` and compare it simply with the `==` operator, IMO `strcmp` should be reserved to C code.
```suggestion
EncryptionParams DtlsSrtpTransport::getEncryptionParams(string suite) {
	if (suite == "SRTP_AES128_CM_SHA1_80")
		return {SRTP_AES_128_KEY_LEN, SRTP_SALT_LEN, srtp_profile_aes128_cm_sha1_80};
	if (suite == "SRTP_AES128_CM_SHA1_32")
		return {SRTP_AES_128_KEY_LEN, SRTP_SALT_LEN, srtp_profile_aes128_cm_sha1_32};
	if (suite == "SRTP_AEAD_AES_128_GCM")
		return {SRTP_AES_128_KEY_LEN, SRTP_AEAD_SALT_LEN, srtp_profile_aead_aes_128_gcm};
	if (suite == "SRTP_AEAD_AES_256_GCM")
		return {SRTP_AES_256_KEY_LEN, SRTP_AEAD_SALT_LEN, srtp_profile_aead_aes_256_gcm};
	throw std::logic_error("Unexpected SRTP suite name: " + suite);
}
```

src/impl/dtlssrtptransport.hpp
```diff
@@ -40,14 +46,14 @@ class DtlsSrtpTransport final : public DtlsTransport {
 	void recvMedia(message_ptr message);
 	bool demuxMessage(message_ptr message) override;
 	void postHandshake() override;
-
+	EncryptionParams getEnctyptionParams(const char* suite);
```
paullouisageneau, line 49:
```suggestion
	EncryptionParams getEncryptionParams(string suite);
```

src/impl/dtlstransport.cpp
```diff
@@ -434,13 +434,20 @@ DtlsTransport::DtlsTransport(shared_ptr<IceTransport> lower, certificate_ptr cer
 		SSL_set_options(mSsl, SSL_OP_SINGLE_ECDH_USE);
 		SSL_set_tmp_ecdh(mSsl, ecdh.get());
 
-		// RFC 8827: The DTLS-SRTP protection profile SRTP_AES128_CM_HMAC_SHA1_80 MUST be supported
-		// See https://www.rfc-editor.org/rfc/rfc8827.html#section-6.5 Warning:
+#if USE_GNUTLS
```
paullouisageneau, line 437: This code is for OpenSSL, no need to handle GnuTLS.

src/impl/dtlssrtptransport.cpp
```diff
@@ -218,33 +218,51 @@ bool DtlsSrtpTransport::demuxMessage(message_ptr message) {
...
 		return;
 
-	static_assert(SRTP_AES_ICM_128_KEY_LEN_WSALT == SRTP_AES_128_KEY_LEN + SRTP_SALT_LEN);
-
-	const size_t materialLen = SRTP_AES_ICM_128_KEY_LEN_WSALT * 2;
-	unsigned char material[materialLen];
+	auto profile = SSL_get_selected_srtp_profile(mSsl);
+	if (!profile)
+		throw std::runtime_error("Failed to get SRTP profile");
+	PLOG_DEBUG << "srtp profile used is: " << profile->name;
```
paullouisageneau, line 240: This is for OpenSSL only and therefore must be in a `#if USE_GNUTLS [...] #else [...] #endif` block. For GnuTLS the correct approach is to call `gnutls_srtp_get_selected_profile()` but you may hardcode the `SRTP_AES128_CM_SHA1_80` profile instead.

paullouisageneau (commented):
Thank you for adding the feature. You need to do some changes to fix compilation with GnuTLS.

src/impl/dtlstransport.cpp
```diff
@@ -434,13 +434,20 @@ DtlsTransport::DtlsTransport(shared_ptr<IceTransport> lower, certificate_ptr cer
...
+		const char *srtpProfiles = "SRTP_AEAD_AES_256_GCM:SRTP_AEAD_AES_128_GCM:SRTP_AES128_CM_SHA1_80";
+#endif
 		// SSL_set_tlsext_use_srtp() returns 0 on success and 1 on error
-		if (SSL_set_tlsext_use_srtp(mSsl, "SRTP_AES128_CM_SHA1_80"))
-			throw std::runtime_error("Failed to set SRTP profile: " +
-			                         openssl::error_string(ERR_get_error()));
-
+		if (SSL_set_tlsext_use_srtp(mSsl, srtpProfiles)) {
+			// RFC 8827: The DTLS-SRTP protection profile SRTP_AES128_CM_HMAC_SHA1_80 MUST be supported
+			// See https://www.rfc-editor.org/rfc/rfc8827.html#section-6.5 Warning:
```
paullouisageneau, line 446: This means the profile must be supported by the remote peer, not by OpenSSL, so it would be better to keep it on top of the whole block to explain why SRTP_AES128_CM_SHA1_80 is the baseline.

paullouisageneau (approved):
Thank you!

PR #796, opened 2023-02-01
Catch exceptions from user callbacks early

paullouisageneau:
This PR introduces try catch blocks to ensure exceptions leaking from user callbacks are caught early, otherwise in some situations it could introduce undesired behavior changes like a second callback not called as it should.

### DIFF
--- a/src/impl/channel.cpp
+++ b/src/impl/channel.cpp
@@ -7,22 +7,44 @@
  */
 
 #include "channel.hpp"
+#include "internals.hpp"
 
 namespace rtc::impl {
 
 void Channel::triggerOpen() {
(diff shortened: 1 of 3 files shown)


themightyoarfish:
Would it be possible to install some kind of exception handler? I've been repeatedly bitten by my errors getting swallowed and it taking a while for me to remember that all exceptions in user callbacks get caught. Now I have to remember to enable logging and then crawl it during development.

paullouisageneau:
@themightyoarfish I get the idea but do you really need a dedicated callback for that? Uncaught exceptions result in warnings that you can easily intercept in the log callback.

PR #811, opened 2023-02-23
Send the DTLS close_notify alert

Dw9:
hi~ @paullouisageneau 

when I use pc->close() to stop sending media data, The peer connection is not closed until timeout.

should we send the dtls close_notify alert for faster hangup without signaling server ?

image
ref: https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html

browse will send a close_notify alert.
image


paullouisageneau:
Good point, the close notify alert should indeed be sent, actually it is with GnuTLS. However, you shouldn't rely on it for reliable termination as the alert might be lost and won't be retransmitted.

paullouisageneau (approved):

PR #812, opened 2023-02-23
Addressing HTTP Header with empty values

web2098:
Hi @paullouisageneau,

I recently encounted an issue were the websocket class was unable to establish a connection to a server because the HTTP response header contained key:value pairs that looked like: ``` X-Custom-Header: ```. 

I looked into and it appears the code in wshandshake assumes that there will be a non ```' '``` character after the ```:``` and tries to create a substring. I created a quick patch to check the subpos before using it to generate a substring.


paullouisageneau (approved):
Thank you for the fix!

PR #818, opened 2023-03-07
Allow using system lib for all submodules

tytan652:
libdatachannel requires usrsctp, Plog and Nlohmann JSON (examples only) submodules.
This PR enable in CMake using system lib for those and add an option `PREFER_SYSTEM_LIB` to enable all `USE_SYSTEM_`.

This change allows libdatachannel to be built and packaged on Debian and Fedora build-systems by packaging and providing all the dependencies independently since those build-systems do not support git submodules.


tytan652:
How does my CMake changes (that do no change the default behavior) make the [macOS OpenSSL](https://github.com/paullouisageneau/libdatachannel/actions/runs/4353738974/jobs/7625327944) CI fail ?

paullouisageneau (approved):
It looks good, thank you!

The test failure was an unrelated spurious one, the TURN test relies on an external server which is not always reliable.

Issue #816, opened 2023-03-02
WebSocket Crash


web2098:
Hi @paullouisageneau,

When testing using websocket with a lost connection, I encountered a crash via a mutex throwing an exception during dtor.

Here is the stack I found:
![image](https://user-images.githubusercontent.com/7141138/222561482-4dd42e27-3ed4-4e2d-9081-045317d5d3de.png)

From what I can tell it starts with this code
_**TcpTransport::process ~440**_
``` 
PLOG_INFO << "TCP disconnected";
PollService::Instance().remove(mSock);
changeState(State::Disconnected);
recv(nullptr);
```

Which changes the state, calling the synchronized_callback object which locks its internal mutex.

The code then calls the lambda for the in the websocket:
_**websocket::setTcpTransport ~231**_
```
auto shared_this = weak_this.lock();
...

case State::Disconnected:
  remoteClose();
  break;
```

I am not really sure what happens here. Either the websocket is deleted at the end of the lambda, or during the remote close the code transport object is deleted. Either way once the transport object is deleted, that is the same instance that is at the start of the stack. During destruction of the transport its synchronized_callback is destroyed which still has the mutex locked.

![image](https://user-images.githubusercontent.com/7141138/222561420-fd215cda-b356-440d-8ed9-6b8831bb52de.png)

paullouisageneau:
Thank you for reporting! The issue seems to be that the `TcpTransport` is not held while `process` is called. A simple fix could be to make `TcpTransport` derive from `std::enable_shared_from_this<TcpTransport>` and locking a shared ptr to self in the process callback.

PR #821, opened 2023-03-08
Fix crash in TcpTransport

web2098:
Addressing issue where the tcptransport object could go out of scope during a process call (#816). As per @paullouisageneau suggestion I modified the process function to grab a shared pointer to prevent the class from going out of scope. I decided to grab the weak_ptr and lock it to avoid the potential throw from shared_from_this.

I test this change with my current setup of letting the WebSocket connection timeout after having been connected. It now correctly closes the WebSocket and doesn't crash.

### DIFF
--- a/src/impl/tcptransport.cpp
+++ b/src/impl/tcptransport.cpp
@@ -384,7 +384,7 @@ void TcpTransport::triggerBufferedAmount(size_t amount) {
 }
 
 void TcpTransport::process(PollService::Event event) {
-	auto self = shared_from_this();
+	auto self = weak_from_this().lock();
 	try {
 		switch (event) {
 		case PollService::Event::Error: {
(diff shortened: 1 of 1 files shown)


src/impl/tcptransport.cpp
```diff
@@ -384,6 +384,7 @@ void TcpTransport::triggerBufferedAmount(size_t amount) {
 }
 
 void TcpTransport::process(PollService::Event event) {
+	auto self = weak_from_this().lock();
```
paullouisageneau, line 387: You should return if `lock()` returns `nullptr`, just to be sure. 
```suggestion
	auto self = weak_from_this().lock();
	if (!self)
		return;
```

paullouisageneau (approved):
The TURN test fails because the external server is down, I'll disable the test.

PR #822, opened 2023-03-10
Add Mbed TLS Support

Sean-Der:
This is still a WIP. If anyone wants to follow my progress/has feedback just want to make it public

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(libdatachannel
 set(PROJECT_DESCRIPTION "C/C++ WebRTC network library featuring Data Channels, Media Transport, and WebSockets")
 
 # Options
+option(USE_MBEDTLS "Use Mbed TLS instead of OpenSSL" OFF)
 option(USE_GNUTLS "Use GnuTLS instead of OpenSSL" OFF)
 option(USE_NICE "Use libnice instead of libjuice" OFF)
 option(PREFER_SYSTEM_LIB "Prefer system libraries over deps folder" OFF)
@@ -21,12 +22,22 @@ option(WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
 option(CAPI_STDCALL "Set calling convention of C API callbacks stdcall" OFF)
 option(SCTP_DEBUG "Enable SCTP debugging output to verbose log" OFF)
 
+if (USE_MBEDTLS AND USE_GNUTLS)
+	message(FATAL_ERROR "Both USE_MBEDTLS and USE_GNUTLS can not be enabled at the same time")
+endif()
+
+
 if(USE_GNUTLS)
 	option(USE_NETTLE "Use Nettle in libjuice" ON)
 else()
 	option(USE_NETTLE "Use Nettle in libjuice" OFF)
+
 	if(NOT USE_SYSTEM_SRTP)
-		option(ENABLE_OPENSSL "Enable OpenSSL crypto engine for SRTP" ON)
+		if (USE_MBEDTLS)
+			option(ENABLE_MBEDTLS "Enable Mbed TLS crypto engine for SRTP" ON)
+		else()
+			option(ENABLE_OPENSSL "Enable OpenSSL crypto engine for SRTP" ON)
+		endif()
 	endif()
 endif()
 
@@ -337,6 +348,13 @@ if (USE_GNUTLS)
 		target_link_libraries(datachannel PRIVATE Nettle::Nettle)
 		target_link_libraries(datachannel-static PRIVATE Nettle::Nettle)
 	endif()
+elseif(USE_MBEDTLS)
+	find_package(MbedTLS 3 REQUIRED)
+
+	target_compile_definitions(datachannel PRIVATE USE_MBEDTLS)
+	target_compile_definitions(datachannel-static PRIVATE USE_MBEDTLS)
+	target_link_libraries(datachannel PRIVATE MbedTLS::MbedTLS)
+	target_link_libraries(datachannel-static PRIVATE MbedTLS::MbedTLS)
 else()
 	if(APPLE)
 		# This is a bug in CMake that causes it to prefer the system version over
--- a/cmake/Modules/FindMbedTLS.cmake
+++ b/cmake/Modules/FindMbedTLS.cmake
@@ -1,0 +1,214 @@
+#[=======================================================================[.rst
+FindMbedTLS
+-----------
+
+FindModule for MbedTLS and associated libraries
+
+Components
+^^^^^^^^^^
+
+This module contains provides several components:
+
+``MbedCrypto``
+``MbedTLS``
+``MbedX509``
+
+Import targets exist for each component.
+
+Imported Targets
+^^^^^^^^^^^^^^^^
+
+This module defines the :prop_tgt:`IMPORTED` targets:
+
+``MbedTLS::MbedCrypto``
+  Crypto component
+
+``MbedTLS::MbedTLS``
+  TLS component
+
+``MbedTLS::MbedX509``
+  X509 component
+
+Result Variables
+^^^^^^^^^^^^^^^^
+
+This module sets the following variables:
+
+``MbedTLS_FOUND``
+  True, if all required components and the core library were found.
+``MbedTLS_VERSION``
+  Detected version of found MbedTLS libraries.
+
+``MbedTLS_<COMPONENT>_VERSION``
+  Detected version of found MbedTLS component library.
+
+Cache variables
+^^^^^^^^^^^^^^^
+
+The following cache variables may also be set:
+
+``MbedTLS_<COMPONENT>_LIBRARY``
+  Path to the library component of MbedTLS.
+``MbedTLS_<COMPONENT>_INCLUDE_DIR``
+  Directory containing ``<COMPONENT>.h``.
+
+Distributed under the MIT License, see accompanying LICENSE file or
+https://github.com/PatTheMav/cmake-finders/blob/master/LICENSE for details.
+(c) 2023 Patrick Heyer
+
+#]=======================================================================]
+
+# cmake-format: off
+# cmake-lint: disable=C0103
+# cmake-lint: disable=C0301
+# cmake-lint: disable=C0307
+# cmake-format: on
+
+include(FindPackageHandleStandardArgs)
+
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(PC_MbedTLS QUIET mbedtls mbedcrypto mbedx509)
+endif()
+
+# MbedTLS_set_soname: Set SONAME on imported library targets
+macro(MbedTLS_set_soname component)
+  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Darwin")
+    execute_process(
+      COMMAND sh -c "otool -D '${Mbed${component}_LIBRARY}' | grep -v '${Mbed${component}_LIBRARY}'"
+      OUTPUT_VARIABLE _output
+      RESULT_VARIABLE _result)
+
+    if(_result EQUAL 0 AND _output MATCHES "^@rpath/")
+      set_property(TARGET MbedTLS::Mbed${component} PROPERTY IMPORTED_SONAME "${_output}")
+    endif()
+  elseif(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux|FreeBSD")
+    execute_process(
+      COMMAND sh -c "objdump -p '${Mbed${component}_LIBRARY}' | grep SONAME"
+      OUTPUT_VARIABLE _output
+      RESULT_VARIABLE _result)
+
+    if(_result EQUAL 0)
+      string(REGEX REPLACE "[ \t]+SONAME[ \t]+([^ \t]+)" "\\1" _soname "${_output}")
+      set_property(TARGET MbedTLS::Mbed${component} PROPERTY IMPORTED_SONAME "${_soname}")
+      unset(_soname)
+    endif()
+  endif()
+  unset(_output)
+  unset(_result)
+endmacro()
+
+find_path(
+  MbedTLS_INCLUDE_DIR
+  NAMES mbedtls/ssl.h
+  HINTS "${PC_MbedTLS_INCLUDE_DIRS}"
+  PATHS /usr/include /usr/local/include
+  DOC "MbedTLS include directory")
+
+if(PC_MbedTLS_VERSION VERSION_GREATER 0)
+  set(MbedTLS_VERSION ${PC_MbedTLS_VERSION})
+elseif(EXISTS "${MbedTLS_INCLUDE_DIR}/mbedtls/build_info.h")
+  file(STRINGS "${MbedTLS_INCLUDE_DIR}/mbedtls/build_info.h" _VERSION_STRING
+       REGEX "#define[ \t]+MBEDTLS_VERSION_STRING[ \t]+.+")
+  string(REGEX REPLACE ".*#define[ \t]+MBEDTLS_VERSION_STRING[ \t]+\"(.+)\".*" "\\1" MbedTLS_VERSION
+                       "${_VERSION_STRING}")
+else()
+  if(NOT MbedTLS_FIND_QUIETLY)
+    message(AUTHOR_WARNING "Failed to find MbedTLS version.")
+  endif()
+  set(MbedTLS_VERSION 0.0.0)
+endif()
+
+find_library(
+  MbedTLS_LIBRARY
+  NAMES libmbedtls mbedtls
+  HINTS "${PC_MbedTLS_LIBRARY_DIRS}"
+  PATHS /usr/lib /usr/local/lib
+  DOC "MbedTLS location")
+
+find_library(
+  MbedCrypto_LIBRARY
+  NAMES libmbedcrypto mbedcrypto
+  HINTS "${PC_MbedTLS_LIBRARY_DIRS}"
+  PATHS /usr/lib /usr/local/lib
+  DOC "MbedCrypto location")
+
+find_library(
+  MbedX509_LIBRARY
+  NAMES libmbedx509 mbedx509
+  HINTS "${PC_MbedTLS_LIBRARY_DIRS}"
+  PATHS /usr/lib /usr/local/lib
+  DOC "MbedX509 location")
+
+if(MbedTLS_LIBRARY
+   AND NOT MbedCrypto_LIBRARY
+   AND NOT MbedX509_LIBRARY)
+  set(CMAKE_REQUIRED_LIBRARIES "${MbedTLS_LIBRARY}")
+  set(CMAKE_REQUIRED_INCLUDES "${MbedTLS_INCLUDE_DIR}")
+
+  check_symbol_exists(mbedtls_x509_crt_init "mbedtls/x590_crt.h" MbedTLS_INCLUDES_X509)
+  check_symbol_exists(mbedtls_sha256_init "mbedtls/sha256.h" MbedTLS_INCLUDES_CRYPTO)
+  unset(CMAKE_REQUIRED_LIBRARIES)
+  unset(CMAKE_REQUIRED_INCLUDES)
+endif()
+
+if(CMAKE_HOST_SYSTEM_NAME MATCHES "Darwin|Windows")
+  set(MbedTLS_ERROR_REASON "Ensure that an MbedTLS distribution is provided as part of CMAKE_PREFIX_PATH.")
+elseif(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux|FreeBSD")
+  set(MbedTLS_ERROR_REASON "Ensure that MbedTLS is installed on the system.")
+endif()
+
+if(MbedTLS_INCLUDES_X509 AND MbedTLS_INCLUDES_CRYPTO)
+  find_package_handle_standard_args(
+    MbedTLS
+    REQUIRED_VARS MbedTLS_LIBRARY MbedTLS_INCLUDE_DIR
+    VERSION_VAR MbedTLS_VERSION REASON_FAILURE_MESSAGE "${MbedTLS_ERROR_REASON}")
+  mark_as_advanced(MbedTLS_LIBRARY MbedTLS_INCLUDE_DIR)
+  list(APPEND _COMPONENTS TLS)
+else()
+  find_package_handle_standard_args(
+    MbedTLS
+    REQUIRED_VARS MbedTLS_LIBRARY MbedCrypto_LIBRARY MbedX509_LIBRARY MbedTLS_INCLUDE_DIR
+    VERSION_VAR MbedTLS_VERSION REASON_FAILURE_MESSAGE "${MbedTLS_ERROR_REASON}")
+  mark_as_advanced(MbedTLS_LIBRARY MbedCrypto_LIBRARY MbedX509_LIBRARY MbedTLS_INCLUDE_DIR)
+  list(APPEND _COMPONENTS TLS Crypto X509)
+endif()
+unset(MbedTLS_ERROR_REASON)
+
+if(MbedTLS_FOUND)
+  foreach(component IN LISTS _COMPONENTS)
+    if(NOT TARGET MbedTLS::Mbed${component})
+      if(IS_ABSOLUTE "${Mbed${component}_LIBRARY}")
+        add_library(MbedTLS::Mbed${component} UNKNOWN IMPORTED)
+        set_property(TARGET MbedTLS::Mbed${component} PROPERTY IMPORTED_LOCATION "${Mbed${component}_LIBRARY}")
+      else()
+        add_library(MbedTLS::Mbed${component} INTERFACE IMPORTED)
+        set_property(TARGET MbedTLS::Mbed${component} PROPERTY IMPORTED_LIBNAME "${Mbed${component}_LIBRARY}")
+      endif()
+
+      mbedtls_set_soname(${component})
+      set_target_properties(
+        MbedTLS::MbedTLS
+        PROPERTIES INTERFACE_COMPILE_OPTIONS "${PC_MbedTLS_CFLAGS_OTHER}"
+                   INTERFACE_INCLUDE_DIRECTORIES "${MbedTLS_INCLUDE_DIR}"
+                   VERSION ${MbedTLS_VERSION})
+    endif()
+  endforeach()
+
+  if(MbedTLS_INCLUDES_X509 AND MbedTLS_INCLUDES_CRYPTO)
+    set(MbedTLS_LIBRARIES ${MbedTLS_LIBRARY})
+    set(MBEDTLS_INCLUDE_DIRS ${MbedTLS_INCLUDE_DIR})
+  else()
+    set(MbedTLS_LIBRARIES ${MbedTLS_LIBRARY} ${MbedCrypto_LIBRARY} ${MbedX509_LIBRARY})
+    set_property(TARGET MbedTLS::MbedTLS PROPERTY INTERFACE_LINK_LIBRARIES MbedTLS::MbedCrypto MbedTLS::MbedX509)
+    set(MBEDTLS_INCLUDE_DIRS ${MbedTLS_INCLUDE_DIR})
+  endif()
+endif()
+
+include(FeatureSummary)
+set_package_properties(
+  MbedTLS PROPERTIES
+  URL "https://www.trustedfirmware.org/projects/mbed-tls"
+  DESCRIPTION
+    "A C library implementing cryptographic primitives, X.509 certificate manipulation, and the SSL/TLS and DTLS protocols."
+)
--- a/src/impl/certificate.cpp
+++ b/src/impl/certificate.cpp
@@ -111,8 +111,6 @@ Certificate::Certificate(shared_ptr<gnutls_certificate_credentials_t> creds)
 
 gnutls_certificate_credentials_t Certificate::credentials() const { return *mCredentials; }
 
-string Certificate::fingerprint() const { return mFingerprint; }
-
 string make_fingerprint(gnutls_certificate_credentials_t credentials) {
 	auto new_crt_list = [credentials]() -> gnutls_x509_crt_t * {
 		gnutls_x509_crt_t *crt_list = nullptr;
@@ -149,7 +147,172 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
 	return oss.str();
 }
 
-#else // USE_GNUTLS==0
+#elif USE_MBEDTLS
+string make_fingerprint(shared_ptr<mbedtls_x509_crt> crt) {
+	const int size = 32;
+	uint8_t buffer[size];
+	std::stringstream fingerprint;
+
+	mbedtls::check(
+	    mbedtls_sha256(crt->raw.p, crt->raw.len, reinterpret_cast<unsigned char *>(buffer), 0),
+	    "Failed to generate certificate fingerprint");
+
+	for (auto i = 0; i < size; i++) {
+		fingerprint << std::setfill('0') << std::setw(2) << std::hex << static_cast<int>(buffer[i]);
+		if (i != (size - 1)) {
+			fingerprint << ":";
+		}
+	}
+
+	return fingerprint.str();
+}
+
+Certificate::Certificate(shared_ptr<mbedtls_x509_crt> crt, shared_ptr<mbedtls_pk_context> pk)
+    : mCrt(crt), mPk(pk), mFingerprint(make_fingerprint(crt)) {}
+
+Certificate Certificate::FromString(string crt_pem, string key_pem) {
+	PLOG_DEBUG << "Importing certificate from PEM string (MbedTLS)";
+
+	auto crt = mbedtls::new_x509_crt();
+	auto pk = mbedtls::new_pk_context();
+
+	mbedtls::check(mbedtls_x509_crt_parse(crt.get(),
+	                                      reinterpret_cast<const unsigned char *>(crt_pem.c_str()),
+	                                      crt_pem.length()),
+	               "Failed to parse certificate");
+	mbedtls::check(mbedtls_pk_parse_key(pk.get(),
+	                                    reinterpret_cast<const unsigned char *>(key_pem.c_str()),
+	                                    key_pem.size(), NULL, 0, NULL, 0),
+	               "Failed to parse key");
+
+	return Certificate(std::move(crt), std::move(pk));
+}
+
+Certificate Certificate::FromFile(const string &crt_pem_file, const string &key_pem_file,
+                                  const string &pass) {
+	PLOG_DEBUG << "Importing certificate from PEM file (MbedTLS): " << crt_pem_file;
+
+	auto crt = mbedtls::new_x509_crt();
+	auto pk = mbedtls::new_pk_context();
+
+	mbedtls::check(mbedtls_x509_crt_parse_file(crt.get(), crt_pem_file.c_str()),
+	               "Failed to parse certificate");
+	mbedtls::check(mbedtls_pk_parse_keyfile(pk.get(), key_pem_file.c_str(), pass.c_str(), 0, NULL),
+	               "Failed to parse key");
+
+	return Certificate(std::move(crt), std::move(pk));
+}
+
+Certificate Certificate::Generate(CertificateType type, const string &commonName) {
+	PLOG_DEBUG << "Generating certificate (MbedTLS)";
+
+	mbedtls_entropy_context entropy;
+	mbedtls_ctr_drbg_context drbg;
+	mbedtls_x509write_cert wcrt;
+	mbedtls_mpi serial;
+	auto crt = mbedtls::new_x509_crt();
+	auto pk = mbedtls::new_pk_context();
+
+	mbedtls_entropy_init(&entropy);
+	mbedtls_ctr_drbg_init(&drbg);
+	mbedtls_ctr_drbg_set_prediction_resistance(&drbg, MBEDTLS_CTR_DRBG_PR_ON);
+	mbedtls_x509write_crt_init(&wcrt);
+	mbedtls_mpi_init(&serial);
+
+	try {
+		mbedtls::check(mbedtls_ctr_drbg_seed(
+		    &drbg, mbedtls_entropy_func, &entropy,
+		    reinterpret_cast<const unsigned char *>(commonName.data()), commonName.size()));
+
+		switch (type) {
+		// RFC 8827 WebRTC Security Architecture 6.5. Communications Security
+		// All implementations MUST support DTLS 1.2 with the
+		// TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 cipher suite and the P-256 curve
+		// See https://www.rfc-editor.org/rfc/rfc8827.html#section-6.5
+		case CertificateType::Default:
+		case CertificateType::Ecdsa: {
+			mbedtls::check(mbedtls_pk_setup(pk.get(), mbedtls_pk_info_from_type(MBEDTLS_PK_ECKEY)));
+			mbedtls::check(mbedtls_ecp_gen_key(MBEDTLS_ECP_DP_SECP256R1, mbedtls_pk_ec(*pk.get()),
+			                                   mbedtls_ctr_drbg_random, &drbg),
+			               "Unable to generate ECDSA P-256 key pair");
+			break;
+		}
+		case CertificateType::Rsa: {
+			const unsigned int nbits = 2048;
+			const int exponent = 65537;
+
+			mbedtls::check(mbedtls_pk_setup(pk.get(), mbedtls_pk_info_from_type(MBEDTLS_PK_RSA)));
+			mbedtls::check(mbedtls_rsa_gen_key(mbedtls_pk_rsa(*pk.get()), mbedtls_ctr_drbg_random,
+			                                   &drbg, nbits, exponent),
+			               "Unable to generate RSA key pair");
+			break;
+		}
+		default:
+			throw std::invalid_argument("Unknown certificate type");
+		}
+
+		auto now = std::chrono::system_clock::now();
+		string notBefore = mbedtls::format_time(now - std::chrono::hours(1));
+		string notAfter = mbedtls::format_time(now + std::chrono::hours(24 * 365));
+
+		const size_t serialBufferSize = 16;
+		unsigned char serialBuffer[serialBufferSize];
+		mbedtls::check(mbedtls_ctr_drbg_random(&drbg, serialBuffer, serialBufferSize),
+		               "Failed to generate certificate");
+		mbedtls::check(mbedtls_mpi_read_binary(&serial, serialBuffer, serialBufferSize),
+		               "Failed to generate certificate");
+
+		std::string name = std::string("O=" + commonName + ",CN=" + commonName);
+		mbedtls::check(mbedtls_x509write_crt_set_serial(&wcrt, &serial),
+		               "Failed to generate certificate");
+		mbedtls::check(mbedtls_x509write_crt_set_subject_name(&wcrt, name.c_str()),
+		               "Failed to generate certificate");
+		mbedtls::check(mbedtls_x509write_crt_set_issuer_name(&wcrt, name.c_str()),
+		               "Failed to generate certificate");
+		mbedtls::check(
+		    mbedtls_x509write_crt_set_validity(&wcrt, notBefore.c_str(), notAfter.c_str()),
+		    "Failed to generate certificate");
+
+		mbedtls_x509write_crt_set_version(&wcrt, MBEDTLS_X509_CRT_VERSION_3);
+		mbedtls_x509write_crt_set_subject_key(&wcrt, pk.get());
+		mbedtls_x509write_crt_set_issuer_key(&wcrt, pk.get());
+		mbedtls_x509write_crt_set_md_alg(&wcrt, MBEDTLS_MD_SHA256);
+
+		const size_t certificateBufferSize = 4096;
+		unsigned char certificateBuffer[certificateBufferSize];
+		std::memset(certificateBuffer, 0, certificateBufferSize);
+
+		auto certificateLen = mbedtls_x509write_crt_der(
+		    &wcrt, certificateBuffer, certificateBufferSize, mbedtls_ctr_drbg_random, &drbg);
+		if (certificateLen <= 0) {
+			throw std::runtime_error("Certificate generation failed");
+		}
+
+		mbedtls::check(mbedtls_x509_crt_parse_der(
+		                   crt.get(), (certificateBuffer + certificateBufferSize - certificateLen),
+		                   certificateLen),
+		               "Failed to generate certificate");
+	} catch (...) {
+		mbedtls_entropy_free(&entropy);
+		mbedtls_ctr_drbg_free(&drbg);
+		mbedtls_x509write_crt_free(&wcrt);
+		mbedtls_mpi_free(&serial);
+		throw;
+	}
+
+	mbedtls_entropy_free(&entropy);
+	mbedtls_ctr_drbg_free(&drbg);
+	mbedtls_x509write_crt_free(&wcrt);
+	mbedtls_mpi_free(&serial);
+	return Certificate(std::move(crt), std::move(pk));
+}
+
+std::tuple<shared_ptr<mbedtls_x509_crt>, shared_ptr<mbedtls_pk_context>>
+Certificate::credentials() const {
+	return {mCrt, mPk};
+}
+
+#else // OPENSSL
 
 namespace {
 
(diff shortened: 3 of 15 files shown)


CMakeLists.txt
```diff
@@ -302,6 +308,15 @@ else()
 	endif()
 endif()
 
+if (USE_GNUTLS OR USE_MBEDTLS)
+	if (NOT NO_WEBSOCKET)
+		# Needed for SHA1, it should be present as GnuTLS/MbedTLS cryptography backend
```
paullouisageneau, line 313: Unless I'm mistaken, MbedTLS does not rely on Nettle, so Nettle should not be linked against in that case. Instead, it looks like the SHA1 API is in `mbedtls/sha1.h`.

CMakeLists.txt
```diff
@@ -17,7 +18,12 @@ option(WARNINGS_AS_ERRORS "Treat warnings as errors" OFF)
 option(CAPI_STDCALL "Set calling convention of C API callbacks stdcall" OFF)
 option(SCTP_DEBUG "Enable SCTP debugging output to verbose log" OFF)
 
-if(USE_GNUTLS)
+if (USE_MBEDTLS AND USE_GNUTLS)
+	message(FATAL_ERROR "Both USE_MBEDTLS and USE_GNUTLS can not be enabled at the same time")
+endif()
+
+
+if(USE_GNUTLS OR USE_MBEDTLS)
```
paullouisageneau, line 26: Same here, Nettle should not be linked against with MbedTLS.

CMakeLists.txt
```diff
@@ -317,11 +323,18 @@ if (USE_GNUTLS)
...
 	target_link_libraries(datachannel-static PRIVATE GnuTLS::GnuTLS)
 	if (NOT NO_WEBSOCKET)
-		# Needed for SHA1, it should be present as GnuTLS cryptography backend
+		# Needed for SHA1, it should be present as GnuTLS/MbedTLS cryptography backend
 		find_package(Nettle REQUIRED)
 		target_link_libraries(datachannel PRIVATE Nettle::Nettle)
 		target_link_libraries(datachannel-static PRIVATE Nettle::Nettle)
 	endif()
+elseif(USE_MBEDTLS)
+	find_package(MbedTLS CONFIG REQUIRED)
```
tytan652, line 332:
```suggestion
	find_package(MbedTLS REQUIRED)
```
I don't think `CONFIG` is needed.

A finder should be added to the project since MbedTLS, since neither CMake or MbedTLS do and it might help you for checking for needed components (e.g. X509). 

Here [one](https://github.com/PatTheMav/obs-studio/blob/macos-build-update/cmake/finders/FindMbedTLS.cmake), you could use. It's the one plan to be used OBS Studio in the future, some OBS repo related comments could be removed.

tytan652 (commented):
I think this PR needs to be tested against MbedTLS 2 and 3.

CMakeLists.txt
Sean-Der: :+1: Fixed, removed the nettle usage for mbed tls

CMakeLists.txt
Sean-Der: :+1: Fixed, removed the nettle usage for mbed tls

CMakeLists.txt
```diff
@@ -317,11 +323,18 @@ if (USE_GNUTLS)
...
+
+	target_compile_definitions(datachannel PRIVATE USE_MBEDTLS=1)
+	target_compile_definitions(datachannel-static PRIVATE USE_MBEDTLS=1)
```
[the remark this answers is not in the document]
Sean-Der: Fixed!

Sean-Der:
@tytan652 What version of mbedtls does OBS use today?

I am hesitant to support mbed tls because it has things like `DTLS 1.0`. I don't want to encourage/support insecure endpoints https://bugs.chromium.org/p/webrtc/issues/detail?id=10261

tytan652:
But you can support only 3 if you want and use GNUTLS for Linux, we already do that for FFmpeg (one different TLS lib per OS).

Sean-Der:
Thanks @tytan652. Your choice @paullouisageneau it is your code base! 

I can add support for both (if you prefer), but personally I prefer less ifdef and lines of code and push people to update dependencies.

CMakeLists.txt
tytan652: Side-note if there any issues with the finder don't hesitate to report them there: https://github.com/obsproject/obs-studio/pull/8247

paullouisageneau:
@Sean-Der As I understand, there is no runtime setting in MbedTLS to force a minimal TLS version? That sounds like a weird design.

In that case, I think you may support only version 3, as said earlier people can link against GnuTLS (or OpenSSL) system libraries on Linux.

Sean-Der:
This compiles/all the WebRTC tests now pass! 

I need to implement the Websocket side now. Did this with `-DNO_WEBSOCKET`

paullouisageneau:
> This compiles/all the WebRTC tests now pass!
> 
> I need to implement the Websocket side now. Did this with `-DNO_WEBSOCKET`

Great news, congrats! Could you please add a Github worflow `build-mbedtls.yml` when you have finished with the TLS transport?

Sean-Der:
@paullouisageneau done!

I am gonna head to bed. In the morning will fix any test failures that have come up. When everything passes I think we are ready for a review!

.gitmodules
```diff
@@ -17,3 +17,6 @@
 	path = deps/libsrtp
 	url = https://github.com/cisco/libsrtp.git
 	shallow = true
+[submodule "deps/mbedtls"]
+	path = deps/mbedtls
+	url = https://github.com/Mbed-TLS/mbedtls.git
```
tytan652, line 22: Is Mbed TLS added as a submodule just because Ubuntu only provides the version 2 ?

For the macOS, you use homebrew like for OpenSSL.

tytan652:
A CMake finder for Mbed TLS is still missing, CMake and Mbed TLS does not provide their own. You need to provide to libdatachannel this finder.

.gitmodules
Sean-Der: Unfortunately I can't use Mbed TLS from homebrew or apt.

`MBEDTLS_SSL_DTLS_SRTP` isn't enabled by default. I opened a PR [here](https://github.com/Homebrew/homebrew-core/pull/125402) to enabled it for homebrew.

CMakeLists.txt
```diff
@@ -335,6 +342,20 @@ if (USE_GNUTLS)
 		target_link_libraries(datachannel PRIVATE Nettle::Nettle)
 		target_link_libraries(datachannel-static PRIVATE Nettle::Nettle)
 	endif()
+elseif(USE_MBEDTLS)
+	if(USE_SYSTEM_MBEDTLS)
+		find_package(MbedTLS REQUIRED)
```
tytan652, line 347: You must set a minimum version in the `find_package`. Replace the `3` by right minimum version if needed.
```suggestion
		find_package(MbedTLS 3 REQUIRED)
```

tytan652 (changes requested):

Sean-Der:
@paullouisageneau All the tests pass!

We have some duplication across GnuTLS/OpenSSL/MbedTLS. I would love to start another PR after this to clean that up a bit (if you want/agree)

paullouisageneau:
> @paullouisageneau All the tests pass!

Looking great, I'll do I thorough review when I get a moment.

> We have some duplication across GnuTLS/OpenSSL/MbedTLS. I would love to start another PR after this to clean that up a bit (if you want/agree)

It would be nice, thank you!

.gitmodules
Sean-Der: @paullouisageneau I agree, but I think this is the 'least' worst?

* Users that don't enable Mbed TLS aren't paying a cost (add_subdirectory doesn't happen)
* Keeps CI+Development simple. Having a pinned version of Mbed TLS and making development easy is nice

In the future if Homebrew and Ubuntu package Mbed TLS with DTLS-SRTP enabled we could drop it then?

Sean-Der:
@paullouisageneau At a high level anything I need to improve?

Also feel free to cherry-pick/rewrite/split commits etc... as a maintainer it can be frustrating to get code that isn't consistent with the code base, so totally understand if you want to rewrite everything :)

cmake/Modules/FindMbedTLS.cmake
```diff
@@ -0,0 +1,208 @@
...
+``MbedTLS_<COMPONENT>_VERSION``
+  Detected version of found MbedTLS component library.
+
+Cache variables
+^^^^^^^^^^^^^^^
+
+The following cache variables may also be set:
+
+``MbedTLS_<COMPONENT>_LIBRARY``
+  Path to the library component of FFmpeg.
```
DDRBoxman, line 51: FFmpeg -> mbedtls

Sean-Der: Fixed! Thanks @DDRBoxman :)

Sean-Der:
@paullouisageneau rebased!

I don't see Mbed TLS tests here, but I opened up a PR against my fork and everything passed. Since I am not a owner of this repo I can't add new jobs?

CMakeLists.txt
```diff
@@ -335,6 +342,20 @@ if (USE_GNUTLS)
...
+		file(READ deps/mbedtls/include/mbedtls/mbedtls_config.h MBEDTLS_CONFIG)
+		string(REPLACE "//#define MBEDTLS_SSL_DTLS_SRTP" "#define MBEDTLS_SSL_DTLS_SRTP" MBEDTLS_CONFIG "${MBEDTLS_CONFIG}")
+		file(WRITE deps/mbedtls/include/mbedtls/mbedtls_config.h "${MBEDTLS_CONFIG}")
+		add_subdirectory(deps/mbedtls EXCLUDE_FROM_ALL)
+	endif()
+
+	target_compile_definitions(datachannel PRIVATE USE_MBEDTLS)
+	target_compile_definitions(datachannel-static PRIVATE USE_MBEDTLS)
+	target_link_libraries(datachannel PRIVATE MbedTLS::mbedtls)
+	target_link_libraries(datachannel-static PRIVATE MbedTLS::mbedtls)
```
tytan652, line 358: The Finder provides `MbedTLS::MbedTLS` and not `MbedTLS::mbedtls`.

tytan652 (changes requested):

CMakeLists.txt
Sean-Der: @tytan652 That breaks the build

```
CMake Error at CMakeLists.txt:357 (target_link_libraries):
  Target "datachannel" links to:

    MbedTLS::MbedTLS

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at CMakeLists.txt:358 (target_link_libraries):
  Target "datachannel-static" links to:

    MbedTLS::MbedTLS

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

CMakeLists.txt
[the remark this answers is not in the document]
Sean-Der: Reverted. This broke the build

CMakeLists.txt
tytan652: Because the submodules and the Finder for system install set different targets. So you need to use `MbedTLS::mbedtls` if `USE_SYSTEM_MBEDTLS=OFF` and `MbedTLS::MbedTLS` if `USE_SYSTEM_MBEDTLS=ON`.

.gitmodules
paullouisageneau: There is still a cost for users because Mbed TLS will be cloned by default, for instance when you run `git clone --recursive` or `git submodule --init`.  Since the Homebrew package now has DTLS-SRTP enabled (thank you for this), I think for now you could remove Ubuntu which doesn't feature MbedTLS 3 anyway and drop the submodule.

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,308 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+	do {
+		std::lock_guard lock(mSendMutex);
+		mCurrentDscp = message->dscp;
+
+		if (message->size() > size_t(mbedtls_ssl_get_max_out_record_payload(&mSsl)))
+			return false;
+
+		auto buff = &(message->data())[0];
+
+		ret = mbedtls_ssl_write(&mSsl, (const unsigned char *) buff, message->size());
```
paullouisageneau, line 502:
```suggestion
		ret = mbedtls_ssl_write(&mSsl, reinterpret_cast<const unsigned char *>(message->data()), message->size());
```

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,308 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+					break;
+				}
+				switch (ret) {
+					case MBEDTLS_ERR_SSL_WANT_READ:
+					case MBEDTLS_ERR_SSL_WANT_WRITE:
+					case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
+					case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS:
+					case MBEDTLS_ERR_SSL_CLIENT_RECONNECT:
+						return;
+				}
```
paullouisageneau, line 591: I would use a simple if here for readability. Also, do all of those return codes make sense here? For instance, shouldn't the connection be closed instead on client reconnect?

paullouisageneau (commented):
I reviewed on left a bunch of comments, on the whole you did a great job on this, thank you for the contribution!

.gitmodules
Sean-Der: @paullouisageneau Done! Removed as submodule, and will install as system lib in github action.

CMakeLists.txt
Sean-Der: Fixed! Mbed TLS is not a submodule anymore, and use `MbedTLS::MbedTLS` properly.

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,308 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+}
+
+void DtlsTransport::stop() {
+	PLOG_DEBUG << "Stopping DTLS transport";
+	unregisterIncoming();
+	mIncomingQueue.stop();
+	enqueueRecv();
+}
+
+void DtlsTransport::SetTimerCallback(void *ctx, uint32_t int_ms, uint32_t fin_ms) {
```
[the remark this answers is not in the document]
Sean-Der: Moved!

Not a nitpick IMO :) you have to maintain this code, so happy to modify this PR in any way that makes it easier.

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,308 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+}
+
+void DtlsTransport::ExportKeysCallback(void *p_expkey,
+                                         mbedtls_ssl_key_export_type type,
+                                         const unsigned char *secret,
+                                         size_t secret_len,
+                                         const unsigned char client_random[32],
+                                         const unsigned char server_random[32],
+                                         mbedtls_tls_prf_types tls_prf_type) {
+	(void) type;
```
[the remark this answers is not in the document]
Sean-Der: Fixed! I didn't know you could just comment out the variable name, I have been doing it wrong :)

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,308 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+	PLOG_DEBUG << "Initializing DTLS transport (MbedTLS)";
+
+	if (!mCertificate)
+		throw std::invalid_argument("DTLS certificate is null");
+
+	mbedtls_entropy_init(&mEntropy);
+	mbedtls_ctr_drbg_init(&mDrbg);
+	mbedtls_ssl_init(&mSsl);
+	mbedtls_ssl_config_init(&mConf);
+	mbedtls_ctr_drbg_set_prediction_resistance(&mDrbg, MBEDTLS_CTR_DRBG_PR_ON);
```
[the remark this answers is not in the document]
Sean-Der: Added to certificate generation.

src/impl/certificate.cpp
```diff
@@ -149,7 +149,184 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
...
 
-#else // USE_GNUTLS==0
+#elif USE_MBEDTLS
+string make_fingerprint(shared_ptr<mbedtls_x509_crt> crt) {
+	uint8_t buffer[MBEDTLS_MD_MAX_SIZE];
+	std::stringstream fingerprint;
+
+	mbedtls::check(mbedtls_sha256(crt->raw.p, crt->raw.len, (unsigned char *) buffer, 0));
+
+	auto size = mbedtls_md_get_size(mbedtls_md_info_from_type(MBEDTLS_MD_SHA256));
```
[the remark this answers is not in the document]
Sean-Der: Switched to `size = 32`

I like defining the size up front better. When I read the code I don't have to assume/guess the return values

src/impl/certificate.cpp
```diff
@@ -149,7 +149,184 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
...
+
+	shared_ptr<mbedtls_pk_context> pk(
+			new mbedtls_pk_context,
+	    [](mbedtls_pk_context *p) {
+		    mbedtls_pk_free(p);
+		    delete p;
+	    });
+
+  mbedtls_x509_crt_init(crt.get());
+	mbedtls::check(mbedtls_x509_crt_parse(crt.get(), (const unsigned char *) crt_pem.c_str(), crt_pem.length()));
```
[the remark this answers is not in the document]
Sean-Der: Fixed! Created `mbedtls::new_x509_crt()` and `mbedtls::new_pk_context()`

src/impl/certificate.cpp
```diff
@@ -149,7 +149,184 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
...
+
+	shared_ptr<mbedtls_pk_context> pk(
+			new mbedtls_pk_context,
+	    [](mbedtls_pk_context *p) {
+		    mbedtls_pk_free(p);
+		    delete p;
+	    });
+
+	mbedtls_x509_crt_init(crt.get());
+	mbedtls::check(mbedtls_x509_crt_parse_file(crt.get(), crt_pem_file.c_str()));
```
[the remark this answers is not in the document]
Sean-Der: Fixed! Created `mbedtls::new_x509_crt()` and `mbedtls::new_pk_context()`

src/impl/dtlstransport.cpp
Sean-Der: `MBEDTLS_SSL_RENEGOTIATION_DISABLED` is the default, so we don't need to handle it ourselves. Converted the other statements to an if.

Sean-Der:
@paullouisageneau I believe I have addressed all the comments. Could I get another review please?

thank you!

pkviet:
Tried to build on visual studio 2022 without success.
I have include errors like:
```
libdatachannel\deps\libsrtp\crypto\include\aes_gcm.h(69,10): fatal  error C1083: Cannot open include file: 'mbedtls/aes.h': No such file or directory
libdatachannel\deps\libsrtp\crypto\cipher\aes_icm_mbedtls.c(48,10): fatal  error C1083: Cannot open include file: 'mbedtls/aes.h': No such file or directory
libdatachannel\deps\libsrtp\crypto\cipher\aes_gcm_mbedtls.c(49,10): fatal  error C1083: Cannot open include file: 'mbedtls/gcm.h': No such file or directory
libdatachannel\deps\libsrtp\crypto\hash\hmac_mbedtls.c(52,10): fatal  error C1083: Cannot open include file: 'mbedtls/md.h': No such file or directory
```

pkviet:
Hi @Sean-Der 
I first tested through Colin's pr on obs-deps.
It failed for me and someone else, the error requested a flag to enable c++20. Tried that to no avail. Didn't fix anything.
After that failure, I tried building directly your libdatachannel PR on visual studio without modifications and came up with the include errors in libsrtp in addition to c++20 errors. 
I've tried to find fixes but was unsuccessful.
So I've told Colin who'll try to debug tonight.

pkviet:
Oh great @Sean-Der !

Sean-Der:
ok pushed, @pkviet would you mind trying 10ad7beeff6218280f6c6521d40230379858c2e5

This is the change I made. The libsrtp CMakeLists was effected by our finder

```
diff --git a/cmake/Modules/FindMbedTLS.cmake b/cmake/Modules/FindMbedTLS.cmake
index 3b13876..a82ba04 100644
--- a/cmake/Modules/FindMbedTLS.cmake
+++ b/cmake/Modules/FindMbedTLS.cmake
@@ -193,9 +193,11 @@ if(MbedTLS_FOUND)

   if(MbedTLS_INCLUDES_X509 AND MbedTLS_INCLUDES_CRYPTO)
     set(MbedTLS_LIBRARIES ${MbedTLS_LIBRARY})
+    set(MBEDTLS_INCLUDE_DIRS ${MbedTLS_INCLUDE_DIR})
   else()
     set(MbedTLS_LIBRARIES ${MbedTLS_LIBRARY} ${MbedCrypto_LIBRARY} ${MbedX509_LIBRARY})
     set_property(TARGET MbedTLS::MbedTLS PROPERTY INTERFACE_LINK_LIBRARIES MbedTLS::MbedCrypto MbedTLS::MbedX509)
+    set(MBEDTLS_INCLUDE_DIRS ${MbedTLS_INCLUDE_DIR})
   endif()
 endif()
```

pkviet:
Will do, will report here and on obs discord.

pkviet:
so the include errors for libsrtp are gone but i still have a bunch of these:
`5>certificate.cpp
5>L:\visual-studio\obs-deps\windows\obs-deps-x64\include\psa\crypto_extra.h(1948,43): error C7555: use of designated initializers requires at least '/std:c++20' `
These includes come from mbedtls.

DDRBoxman:
Okay, so some updates about the build errors for OBS deps on windows.

Just for kicks I tried mbedtls 3.2.1, but it has compile issues in certificate.cpp which I assume is related to the recent change Sean made.
`\libdatachannel\src\impl\certificate.cpp(157,17): error C3861: 'mbedtls_sha256': identifier not found`
It does not show the designated initializers issue though.

Turns out that the designated initializers being an issue in C++ projects is a known issue in mbedtls that might get fixed next quarter https://github.com/Mbed-TLS/mbedtls/issues/7087

Sean-Der:
@DDRBoxman is that the only error it has with `mbedtls_sha256`? That feels really close?

Maybe we are passing the wrong types, are that isn't enabled?

cmake/Modules/FindMbedTLS.cmake
```diff
@@ -0,0 +1,210 @@
...
+  check_symbol_exists(mbedtls_sha256_init "mbedtls/sha256.h" MbedTLS_INCLUDES_CRYPTO)
+  unset(CMAKE_REQUIRED_LIBRARIES)
+  unset(CMAKE_REQUIRED_INCLUDES)
+endif()
+
+if(CMAKE_HOST_SYSTEM_NAME MATCHES "Darwin|Windows")
+  set(MbedTLS_ERROR_REASON "Ensure that obs-deps is provided as part of CMAKE_PREFIX_PATH.")
+elseif(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux|FreeBSD")
+  set(MbedTLS_ERROR_REASON "Ensure that MbedTLS is installed on the system.")
+endif()
```
tytan652, line 155:
```suggestion
```

cmake/Modules/FindMbedTLS.cmake
```diff
@@ -0,0 +1,210 @@
...
+    MbedTLS
+    REQUIRED_VARS MbedTLS_LIBRARY MbedTLS_INCLUDE_DIR
+    VERSION_VAR MbedTLS_VERSION REASON_FAILURE_MESSAGE "${MbedTLS_ERROR_REASON}")
+  mark_as_advanced(MbedTLS_LIBRARY MbedTLS_INCLUDE_DIR)
+  list(APPEND _COMPONENTS TLS)
+else()
+  find_package_handle_standard_args(
+    MbedTLS
+    REQUIRED_VARS MbedTLS_LIBRARY MbedCrypto_LIBRARY MbedX509_LIBRARY MbedTLS_INCLUDE_DIR
+    VERSION_VAR MbedTLS_VERSION REASON_FAILURE_MESSAGE "${MbedTLS_ERROR_REASON}")
```
tytan652, line 168:
```suggestion
    VERSION_VAR MbedTLS_VERSION)
```

cmake/Modules/FindMbedTLS.cmake
```diff
@@ -0,0 +1,210 @@
...
+  mark_as_advanced(MbedTLS_LIBRARY MbedCrypto_LIBRARY MbedX509_LIBRARY MbedTLS_INCLUDE_DIR)
+  list(APPEND _COMPONENTS TLS Crypto X509)
+endif()
+unset(MbedTLS_ERROR_REASON)
```
tytan652, line 172:
```suggestion
```

tytan652 (changes requested):
`REASON_FAILURE_MESSAGE` is not needed.

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,303 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+		// Handle handshake if connecting
+		if (state() == State::Connecting) {
+			auto ret = mbedtls_ssl_handshake(&mSsl);
+
+			if (ret == MBEDTLS_ERR_SSL_WANT_READ ||
+					ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+					ret == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS ||
+					ret == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS)
+			{
+				auto timeout = milliseconds(this->mFin_ms);
```
paullouisageneau, line 529: Shouldn't the timeout be set to the remaining duration by accounting for time already elapsed since `mTimerSetAt`?

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,303 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+	do {
+		std::lock_guard lock(mMutex);
+		mCurrentDscp = message->dscp;
+
+		if (message->size() > size_t(mbedtls_ssl_get_max_out_record_payload(&mSsl)))
+			return false;
+
+		ret = mbedtls_ssl_write(&mSsl, reinterpret_cast<const unsigned char *>(message->data()), message->size());
+	} while (ret == MBEDTLS_ERR_SSL_WANT_WRITE);
+	mbedtls::check(ret, "Failed writing to Mbed TLS Context");
```
paullouisageneau, line 475:
```suggestion
	mbedtls::check(ret);
```
(for consistency with other implementations and TLS)

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,303 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+			{
+				auto timeout = milliseconds(this->mFin_ms);
+				ThreadPool::Instance().schedule(timeout, [weak_this = weak_from_this()]() {
+					if (auto locked = weak_this.lock())
+						locked->doRecv();
+				});
+				return;
+			}
+
+			mbedtls::check(ret, "Failed Mbed DTLS Handshake");
```
paullouisageneau, line 537:
```suggestion
			mbedtls::check(ret, "DTLS handshake failed");
```

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,303 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+				}
+
+				if (ret == MBEDTLS_ERR_SSL_WANT_READ ||
+						ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+						ret == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS ||
+						ret == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS)
+				{
+					return;
+				}
+				mbedtls::check(ret, "Failed Mbed DTLS Read");
```
paullouisageneau, line 562:
```suggestion
				mbedtls::check(ret);
```

src/impl/dtlstransport.hpp
```diff
@@ -70,7 +70,35 @@ class DtlsTransport : public Transport, public std::enable_shared_from_this<Dtls
...
+
+#elif USE_MBEDTLS
+	std::mutex mMutex;
+
+	mbedtls_entropy_context mEntropy;
+	mbedtls_ctr_drbg_context mDrbg;
+	mbedtls_ssl_config mConf;
+	mbedtls_ssl_context mSsl;
+
+	uint32_t mFin_ms, mInt_ms;
```
paullouisageneau, line 82: I'd prefer `mFinMs` and `mIntMs` to prevent mixing camel-case and snake-case.

src/impl/tlstransport.cpp
```diff
@@ -292,7 +292,223 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+					t->recv(message); // Pass zero-sized messages through
+			}
+		}
+
+		if (message) {
+			size_t available = message->size() - position;
+			ssize_t writeLen = std::min((unsigned long) len, available);
+			std::memcpy(buf, message->data() + position, writeLen);
+			position += len;
+			return len;
```
paullouisageneau, line 498: I think the wrong size if added and returned.
```suggestion
			size_t available = message->size() - position;
			size_t writeLen = std::min(len, available);
			std::memcpy(buf, message->data() + position, writeLen);
			position += writeLen;
			return writeLen;
```

src/impl/dtlstransport.hpp
Sean-Der: Fixed! Renamed

src/impl/dtlstransport.cpp
Sean-Der: Yep you are right! Took into account the elapsed time

src/impl/certificate.cpp
```diff
@@ -149,7 +148,149 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
...
+		mbedtls_x509write_crt_set_subject_key(&wcrt, pk.get());
+		mbedtls_x509write_crt_set_issuer_key(&wcrt, pk.get());
+		mbedtls_x509write_crt_set_md_alg(&wcrt, MBEDTLS_MD_SHA256);
+
+		const size_t certificateBufferSize = 4096;
+		unsigned char certificateBuffer[certificateBufferSize];
+		std::memset(certificateBuffer, 0, certificateBufferSize);
+
+		auto certificateLen = mbedtls_x509write_crt_der(&wcrt, certificateBuffer, certificateBufferSize, mbedtls_ctr_drbg_random, &drbg);
+		if (certificateLen <= 0){
```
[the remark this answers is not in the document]
Sean-Der: Ran clang-format

Sean-Der:
Hey @paullouisageneau 

Addressed all the issues and rebased, mind giving another review? thanks!

jordo:
(Just a comment from the sidelines), but thanks for all the hard work on this everyone BTW... I'm really looking forward to trying this out in our games!

Sean-Der:
@paullouisageneau anything else I can do/change?

Also after this PR would you be interested in me adding sanitizers? Running TSAN locally I noticed some things around logging etc...

Sean-Der:
Hey @paullouisageneau 

Would it help to get another reviewer? Anyone else involved in libdatachannel I should ping?

thanks!

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+		// Handle handshake if connecting
+		if (state() == State::Connecting) {
+			auto handshakeStatus = mbedtls_ssl_handshake(&mSsl);
+			switch (handshakeStatus) {
+			case 0:
+				break;
+			case MBEDTLS_ERR_SSL_WANT_WRITE:
+			case MBEDTLS_ERR_SSL_WANT_READ:
+			case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
+			case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS: {
```
paullouisageneau, line 420: Could you please refactor this the same was as in `DtlsTransport`?

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+					PLOG_DEBUG << "TLS connection cleanly closed";
+					break;
+				}
+				switch (ret) {
+				case MBEDTLS_ERR_SSL_WANT_READ:
+				case MBEDTLS_ERR_SSL_WANT_WRITE:
+				case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
+				case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS:
+				case MBEDTLS_ERR_SSL_CLIENT_RECONNECT:
+					return;
```
paullouisageneau, line 451: Could you please refactor this the same was as in `DtlsTransport`?

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+				break;
+			case MBEDTLS_ERR_SSL_WANT_WRITE:
+			case MBEDTLS_ERR_SSL_WANT_READ:
+			case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
+			case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS: {
+				auto timeout = milliseconds(100);
+				ThreadPool::Instance().schedule(timeout, [weak_this = weak_from_this()]() {
+					if (auto locked = weak_this.lock())
+						locked->doRecv();
+				});
```
paullouisageneau, line 425: This should be removed as the connection is reliable and there are no retransmissions here.

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+	changeState(State::Disconnected);
+	recv(nullptr);
+}
+
+int TlsTransport::WriteCallback(void *ctx, const unsigned char *buf, size_t len) {
+	auto *t = static_cast<TlsTransport *>(ctx);
+	auto *b = reinterpret_cast<const byte *>(buf);
+	if (!t->outgoing(make_message(b, b + len))) {
+		return MBEDTLS_ERR_SSL_WANT_READ;
+	}
```
paullouisageneau, line 473:
```suggestion
	t->outgoing(make_message(b, b + len));
```

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+		enqueueRecv();
+		return;
+	}
+
+	PLOG_VERBOSE << "Incoming size=" << message->size();
+	mIncomingQueue.push(message);
+	enqueueRecv();
+}
+
+bool TlsTransport::outgoing(message_ptr message) { return Transport::outgoing(std::move(message)); }
```
paullouisageneau, line 394:
```suggestion
bool TlsTransport::outgoing(message_ptr message) {
	bool result = Transport::outgoing(std::move(message));
	mOutgoingResult = result;
	return result;
}
```

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,224 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+
+	if (!message || message->size() == 0)
+		return outgoing(message); // pass through
+
+	PLOG_VERBOSE << "Send size=" << message->size();
+
+	mbedtls::check(mbedtls_ssl_write(
+	    &mSsl, reinterpret_cast<const unsigned char *>(message->data()), int(message->size())));
+
+	return true;
```
paullouisageneau, line 379:
```suggestion
	return mOutgoingResult;
```

paullouisageneau:
Here is my (hopefully last) review, I think after addressing the comments this should be good to go.

> Also after this PR would you be interested in me adding sanitizers? Running TSAN locally I noticed some things around logging etc...

Yes sure!

> Also after this is merged will it be tagged? I am happy to grab some other bugs/issues to warrant a new release. The OBS build system requires that it be tagged.

Sure, I'll create a new release with it. Are there other changes needed for OBS?

src/impl/certificate.cpp
```diff
@@ -149,7 +147,173 @@ string make_fingerprint(gnutls_x509_crt_t crt) {
...
+			                                   &drbg, nbits, exponent),
+			               "Unable to generate RSA key pair");
+			break;
+		}
+		default:
+			throw std::invalid_argument("Unknown certificate type");
+		}
+
+		auto now =
+		    std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
```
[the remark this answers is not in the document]
Sean-Der: Removed cast

src/impl/dtlstransport.hpp
```diff
@@ -70,7 +70,33 @@ class DtlsTransport : public Transport, public std::enable_shared_from_this<Dtls
...
+
+#elif USE_MBEDTLS
+	std::mutex mMutex;
+
+	mbedtls_entropy_context mEntropy;
+	mbedtls_ctr_drbg_context mDrbg;
+	mbedtls_ssl_config mConf;
+	mbedtls_ssl_context mSsl;
+
+	uint32_t mFinMs, mIntMs;
```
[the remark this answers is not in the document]
Sean-Der: done

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,305 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+		char buffer[bufferSize];
+
+		// Handle handshake if connecting
+		if (state() == State::Connecting) {
+			auto ret = mbedtls_ssl_handshake(&mSsl);
+
+			if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+			    ret == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS ||
+			    ret == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS) {
+				auto elapsed = std::chrono::steady_clock::now() - mTimerSetAt;
```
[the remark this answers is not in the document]
Sean-Der: Done! Condensed into one liner

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,305 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+}
+
+int DtlsTransport::GetTimerCallback(void *ctx) {
+	auto dtlsTransport = static_cast<DtlsTransport *>(ctx);
+	auto elapsed = std::chrono::steady_clock::now() - dtlsTransport->mTimerSetAt;
+	auto ms = uint32_t(std::chrono::duration_cast<std::chrono::milliseconds>(elapsed).count());
+
+	if (dtlsTransport->mFinMs == 0) {
+		return -1;
+	} else if (ms >= dtlsTransport->mFinMs) {
```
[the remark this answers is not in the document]
Sean-Der: Fixed!

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,305 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+		return 2;
+	} else if (ms >= dtlsTransport->mIntMs) {
```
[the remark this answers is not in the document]
Sean-Der: Fixed!

src/impl/tlstransport.cpp
Sean-Der: Fixed!

src/impl/tlstransport.cpp
Sean-Der: I believe we still need this @paullouisageneau when running locally I would occasionally see handshake failures without it.  

```
   * \return         #MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS if a cryptographic
   *                 operation is in progress (see mbedtls_ecp_set_max_ops()) -
   *                 in this case you must call this function again to complete
   *                 the handshake when you're done attending other tasks.
```

Sean-Der:
Ok addressed all comments @paullouisageneau 

re: tagging a release. We have a little bit of a chicken and egg problem

* I need to get a version into obs-deps so people can test
* I can't fully know if things are going to work across all platforms, so might need to make more libdatachannel changes

I would like to get a release so I can move ahead with OBS, but I might need to come back with other fixes if I hit something. Would you want to do a beta or rc tag? Does that make things easier on your side

RytoEX:
Following up on [this review thread](https://github.com/paullouisageneau/libdatachannel/pull/822#discussion_r1132629334), posting as a standalone comment for visibility:
> Following up on this, I noticed that the finder was initially pulled from the linked repository/PR. While the finder included here has since been modified, if an explicitly licensed file is preferred as the basis of the work here (the finder previously linked to did not have an explicit license), you can find MIT-licensed finders [here](https://github.com/PatTheMav/cmake-finders).


> * I need to get a version into obs-deps so people can test

On the obs-deps side, our WIP PR for this effort is currently just pulling in a commit from your fork instead of from this repo. We will continue to do that until we no longer need to.

CMakeLists.txt
Sean-Der: @RytoEX Fixed! Thanks for calling that out.

Sean-Der:
@RytoEX Fixed the Mbed TLS finder. The diff contained no functional changes, but now attributes properly.

src/impl/tlstransport.cpp
paullouisageneau: Calling again after an arbitrary timeout is kind of a hack and could indicate a return code is not handled as expected currently. Do you still see handshake failures after fixing `WriteCallback` not to incorrectly return `MBEDTLS_ERR_SSL_WANT_READ`? This could have caused the reception to wait for input and stall instead of reading available incoming data.

If I understand the doc correctly, the `X_IN_PROGRESS` functions require the user to call again as soon as possible, so there should be no wait in those cases (same for DTLS). Also, it sounds like `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` may only be returned if `mbedtls_ecp_set_max_ops()` was called to limit the number of operations before interrupting as the default is unlimited, so I guess it's never returned here.

paullouisageneau:
> I would like to get a release so I can move ahead with OBS, but I might need to come back with other fixes if I hit something. Would you want to do a beta or rc tag? Does that make things easier on your side

@Sean-Der A beta tag sounds fine to me!

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,300 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+				return MBEDTLS_ERR_SSL_WANT_READ;
+			}
+
+			message_ptr message = std::move(*next);
+			if (t->demuxMessage(message))
+				continue;
+
+			auto bufMin = std::min(len, size_t(message->size()));
+			std::memcpy(buf, message->data(), bufMin);
+			return len;
```
paullouisageneau, line 618:
```suggestion
			return int(len);
```

src/impl/tlstransport.cpp
Sean-Der: I still failures after the `WriteCallback` changed

I updated the code to call again immediately on the suggest return codes (instead of timing out)

Sean-Der:
@paullouisageneau done! thank you so much for the fast reviews

src/impl/dtlstransport.cpp
```diff
@@ -360,7 +360,300 @@ int DtlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* m
...
+
+bool DtlsTransport::send(message_ptr message) {
+	if (!message || state() != State::Connected)
+		return false;
+
+	PLOG_VERBOSE << "Send size=" << message->size();
+
+	int ret;
+	do {
+		std::lock_guard lock(mMutex);
```
paullouisageneau, line 470: The same mutex is missing from `TlsTransport`. It looks like Mbed TLS indeed can't send and receive concurrently on the same session, which is quite a pain.

src/impl/tlstransport.cpp
```diff
@@ -296,7 +296,218 @@ int TlsTransport::TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int /* ms
...
+		char buffer[bufferSize];
+
+		// Handle handshake if connecting
+		if (state() == State::Connecting) {
+			while (true) {
+				auto ret = mbedtls_ssl_handshake(&mSsl);
+				if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+				    ret == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS ||
+				    ret == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS) {
+					continue;
```
paullouisageneau, line 422: You must not call again on `WANT_READ` or `WANT_WRITE` as it would spin when no incoming data is available.

As I understand, you should call again immediately on `X_IN_PROGRESS` and return on `WANT_READ` or `WANT_WRITE` (while scheduling after the timeout for DTLS).

This should be the behavior for each call to `mbedtls_ssl_handshake` and `mbedtls_ssl_read`.

src/impl/tlstransport.hpp
```diff
@@ -65,6 +65,21 @@ class TlsTransport : public Transport, public std::enable_shared_from_this<TlsTr
 	static ssize_t WriteCallback(gnutls_transport_ptr_t ptr, const void *data, size_t len);
 	static ssize_t ReadCallback(gnutls_transport_ptr_t ptr, void *data, size_t maxlen);
 	static int TimeoutCallback(gnutls_transport_ptr_t ptr, unsigned int ms);
+#elif USE_MBEDTLS
+	std::mutex mSendMutex;
```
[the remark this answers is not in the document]
Sean-Der: Fixed. Updated with locking pattern from DTLS

src/impl/dtlstransport.cpp
Sean-Der: Fixed. Updated with locking pattern from DTLS

src/impl/tlstransport.cpp
Sean-Der: @paullouisageneau If I remove those all websocket tests fail with. I also get DTLS failures of the same sort

```
2023-03-29 21:09:26.016 ERROR [130773] [rtc::mbedtls::check@106] MbedTLS error: SSL - No data of requested type currently available on underlying transport
2023-03-29 21:09:26.016 ERROR [130773] [rtc::impl::TlsTransport::doRecv@458] TLS recv: MbedTLS error: SSL - No data of requested type currently available on underlying transport
```

paullouisageneau: Hum, `SSL - No data of requested type currently available on underlying transport` is actually `MBEDTLS_ERR_SSL_WANT_READ`, so it would mean `MBEDTLS_ERR_SSL_WANT_READ` ends up in `mbedtls::check()` instead of triggering a return.

paullouisageneau:
I think the last issue is with `mbedtls_ssl_handshake` and `mbedtls_ssl_read` return value handling [here](https://github.com/paullouisageneau/libdatachannel/pull/822#discussion_r1152465327), otherwise it looks good to me!

After fixing the issue, I'll merge it and create a tag.

Sean-Der:
Ok fixed! @paullouisageneau fingers crossed I think I got it all right :D 

Thanks for all your patience on your PR. Getting WHIP into OBS is going to be so great for people I can't wait.

If someone can broadcast to their friends/co-workers/x$ without setting up a expensive server it is going to be massive for people. Really excited.

paullouisageneau:
> Ok fixed! @paullouisageneau fingers crossed I think I got it all right :D

@Sean-Der Looks good, could you just fix the DTLS handshake and read too before merging?

Thank you for your work and patience! WHIP in OBS is indeed very exciting, but I know Mbed TLS support will also unlock new possibilities for a lot of people.

Sean-Der:
@paullouisageneau done! I think I got it this time :D

paullouisageneau (approved):
It looks good, thank you!

paullouisageneau:
@Sean-Der Tagged as ` 0.19.0-alpha.1`

pkviet:
Thanks from obs devs. @paullouisageneau

PR #827, opened 2023-03-14
Websocket Proxy Support

web2098:
Hi,

In my use it looks like I will need proxy support on the websocket, so I created this initial implementation. I don't know much about proxy servers, but from my internet searching and testing this appears to work. 

I created a new Transport class called TcpProxyTransport that I insert into the websockets transport "stack" after the initial establishment of the tcp connection and between adding TLS support and/or WS upgrade. This TcpProxyTransport class will send a connect request on HTTP and checks for a response with status 200.

In addition I moved some of the http parsing functions from WSHandshake into the utils file to allow the new TcpProxyTransport to share use of the code.


src/impl/tlstransport.cpp
```diff
@@ -58,10 +59,15 @@ void TlsTransport::Cleanup() {
...
+TlsTransport::TlsTransport(variant<shared_ptr<TcpTransport>, shared_ptr<TcpProxyTransport>> lower, optional<string> host,
                            certificate_ptr certificate, state_callback callback)
-    : Transport(lower, std::move(callback)), mHost(std::move(host)), mIsClient(lower->isActive()),
-      mIncomingQueue(RECV_QUEUE_LIMIT, message_size_func) {
+    : Transport(std::visit([](auto l) { return std::static_pointer_cast<Transport>(l); }, lower),
+			    std::move(callback)), mHost(std::move(host))
+	, mIsClient(
+		std::visit(rtc::overloaded{[](shared_ptr<TcpTransport> l) { return l->isActive(); },
+                                   [](shared_ptr<TcpProxyTransport> l) { return l->isActive(); }},
+                   lower))
```
paullouisageneau, line 69: You should be able to simplify this with:
```suggestion
	, mIsClient(std::visit([](auto l) { return l->isActive(); }, lower))
```

src/impl/tlstransport.cpp
```diff
@@ -308,10 +314,15 @@ void TlsTransport::Cleanup() {
...
+TlsTransport::TlsTransport(variant<shared_ptr<TcpTransport>, shared_ptr<TcpProxyTransport>> lower, optional<string> host,
                            certificate_ptr certificate, state_callback callback)
-    : Transport(lower, std::move(callback)), mHost(std::move(host)), mIsClient(lower->isActive()),
-      mIncomingQueue(RECV_QUEUE_LIMIT, message_size_func) {
+    : Transport(std::visit([](auto l) { return std::static_pointer_cast<Transport>(l); }, lower),
+				std::move(callback)), mHost(std::move(host))
+	, mIsClient(
+          std::visit(rtc::overloaded{[](shared_ptr<TcpTransport> l) { return l->isActive(); },
+                                     [](shared_ptr<TcpProxyTransport> l) { return l->isActive(); }},
+                     lower))
```
paullouisageneau, line 324:
```suggestion
	, mIsClient(std::visit([](auto l) { return l->isActive(); }, lower))
```

src/impl/utils.hpp
```diff
@@ -32,6 +34,12 @@ string base64_encode(const binary &data);
 // Return a random seed sequence
 std::seed_seq random_seed();
 
+// Parse an http packet into lines
```
paullouisageneau, line 37:
```suggestion
// Parse an HTTP message into lines
```

src/impl/websocket.cpp
```diff
@@ -250,16 +263,73 @@ shared_ptr<TcpTransport> WebSocket::setTcpTransport(shared_ptr<TcpTransport> tra
...
 			return transport;
 
-		auto lower = std::atomic_load(&mTcpTransport);
-		if (!lower)
-			throw std::logic_error("No underlying TCP transport for TLS transport");
+		variant<shared_ptr<TcpTransport>, shared_ptr<TcpProxyTransport>> lower;
+		if (mIsProxied) {
+			auto transport = std::atomic_load(&mProxyTransport);
+			if (!transport)
+				throw std::logic_error("No underlying TLS transport for WebSocket transport");
```
paullouisageneau, line 323:
```suggestion
				throw std::logic_error("No underlying proxy transport for TLS transport");
```

src/impl/websocket.cpp
```diff
@@ -250,16 +263,73 @@ shared_ptr<TcpTransport> WebSocket::setTcpTransport(shared_ptr<TcpTransport> tra
...
+
+			lower = transport;
+		} else {
+			auto transport = std::atomic_load(&mTcpTransport);
+			if (!transport)
+				throw std::logic_error("No underlying TCP transport for WebSocket transport");
```
paullouisageneau, line 329:
```suggestion
				throw std::logic_error("No underlying TCP transport for TLS transport");
```

src/impl/utils.hpp
```diff
@@ -13,6 +13,8 @@
 
 #include <climits>
 #include <limits>
+#include <list>
+#include <map>
```
paullouisageneau, line 17: You may clean up the includes in `src/impl/wshandshake.hpp`.

src/impl/websocket.cpp
```diff
@@ -48,7 +49,8 @@ void WebSocket::open(const string &url) {
 		throw std::logic_error("WebSocket must be closed before opening");
 
 	if (config.proxyServer) {
-		PLOG_WARNING << "Proxy server support for WebSocket is not implemented";
+		mIsProxied = true;
+		mProxy = config.proxyServer;
```
paullouisageneau, line 53: You should not need to copy the field as `config` is actually a member. Also, you might use directly `config.proxyServer` instead of introducing the `mIsProxied` state.

src/impl/websocket.cpp
```diff
@@ -101,11 +103,20 @@ void WebSocket::open(const string &url) {
...
+	mHostname = hostname; // for TLS SNI and Proxy
+	mService = service; //For proxy
 	std::atomic_store(&mWsHandshake, std::make_shared<WsHandshake>(host, path, config.protocols));
 
 	changeState(State::Connecting);
-	setTcpTransport(std::make_shared<TcpTransport>(hostname, service, nullptr));
+
+	if (mIsProxied)
+	{
+		setTcpTransport(std::make_shared<TcpTransport>(mProxy.value().hostname, std::to_string(mProxy.value().port), nullptr));
```
paullouisageneau, line 114:
```suggestion
		setTcpTransport(std::make_shared<TcpTransport>(config.proxyServer->hostname, to_string(config.proxyServer->port), nullptr));
```

paullouisageneau (commented):
Thank you for implementing the feature. I added a bunch of comments but overall it looks good, I think implementing it with a transport is indeed the way to go here.

src/impl/websocket.cpp
```diff
@@ -250,16 +258,73 @@ shared_ptr<TcpTransport> WebSocket::setTcpTransport(shared_ptr<TcpTransport> tra
...
+		auto stateChangeCallback = [this, weak_this = weak_from_this()](State transportState) {
+			auto shared_this = weak_this.lock();
+			if (!shared_this)
+				return;
+			switch (transportState) {
+			case State::Connected:
+				if (mIsSecure)
+					initTlsTransport();
+				else
+					initWsTransport();
```
paullouisageneau, line 281: Unless I'm mistaken, `initWsTransport()` is not modified to pick up the proxy transport so it'll instead use `TcpTransport` as lower layer.

paullouisageneau (commented):
It looks good, thanks!

src/impl/websocket.cpp
web2098: Nice catch! I have updated the code to supply the http proxy transport as another option in the variant.

paullouisageneau (approved):

Issue #837, opened 2023-03-26
WebSocket socks5 proxy deadlock on Windows


shithappens2016:
![deadlock_on_win10](https://user-images.githubusercontent.com/13523412/227782615-825fce6c-f934-4fbd-a9d9-2110e30d10c8.png)

```c++
  rtc::WebSocket::Configuration cfg;
  cfg.proxyServer = {"socks5://127.0.0.1:6666"};

  auto ws = std::make_shared<rtc::WebSocket>(cfg);

  std::promise<void> wsPromise;
  auto wsFuture = wsPromise.get_future();

  ws->onOpen([&ws]() {
    std::cout << "WebSocket connected, signaling ready" << std::endl;
    ws->send("something");
  });

  ws->onError([&wsPromise](std::string s) {
    std::cout << "WebSocket error" << std::endl;
    wsPromise.set_exception(std::make_exception_ptr(std::runtime_error(s)));
  });

  ws->onClosed([&wsPromise]() {
    std::cout << "WebSocket closed" << std::endl;
    wsPromise.set_value();
  });

  ws->onMessage([](auto data) {
    std::cout << "ws msg\n";
  });

  ws->open("wss://tracker.openwebtorrent.com/");
  wsFuture.get();
```

@web2098 
@paullouisageneau

paullouisageneau:
Sock5 proxies are not supported for Websockets, only HTTP proxies. 

Are you sure there is a deadlock? I would expect the connection to hang. The issue is that PR https://github.com/paullouisageneau/libdatachannel/pull/827 adding support for HTTP proxy does not introduce a check for the proxy type so it will always try to use the host as HTTP proxy @web2098

PR #839, opened 2023-03-27
Improve error handling for unsupported proxies

web2098:
This PR cleans up a side-effect from the PR #827, that is descripted in issue #837. 

This PR modifies the open function to throw an exception if a non-http proxy is supplied in the WebSocket configuration. In addition I added a warning that username + password are not supported on the http proxy for now.

### DIFF
--- a/src/impl/websocket.cpp
+++ b/src/impl/websocket.cpp
@@ -51,8 +51,9 @@ void WebSocket::open(const string &url) {
 	if (config.proxyServer) {		
 		if( config.proxyServer->type == ProxyServer::Type::Socks5)
 			throw std::invalid_argument("Proxy server support for WebSocket is not implemented for Socks5");
-		if (config.proxyServer->username || config.proxyServer->password)
+		if (config.proxyServer->username || config.proxyServer->password) {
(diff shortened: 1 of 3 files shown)


src/impl/websocket.cpp
```diff
@@ -47,6 +47,14 @@ void WebSocket::open(const string &url) {
...
 	if (state != State::Closed)
 		throw std::logic_error("WebSocket must be closed before opening");
+	
+	if (config.proxyServer) {		
+		if( config.proxyServer->type == ProxyServer::Type::Socks5)
+			throw std::invalid_argument("Proxy server support for WebSocket is not implemented for Socks5");
+		if (config.proxyServer->username || config.proxyServer->password) {
+			PLOG_WARNING << "HTTP authentication support for proxy is not implemented";
+		}
+	}
```
paullouisageneau, line 57: Could you please move the check to the constructor? As it takes the configuration, it would make sense to validate it there (That's what `PeerConnection` does).

paullouisageneau (approved):
Thank you for this improvement!

PR #851, opened 2023-04-06
Change assert to allow empty messages to be forwarded downstream

omigamedev:
This will allow chain elements such as a defragmenter to buffer fragmented messages in order to forward the whole message once all the fragments are available.


paullouisageneau (approved):
Thank you for the PR!

PR #855, opened 2023-04-18
Fix RtcpSr and RtcpRr header mixup when dispatching

tljdebrouwer:
RtcpSr and RtcpRr do not have the same header struct, therefore they cannot be used in a single cast. Split these.


paullouisageneau (approved):
Thank you for the fix!

paullouisageneau:
Merged in `v0.18` and `master`

PR #856, opened 2023-04-19
fix juice debug level

kennethanderson:
This fixes the juice debug level when libdatachannel's plog level is set to fatal.


paullouisageneau (approved):
Thanks for the fix!!

PR #880, opened 2023-05-11
Fix race condition in `Processor`

themightyoarfish:


paullouisageneau:
This should not change the behavior as the `mPending` bool is only accessed with locked mutex ([here](https://github.com/paullouisageneau/libdatachannel/blob/3fd1b1a2ac24a4a95f6a063ea8be518b7b01f9fa/src/impl/processor.hpp#L66), [here](https://github.com/paullouisageneau/libdatachannel/blob/3fd1b1a2ac24a4a95f6a063ea8be518b7b01f9fa/src/impl/processor.cpp#L19), and [here](https://github.com/paullouisageneau/libdatachannel/blob/3fd1b1a2ac24a4a95f6a063ea8be518b7b01f9fa/src/impl/processor.cpp#L28)).

Issue #891, opened 2023-05-20
Simulcast Sender


Sean-Der:
Hey @paullouisageneau 

I would like to add the ability to send Simulcast from libdatachannel. Would you accept this feature? Happy to type up justifications if you aren't sure if it is worth it yet. Do you have any opinions on API? I see these options.

### Add a new constructor to `Media`
Users could construct multiple `Media`s with different RIDs. Then would call `Send` as usual.

```
rtc::Description::Video highQuality("video", rtc::Description::Direction::SendOnly, "h");
rtc::Description::Video mediumQuality("video", rtc::Description::Direction::SendOnly, "m");
rtc::Description::Video lowQuality("video", rtc::Description::Direction::SendOnly, "l");

auto highQualityTrack = pc->addTrack(highQuality);
highQualityTrack->send(buffer, len);
```

### Express multiple RIDs on one Track or Media
```
rtc::Description::Video video("video", rtc::Description::Direction::SendOnly);

auto highRid = video->addRid("h");
auto mediumRid = video->addRid("m");
auto lowRid = video->addRid("l");

auto videoTrack = pc->addTrack(video);
videoTrack->send(buffer, len, highRid);
```

paullouisageneau:
It looks good! Feel free to open a PR with the changes to `rtc::Description::Video`, I'd be happy to review it.

You might want to add a new function to the C API, for instance `rtcAddTrackRid(int tr, const char *rid)`, since WHIP in OBS relies on the C API.

Sean-Der:
@paullouisageneau 

Unfortunately the new API doesn't work for non-packetized media :/ 

If I only have access to one track handle how do I set MID/RID on it? If I packetize myself I am able to send media though.

Would you suggest that I do packetization myself and not use libdatachannel's here?

paullouisageneau:
Ah yes, the packetizer is independent and you must fill the `RtpPacketizationConfig` and update it. I think you'd need to add something similar to video orientation for both rid and mid, and update the rid before sending: 
https://github.com/paullouisageneau/libdatachannel/blob/1ead0ce3040fb2c08fba3919bb503e6ee78b025d/include/rtc/rtppacketizationconfig.hpp#L54

This current design is not great and comes from when the packetizer was contributed. I think it should be more integrated to the track object: it should read media settings from the description and per-frame metadata should come with the frame, possibly with a dedicated send method. It's something I'd need to work on.

Sean-Der:
[...]

PR #903, opened 2023-06-05
Add Simulcast support

Sean-Der:
This allow senders to send multiple quality levels as one stream. 

I also added an example. It just works against the public WHIP/WHEP server I run. I will adjust to w/e you think is best @paullouisageneau. I need to make everything argv driven, but currently you can 

* run example
* `ffmpeg -re -f lavfi -i testsrc=size=640x480:rate=30 -pix_fmt yuv420p -c:v libx264 -g 10 -preset ultrafast -tune zerolatency -f rtp -payload_type 96 'rtp://127.0.0.1:6000?pkt_size=1200`
* Load https://b.siobud.com/seanTest

Then you will see a drop down where you can select different quality levels.

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -532,6 +532,10 @@ void Description::addAttribute(string attr) {
 		mAttributes.emplace_back(std::move(attr));
 }
 
+void Description::Entry::addRid(string rid) {
+	mRids.emplace_back(rid);
+}
+
 void Description::removeAttribute(const string &attr) {
 	mAttributes.erase(
 	    std::remove_if(mAttributes.begin(), mAttributes.end(),
@@ -597,8 +601,34 @@ string Description::Entry::generateSdpLines(string_view eol) const {
 	if (mDirection != Direction::Unknown)
 		sdp << "a=" << mDirection << eol;
 
-	for (const auto &attr : mAttributes)
+	for (const auto &attr : mAttributes) {
+		if (mRids.size() != 0 && match_prefix(attr, "ssrc:")) {
+			continue;
+		}
+
 		sdp << "a=" << attr << eol;
+	}
+
+	for (const auto &rid : mRids) {
+		sdp << "a=rid:" << rid << " send" << eol;
+	}
+
+	if (mRids.size() != 0) {
+		sdp << "a=simulcast:send ";
+
+		bool first = true;
+		for (const auto &rid : mRids) {
+			if (first) {
+				first = false;
+			} else {
+				sdp << ";";
+			}
+
+			sdp << rid;
+		}
+
+		sdp << eol;
+	}
 
 	return sdp.str();
 }
@@ -695,9 +725,12 @@ void Description::Media::addSSRC(uint32_t ssrc, optional<string> name, optional<
 		mAttributes.emplace_back("ssrc:" + std::to_string(ssrc));
 	}
 
-	if (msid)
+	if (msid) {
 		mAttributes.emplace_back("ssrc:" + std::to_string(ssrc) + " msid:" + *msid + " " +
 		                         trackId.value_or(*msid));
+		mAttributes.emplace_back("msid:" + *msid + " " +
+				trackId.value_or(*msid));
+	}
 
 	mSsrcs.emplace_back(ssrc);
 }
--- a/src/rtp.cpp
+++ b/src/rtp.cpp
@@ -130,12 +130,20 @@ void RtpExtensionHeader::setHeaderLength(uint16_t headerLength) {
 
 void RtpExtensionHeader::clearBody() { std::memset(getBody(), 0, getSize()); }
 
-void RtpExtensionHeader::writeCurrentVideoOrientation(size_t offset, uint8_t id, uint8_t value) {
-	if ((id == 0) || (id > 14) || ((offset + 2) > getSize()))
+void RtpExtensionHeader::writeOneByteHeader(size_t offset, uint8_t id, const byte *value, size_t size) {
+	if ((id == 0) || (id > 14) || (size == 0) || (size > 16) || ((offset + 1 + size) > getSize()))
 		return;
 	auto buf = getBody() + offset;
 	buf[0] = id << 4;
-	buf[1] = value;
+	if (size != 1) {
+		buf[0] |= (uint8_t(size) - 1);
+	}
+	std::memcpy(buf + 1, value, size);
+}
+
+void RtpExtensionHeader::writeCurrentVideoOrientation(size_t offset, const uint8_t id, uint8_t value) {
+	auto v = std::byte{value};
+	writeOneByteHeader(offset, id, &v, 1);
 }
 
 SSRC RtcpReportBlock::getSSRC() const { return ntohl(_ssrc); }
--- a/include/rtc/description.hpp
+++ b/include/rtc/description.hpp
@@ -91,6 +91,7 @@ public:
 		std::vector<string> attributes() const;
 		void addAttribute(string attr);
 		void removeAttribute(const string &attr);
+		void addRid(string rid);
 
 		struct RTC_CPP_EXPORT ExtMap {
 			static int parseId(string_view description);
@@ -119,6 +120,7 @@ public:
 
 	protected:
 		Entry(const string &mline, string mid, Direction dir = Direction::Unknown);
+
 		virtual string generateSdpLines(string_view eol) const;
 
 		std::vector<string> mAttributes;
@@ -128,6 +130,7 @@ public:
 		string mType;
 		string mDescription;
 		string mMid;
+		std::vector<string> mRids;
 		Direction mDirection;
 		bool mIsRemoved;
 	};
--- a/include/rtc/rtp.hpp
+++ b/include/rtc/rtp.hpp
@@ -39,6 +39,7 @@ struct RTC_CPP_EXPORT RtpExtensionHeader {
 
 	void clearBody();
 	void writeCurrentVideoOrientation(size_t offset, uint8_t id, uint8_t value);
+	void writeOneByteHeader(size_t offset, uint8_t id, const byte *value, size_t size);
 };
 
 struct RTC_CPP_EXPORT RtpHeader {


Sean-Der:
@paullouisageneau Ok everything is green!

* What additional tests should I add? I will add some simple unit tests, but do you have anything in particular?
* How would you like the example to flow
* Anything else?

paullouisageneau:
Thank you for the addition. It looks good overall, I'll look into it over the weekend.

Sean-Der:
Nice! If the C++ API looks ok will add the C equivalent 

Hopefully I kept the new API/designs in spirit with current. Don’t want to just bolt something ugly on the side :(

murillo128:
Not sure how RTX is handled in libdatachannel, but rtx packets should use the repaired-rtp-stream-id not the rtp-stream-id, is it possible to be handled with this API?

src/impl/peerconnection.cpp
```diff
@@ -736,7 +742,7 @@ shared_ptr<Track> PeerConnection::emplaceTrack(Description::Media description) {
 
 void PeerConnection::iterateTracks(std::function<void(shared_ptr<Track> track)> func) {
 	std::shared_lock lock(mTracksMutex); // read-only
-	for (auto it = mTrackLines.begin(); it != mTrackLines.end(); ++it) {
+	for (auto it = mTracks.begin(); it != mTracks.end(); ++it) {
```
paullouisageneau, line 745: What is the point of the change? This idea was that tracks are iterated according to m= line order.

src/description.cpp
```diff
@@ -307,10 +307,23 @@ string Description::generateSdp(string_view eol) const {
 
 	// Entries
 	bool first = true;
+
+	std::map<std::string, std::vector<shared_ptr<Entry>>> mLines;
```
paullouisageneau, line 311: It should be called `lines` as a local variable.

src/description.cpp
```diff
@@ -307,10 +307,23 @@ string Description::generateSdp(string_view eol) const {
...
+	for (const auto &mLine : mLines) {
+		std::vector<string> rids;
+		for (const auto &entry : mLine.second) {
+			if (entry->rid() != "") {
+				rids.push_back(entry->rid());
+			}
+		}
 
-		if (!entry->isRemoved() && std::exchange(first, false)) {
+		sdp << mLine.second.front()->generateSdp(eol, addr, port, rids);
```
paullouisageneau, line 324: Does this mean that only the first entry is serialized as an m= line? For some reason I missed that when looking at the API.

I don't think this is a good idea as description entries are supposed to represent m= lines, and it creates a convoluted implementation where you gather and reinject rids. The parsing methods would also need to be updated with the same logic. 

Maybe I'm missing something, but it seems that it would be way more straightforward to implement support for rids inside `Description::Media` and use the same track for the different quality levels.

paullouisageneau (commented):
Wouldn't it be easier to implement support for rids inside `Description::Media` and use the same track for the different quality levels?

paullouisageneau:
> Not sure how RTX is handled in libdatachannel, but rtx packets should use the repaired-rtp-stream-id not the rtp-stream-id, is it possible to be handled with this API?

RTX are generated by `RtcpNackResponder`. I think this is another argument to use the same track for the different quality levels: otherwise, it may not be possible to properly leverage per-track media handlers for simulcast.

src/impl/peerconnection.cpp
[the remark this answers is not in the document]
Sean-Der: Fixed! No longer needed since we don't do `rid` per Track.

src/impl/peerconnection.cpp
```diff
@@ -1230,9 +1227,13 @@ void PeerConnection::updateTrackSsrcCache(const Description &description) {
...
-			        if (auto it = mTracks.find(media->mid()); it != mTracks.end())
-				        if (auto track_for_mid = it->second.lock())
-					        track = track_for_mid;
+			        for (auto &t : mTracks) {
+				        track = t.lock();
+				        if (track == nullptr || track->mid() != media->mid()) {
+					        track = nullptr;
+					        continue;
+				        }
+			        }
```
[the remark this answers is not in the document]
Sean-Der: Fixed! No longer needed since we don't do `rid` per Track.

src/description.cpp
Sean-Der: @paullouisageneau done! Rewrote the PR to follow this design instead.

Sean-Der:
Hey @paullouisageneau can I get another review please, thank you!

src/rtp.cpp
```diff
@@ -138,6 +138,22 @@ void RtpExtensionHeader::writeCurrentVideoOrientation(size_t offset, uint8_t id,
 	buf[1] = value;
 }
 
+void RtpExtensionHeader::writeMid(size_t offset, uint8_t id, string mid) {
+	if ((id == 0) || (id > 14) || ((offset + 2) > getSize()))
```
paullouisageneau, line 142: Same here

src/rtp.cpp
```diff
@@ -138,6 +138,22 @@ void RtpExtensionHeader::writeCurrentVideoOrientation(size_t offset, uint8_t id,
 	buf[1] = value;
 }
 
+void RtpExtensionHeader::writeMid(size_t offset, uint8_t id, string mid) {
```
paullouisageneau, line 141: The `writeX` methods write the same kind of SDES headers and could be internally implemented with a common `writeSdesHeader(size_t offset, uint8_t id, const byte *value, size_t size)`.

examples/simulcast-sender/main.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+	std::string request =
+	    std::string("POST /api/whip HTTP/1.1\r\n") + "Host: whip.siobud.com\r\n" +
+	    "Authorization: Bearer seanTest\r\n" + "Content-Type: application/sdp\r\n" +
+	    "Content-Length: " + std::to_string(description.length()) + "\r\n\r\n" + description;
+	if (send(client_fd, request.c_str(), strlen(request.c_str()), 0) == -1) {
+		throw std::runtime_error("Failed to send offer to WHIP server");
+	}
+
+	if (read(client_fd, buffer, BUFFER_SIZE) == -1) {
```
paullouisageneau, line 86: I think there is no guarantee this works as `read()` returns as soon as 1 byte or more could be read. By chance the HTTP answer is typically received in a single TCP segment, but you should not rely on the behavior. It would be better practice to concatenate received data into a string until `read()` returns 0.

paullouisageneau (commented):
Thank you for the changes, this looks better to me, the implementation is way simpler!

examples/simulcast-sender/main.cpp
Sean-Der: @paullouisageneau I was thinking we probably shouldn't commit this example at all. I just wanted it as a demonstration for you.

https://b.siobud.com probably won't be up forever. Don't want to break your examples if I change anything!

src/rtp.cpp
Sean-Der: @paullouisageneau What do you think off adding a `writeOneByteHeader` API? I will not remove `writeCurrentVideoOrientation` so we don't break the public API.

I don't believe sharing code with SDES is worthwhile here. The format being the same is coincidental so it feels wrong to conflate them? As a reader I would find that confusing

Sean-Der:
Hi @paullouisageneau responded to all the reviews. Mind taking a look again, thank you!

src/description.cpp
```diff
@@ -597,8 +601,34 @@ string Description::Entry::generateSdpLines(string_view eol) const {
 	if (mDirection != Direction::Unknown)
 		sdp << "a=" << mDirection << eol;
 
-	for (const auto &attr : mAttributes)
+	for (const auto &attr : mAttributes) {
+		if (mRids.size() != 0 && attr.rfind("ssrc:", 0) == 0) {
```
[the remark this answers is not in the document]
paullouisageneau: Sorry, it's [match_prefix](https://github.com/paullouisageneau/libdatachannel/blob/ca8bc9018114e9848ecc76b45dc60d283386c357/src/description.cpp#L31), my bad. It's called in other places to match SDP lines so you should use it here too.

While it works here, I think reverse lookup is not a great way to match a prefix. It's less efficient and error prone in general as sometimes the substring may be present multiple times in the string.

src/rtp.cpp
paullouisageneau: It looks good, thanks! You could keep specialized methods `writeMid(size_t offset, uint8_t id, string mid)` and `writeRid(size_t offset, uint8_t id, string rid)` calling `writeOneByteHeader()` internally like `writeCurrentVideoOrientation()`.

src/rtp.cpp
```diff
@@ -130,12 +130,17 @@ void RtpExtensionHeader::setHeaderLength(uint16_t headerLength) {
...
 
-void RtpExtensionHeader::writeCurrentVideoOrientation(size_t offset, uint8_t id, uint8_t value) {
-	if ((id == 0) || (id > 14) || ((offset + 2) > getSize()))
+void RtpExtensionHeader::writeOneByteHeader(size_t offset, uint8_t id, const byte *value, size_t size) {
+	if ((id == 0) || (id > 14) || ((offset + 1 + size) > getSize()))
 		return;
 	auto buf = getBody() + offset;
-	buf[0] = id << 4;
-	buf[1] = value;
+	buf[0] = id << 4 | (uint8_t(size) - 1);
```
paullouisageneau, line 137: You need to check that `size > 0`.

examples/simulcast-sender/main.cpp
paullouisageneau: OK, makes sense!

src/rtp.cpp
Sean-Der: Fixed! I check that `size > 0` before writing length.

Also made sure to check size before copying.

src/description.cpp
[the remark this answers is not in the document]
Sean-Der: :+1 use match_prefix now

Sean-Der:
@paullouisageneau done! 

Pulled out the example also. PR is much smaller now :)

src/rtp.cpp
```diff
@@ -130,12 +130,20 @@ void RtpExtensionHeader::setHeaderLength(uint16_t headerLength) {
...
-	if ((id == 0) || (id > 14) || ((offset + 2) > getSize()))
+void RtpExtensionHeader::writeOneByteHeader(size_t offset, uint8_t id, const byte *value, size_t size) {
+	if ((id == 0) || (id > 14) || (size > 16) || ((offset + 1 + size) > getSize()))
 		return;
 	auto buf = getBody() + offset;
 	buf[0] = id << 4;
-	buf[1] = value;
+	if (size > 0) {
+		buf[0] |= (uint8_t(size) - 1);
+	}
```
paullouisageneau, line 140: You need to check `size > 0` at the beginning of the function, otherwise the size will implicitly be 1 byte while nothing is actually copied.

paullouisageneau, line 140: You need to check `size > 0` at the beginning of the method, otherwise the size will implicitly be 1 byte while nothing is actually copied.

src/rtp.cpp
paullouisageneau: Good point, keep it like this!

src/rtp.cpp
Sean-Der: @paullouisageneau fixed!

Sean-Der:
@paullouisageneau sorry for all the mistakes :/ 

Pushed mind reviewing again?

paullouisageneau (approved):
It's looking good, thank you!

pkviet:
@paullouisageneau could you create a tag for obs ? Unless you plan an imminent new release.
Thanks.
Btw we've bypassed the issue we were having with cross compiled DLLs by switching to a full msvc build.

paullouisageneau:
> @paullouisageneau could you create a tag for obs ? Unless you plan an imminent new release.
> Thanks.

@pkviet I'll create a tag when a solution to https://github.com/paullouisageneau/libdatachannel/issues/891#issuecomment-1603181274 is merged, so this can be used with OBS.

> Btw we've bypassed the issue we were having with cross compiled DLLs by switching to a full msvc build.

Great news!

pkviet:
@paullouisageneau hi, pinging again for a new tag for obs !

paullouisageneau:
@pkviet Sorry for the delay, it's now tagged as `v0.19.0-alpha.4`.

PR #914, opened 2023-06-22
Add RID + ExtMap support to capi

Sean-Der:
Add more optionals to rtcTrackInit

### DIFF
--- a/include/rtc/rtc.h
+++ b/include/rtc/rtc.h
@@ -253,14 +253,21 @@ RTC_C_EXPORT int rtcGetDataChannelReliability(int dc, rtcReliability *reliabilit
 // Track
 
 typedef struct {
+	int id;
+	char *uri;
+} rtcExtMapInit;
+
+typedef struct {
 	rtcDirection direction;
 	rtcCodec codec;
 	int payloadType;
 	uint32_t ssrc;
 	const char *mid;
-	const char *name;    // optional
-	const char *msid;    // optional
-	const char *trackId; // optional, track ID used in MSID
+	const char *name;             // optional
+	const char *msid;             // optional
+	const char *trackId;          // optional, track ID used in MSID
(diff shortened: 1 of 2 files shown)


src/capi.cpp
```diff
@@ -1075,6 +1075,18 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 		             init->msid ? std::make_optional(string(init->msid)) : nullopt,
 		             init->trackId ? std::make_optional(string(init->trackId)) : nullopt);
 
+		auto currentRid = init->rids;
+		while (*currentRid != NULL) {
```
paullouisageneau, line 1079: In other places (for instance `iceServers`), the C API uses an additional `*Count` field for arrays instead of a NULL terminator, so you should do the same here for consistency.

include/rtc/rtc.h
```diff
@@ -252,15 +252,22 @@ RTC_C_EXPORT int rtcGetDataChannelReliability(int dc, rtcReliability *reliabilit
...
 	int payloadType;
 	uint32_t ssrc;
 	const char *mid;
-	const char *name;    // optional
-	const char *msid;    // optional
-	const char *trackId; // optional, track ID used in MSID
+	const char *name;             // optional
+	const char *msid;             // optional
+	const char *trackId;          // optional, track ID used in MSID
+	const rtcExtMapInit **extMaps; // optional
```
paullouisageneau, line 269: If you add an `extMapsCount` field instead of relying on a NULL terminator, you could remove the double pointer and make this a simple array of `rtcExtMapInit`, which is simpler to initialize on user side.

Sean-Der:
@paullouisageneau Sorry for wasting your time on the review! With the OBS maintainers making C++ work I don't think we need this anymore :)

PR #915, opened 2023-06-24
Add type check on track to prevent PLI for non-video tracks.

mondain:
While looking at where to handle incoming PLI messages, I made the following minor changes:
- Fix typo in rtcpnackrsponder.hpp
- Remove irrelevant comment in rtcpreceivingsession.cpp
- Modify requestKeyframe method to prevent PLI on non-video tracks

### DIFF
--- a/include/rtc/rtcpnackresponder.hpp
+++ b/include/rtc/rtcpnackresponder.hpp
@@ -44,7 +44,7 @@ class RTC_CPP_EXPORT RtcpNackResponder final : public MediaHandlerElement {
 		/// Maximum storage size
 		const unsigned maximumSize;
 
-		/// Returnst current size
+		/// Returns current size
 		unsigned size();
 
 	public:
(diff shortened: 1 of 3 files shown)


mondain:
@paullouisageneau is there a method or callback for a streaming application to catch PLI or FIR coming from the end-point to which they are connected, so that the application can provide a proper response; such as restarting their video encoder?

paullouisageneau:
Thank you for the changes!

There is no callback to catch a PLI, because it would be port of an `RtcpSendingSession` which is not present in the library for now, meaning the sender applications has to handle RTCP messages.

paullouisageneau:
Thank you for the changes!

There is no callback to catch a PLI, because it would be part of a `RtcpSendingSession` which is not present in the library for now, meaning the sender application has to handle RTCP messages.

paullouisageneau (approved):

PR #920, opened 2023-06-27
Add PeerConnection::setConfiguration() method

murillo128:
This PR allows changing the peerconnection configuration before the ICE transport has been inited on sLD or sRD.

This is needed in order to support setting the STUN/TURN servers received on the POST answer of a WHIP call.

### DIFF
--- a/src/impl/icetransport.cpp
+++ b/src/impl/icetransport.cpp
@@ -117,27 +117,6 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
 		}
 	}
 
-	juice_turn_server_t turn_servers[MAX_TURN_SERVERS_COUNT];
-	std::memset(turn_servers, 0, sizeof(turn_servers));
-
-	// Add TURN servers
-	int k = 0;
-	for (auto &server : servers) {
-		if (!server.hostname.empty() && server.type == IceServer::Type::Turn) {
-			if (server.port == 0)
-				server.port = 3478; // TURN UDP port
-			PLOG_INFO << "Using TURN server \"" << server.hostname << ":" << server.port << "\"";
-			turn_servers[k].host = server.hostname.c_str();
-			turn_servers[k].username = server.username.c_str();
-			turn_servers[k].password = server.password.c_str();
-			turn_servers[k].port = server.port;
-			if (++k >= MAX_TURN_SERVERS_COUNT)
-				break;
-		}
-	}
-	jconfig.turn_servers = k > 0 ? turn_servers : nullptr;
-	jconfig.turn_servers_count = k;
-
 	// Bind address
 	if (config.bindAddress) {
 		jconfig.bind_address = config.bindAddress->c_str();
@@ -154,6 +133,44 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
 	mAgent = decltype(mAgent)(juice_create(&jconfig), juice_destroy);
 	if (!mAgent)
 		throw std::runtime_error("Failed to create the ICE agent");
+
+	// Filter STUN servers
+	servers.erase(std::remove_if(servers.begin(), servers.end(),
+	                             [](const IceServer &server) {
+		                             return server.hostname.empty() ||
+		                                    server.type != IceServer::Type::Turn;
+	                             }),
+	              servers.end());
+
+	// Add TURN servers
+	for (auto &server : servers) {
+		if (mTURNServersAdded++ >= MAX_TURN_SERVERS_COUNT) {
+			break;
+		}
+
+		addIceServer(server);
+	}
+}
+
+void IceTransport::addIceServer(IceServer server) {
+	if (server.hostname.empty() || server.type != IceServer::Type::Turn) {
+		PLOG_WARNING << "Only TURN servers are supported as additional ICE servers";
+		return;
+	}
+
+	if (server.port == 0)
+		server.port = 3478; // TURN UDP port
+
+	PLOG_INFO << "Using TURN server \"" << server.hostname << ":" << server.port << "\"";
+	juice_turn_server_t turn_server = {};
+	turn_server.host = server.hostname.c_str();
+	turn_server.username = server.username.c_str();
+	turn_server.password = server.password.c_str();
+	turn_server.port = server.port;
+
+	if (juice_add_turn_server(mAgent.get(), &turn_server) != 0) {
+		throw std::runtime_error("Failed to add TURN server");
+	}
 }
 
 IceTransport::~IceTransport() {
@@ -210,8 +227,17 @@ bool IceTransport::addRemoteCandidate(const Candidate &candidate) {
 	return juice_add_remote_candidate(mAgent.get(), string(candidate).c_str()) >= 0;
 }
 
-void IceTransport::gatherLocalCandidates(string mid) {
+void IceTransport::gatherLocalCandidates(string mid, std::vector<IceServer> additionalIceServers) {
 	mMid = std::move(mid);
+	std::shuffle(additionalIceServers.begin(), additionalIceServers.end(), utils::random_engine());
+
+	for (auto &server : additionalIceServers) {
+		if (mTURNServersAdded++ >= MAX_TURN_SERVERS_COUNT) {
+			break;
+		}
+
+		addIceServer(server);
+	}
 
 	// Change state now as candidates calls can be synchronous
 	changeGatheringState(GatheringState::InProgress);
@@ -533,59 +559,17 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
 			break;
 	}
 
+	// Filter STUN servers
+	servers.erase(std::remove_if(servers.begin(), servers.end(),
+	                             [](const IceServer &server) {
+		                             return server.hostname.empty() ||
+		                                    server.type != IceServer::Type::Turn;
+	                             }),
+	              servers.end());
+
 	// Add TURN servers
 	for (auto &server : servers) {
-		if (server.hostname.empty())
-			continue;
-		if (server.type != IceServer::Type::Turn)
-			continue;
-		if (server.port == 0)
-			server.port = server.relayType == IceServer::RelayType::TurnTls ? 5349 : 3478;
-
-		struct addrinfo hints = {};
-		hints.ai_family = AF_UNSPEC;
-		hints.ai_socktype =
-		    server.relayType == IceServer::RelayType::TurnUdp ? SOCK_DGRAM : SOCK_STREAM;
-		hints.ai_protocol =
-		    server.relayType == IceServer::RelayType::TurnUdp ? IPPROTO_UDP : IPPROTO_TCP;
-		hints.ai_flags = AI_ADDRCONFIG;
-		struct addrinfo *result = nullptr;
-		if (getaddrinfo(server.hostname.c_str(), std::to_string(server.port).c_str(), &hints,
-		                &result) != 0) {
-			PLOG_WARNING << "Unable to resolve TURN server address: " << server.hostname << ':'
-			             << server.port;
-			continue;
-		}
-
-		for (auto p = result; p; p = p->ai_next) {
-			if (p->ai_family == AF_INET || p->ai_family == AF_INET6) {
-				char nodebuffer[MAX_NUMERICNODE_LEN];
-				char servbuffer[MAX_NUMERICSERV_LEN];
-				if (getnameinfo(p->ai_addr, p->ai_addrlen, nodebuffer, MAX_NUMERICNODE_LEN,
-				                servbuffer, MAX_NUMERICSERV_LEN,
-				                NI_NUMERICHOST | NI_NUMERICSERV) == 0) {
-					PLOG_INFO << "Using TURN server \"" << server.hostname << ":" << server.port
-					          << "\"";
-					NiceRelayType niceRelayType;
-					switch (server.relayType) {
-					case IceServer::RelayType::TurnTcp:
-						niceRelayType = NICE_RELAY_TYPE_TURN_TCP;
-						break;
-					case IceServer::RelayType::TurnTls:
-						niceRelayType = NICE_RELAY_TYPE_TURN_TLS;
-						break;
-					default:
-						niceRelayType = NICE_RELAY_TYPE_TURN_UDP;
-						break;
-					}
-					nice_agent_set_relay_info(mNiceAgent.get(), mStreamId, 1, nodebuffer,
-					                          std::stoul(servbuffer), server.username.c_str(),
-					                          server.password.c_str(), niceRelayType);
-				}
-			}
-		}
-
-		freeaddrinfo(result);
+		addIceServer(server);
 	}
 
 	g_signal_connect(G_OBJECT(mNiceAgent.get()), "component-state-changed",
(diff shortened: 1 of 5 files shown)


paullouisageneau (commented):
Thank you for the PR, it looks good, basically my only comment is that I think `setConfiguration()` should accept the configuration by value and move it instead of accepting a reference.

src/impl/peerconnection.cpp
```diff
@@ -1247,4 +1247,11 @@ void PeerConnection::updateTrackSsrcCache(const Description &description) {
 		    description.media(i));
 }
 
+void PeerConnection::setConfiguration(const Configuration& config_) {
```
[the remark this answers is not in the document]
murillo128: I can merge the suggestion, but what benefits do you expect the move configuration? Passing by value is going to create a new copy of the structs, and as the struct does not store any resource that is costly to acquire/release (like a std::shared_ptr), the move will behave as another copy. It will probably be optimized out by the encoder, but at most it will be the same as passing by const ref and doing a copy later.

murillo128:
In order to fully support the WHIP stun/turn configuration I would need to add another change, but I am not sure exactly how to proceed.

In browsers, the ICE gathering phase is started when the SLD is called, so the flow is as follows:

- create PC with emtpy config
- create sdp offer
- send POST with sdp offer 
- receive POST response with ice server info and sdp anser
- update pc config with ice servers
- set local description
- set remote description

murillo128:
But in libdatachannel the createOffer and sLD are merged into a single method, so the flow is as follow:

- create PC with emtpy config
- create sdp offer and sLD
- send POST with sdp offer
- receive POST response with ice server info and sdp anser
- update pc config with ice servers
- set remote description

The sLD call creates and starts the ICE transport, so not sure what happens to the ice gathering if there have not been any server specified in the config and if we can just update the ice servers before setting the set remote description and trigger the ice gathering phase or if we need to split the ICE transport creation (to get the offer) and the ICE transport start (to start the ICE gathering phase) in two phases.

What do you think it is the best approach?

src/impl/peerconnection.cpp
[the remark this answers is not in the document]
paullouisageneau: As you say it's not a matter of performance here, just consistency with how configuration structs are handled in the codebase, and in general I like passing objects the function stores by value, and parameters the function only accesses by reference. Thank you for merging the suggestion.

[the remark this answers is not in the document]
murillo128: If you are fine, I am fine :)

paullouisageneau:
> The sLD call creates and starts the ICE transport, so not sure what happens to the ice gathering if there have not been any server specified in the config and if we can just update the ice servers before setting the set remote description and trigger the ice gathering phase or if we need to split the ICE transport creation (to get the offer) and the ICE transport start (to start the ICE gathering phase) in two phases.
> 
> What do you think it is the best approach?

@murillo128 Indeed libdatachannel features a simplified API where `createOffer()` and `createAnswer()` do not exist, and `setLocalDescription()` only supports being called with implicit description. It makes the implementation way simpler as you don't have to re-setup things according the set local description.

If I understand correctly, the WHIP flow in browsers prevents from transmitting ICE candidates from client to server, so you have to rely only on peer reflexive candidates on server-side. Is my understanding correct? In this case, it sounds like STUN servers won't make any difference as the server reflexive candidates will be ignored. Is this mechanism actually designed for TURN servers only?

Currently ICE gathering starts on `setLocalDescription()`, if no server is specified in the config then only host candidates are gathered. So I think what you want could be achieved with two retro-compatible modifications:
- Adding an option to prevent gathering on `setLocalDescription()`, for instance `disableAutoGathering`, in the configuration (on the model of `disableAutoNegotiation`), and a dedicated method `void gatherLocalCandidates(std::vector<IceServer> additionalIceServers = {})` to start the process manually, optionally with additional ICE servers.
- Changing the ICE transport to accept additional ICE servers in `impl::IceTransport::gatherLocalCandidates()`. For libnice, this is easy as I guess `nice_agent_set_relay_info()` may be called at any point before gathering, but for libjuice it requires adding a new API function to add TURN servers after agent creation.

Choosing this approach however makes this PR unnecessary.

murillo128:
>If I understand correctly, the WHIP flow in browsers prevents from transmitting ICE candidates from client to server, so you have to rely only on peer reflexive candidates on server-side. Is my understanding correct? In this case, it sounds like STUN servers won't make any difference as the server reflexive candidates will be ignored. Is this mechanism actually designed for TURN servers only?

That is not exactly how WHIP is intended to work. WHIP servers are meant to be either on the same private network as the WHIP client or with an reachable public ip address. 

So the WHIP server will return an SDP answer containing the ICE candidates required by the WHIP client to connect to. In case that the client is behind a firewall, it will not be able to connect to the server ICE candidates, and that's were TURN helps. 

When the WHIP client gathers the TURN reflexive candidates, it should send a STUN request to the ICE candidates, so the WHIP server will learn dynamically about the client new candidate without even needing to signal it via a subsequent PATCH request.

For completeness, the WHIP spec also allows the server to be behind a firewall (although I think this is an edge case), so that's when STUN makes sense. After exchanging the SDP O/A in the POST request, the WHIP client will send the gathered local candidates back to the WHIP server in an HTTP PATCH request. This way, the WHIP server can send back an STUN request to those candidates to open the NAT ports on its side.

murillo128:
Regarding this PR, I think it still makes sense as one of the strong points of libdatachannel is that is has a similar API than the w3c one. So I would more prone of hiding the details internally. What would you think about this:

- Do not start the ice gathering phase if the iceServers are empty when calling the sLD
- Allow updating the iceServer in the pc.setConfiguration if the current iceServers are empty
- Start the ICE gathering automatically (if not already started) on the sRD

The behavior and apis would be quite similar to chrome/webrtc one (with the nits of missing createOffer/createAnswer)

murillo128:
I have added an initial version of how updating the servers on `setConfiguration` would look like. Should work for libnice.

paullouisageneau:
> Regarding this PR, I think it still makes sense as one of the strong points of libdatachannel is that is has a similar API than the w3c one. So I would more prone of hiding the details internally. What would you think about this:
> 
>     * Do not start the ice gathering phase if the iceServers are empty when calling the sLD
> 
>     * Allow updating the iceServer in the pc.setConfiguration if the current iceServers are empty
> 
>     * Start the ICE gathering automatically (if not already started) on the sRD
> 
> 
> The behavior and apis would be quite similar to chrome/webrtc one (with the nits of missing createOffer/createAnswer)

I'm not really comfortable with fitting this in by changing the current behavior of sLD/sRD as I would break other people's use cases. For instance, users expect sLD to always start gathering and may rely on the gathering complete event to get the non-trickled description, and it is a perfectly normal use case to set no ICE servers when you know the peer is reachable. I'd also like to maintain a consistent behavior with [datachannel-wasn](https://github.com/paullouisageneau/datachannel-wasm), which calls the browser implementation under the hood.

Updating the ice servers could indeed be done with `setConfiguration()`, however there are a couple drawbacks: the rest of the configuration is ignored, and allowing additional servers would require to dedupe the list.

In order to stick to an API closer to the W3C without requiring architectural changes, I think it could be a good compromise to add a `createOffer()` method which initializes the ICE transport if not initialized already, generates and returns the local offer description, but does not set it and does not start gathering. `setLocalDescription()` would still work as usual with implicit description and start gathering, and could be called after changing the ice servers. What do you think?

murillo128:
I am fine for adding an flag for controlling if sLD triggerst the ice gathering or not, adding a new createOffer woiuld be ideal, although I would need more support from your side in order to understand the required changes.

Regarding changing the rest of parameters in `setConfiguration()`, in the webrtc w3c api can be done, but it will trigger an ICE restart if some configuration parameters are changed. 

What we can do is to check if the configuration is actually changing any other parameter and raise an exception if so. 
The usage of `setConfiguration()` would work as follows for updating the ice servers:

```
auto config = pc.getConfiguration().
config.iceServers.push(....);
pc.setConfiguration(config);
```

regarding turn dedup, are you performing any at the moment? what prevents a client inserting the same turn server twice?

paullouisageneau:
> I am fine for adding an flag for controlling if sLD triggerst the ice gathering or not, adding a new createOffer woiuld be ideal, although I would need more support from your side in order to understand the required changes.

If I'm not mistaken, it would boil down to:
```
namespace rtc {
string PeerConnection::createOffer() {
	auto iceTransport = impl()->initIceTransport();
	if (!iceTransport)
		throw std::runtime_error("Peer connection is closed");

	Description offer = iceTransport->getLocalDescription(type);
	return impl()->populateLocalDescription(std::move(offer));
}
}
```
where `impl::PeerConnection::populateLocalDescription()` would do the operations in the first part of `impl::PeerConnection::processLocalDescription()` until `description.setFingerprint()`.

> What we can do is to check if the configuration is actually changing any other parameter and raise an exception if so. The usage of `setConfiguration()` would work as follows for updating the ice servers:
> 
> ```
> auto config = pc.getConfiguration().
> config.iceServers.push(....);
> pc.setConfiguration(config);
> ```

This works for me!

> regarding turn dedup, are you performing any at the moment? what prevents a client inserting the same turn server twice?

Nothing prevents it, it should deduped in libjuice in the end (because otherwise the TURN server would get confused about the same client authentifying twice). I'm not sure about libnice though.

src/impl/icetransport.cpp
```diff
@@ -210,9 +215,17 @@ bool IceTransport::addRemoteCandidate(const Candidate &candidate) {
...
+void IceTransport::gatherLocalCandidates(string mid, std::vector<IceServer> additionalIceServers) {
 	mMid = std::move(mid);
 
+	int k = 0;
+	for (auto &server : additionalIceServers) {
+		addTURNServer(server);
+		if (++k >= MAX_TURN_SERVERS_COUNT) {
+			break;
+		}
+	}
```
paullouisageneau, line 227: This mechanism will silently ignore any STUN server added in `additionalIceServers`. Is it by design?

src/peerconnection.cpp
```diff
@@ -146,11 +146,21 @@ void PeerConnection::setLocalDescription(Description::Type type) {
...
 	signalingLock.unlock();
 
-	if (impl()->gatheringState == GatheringState::New) {
+	if (impl()->gatheringState == GatheringState::New && !impl()->config.disableAutoGathering) {
 		iceTransport->gatherLocalCandidates(impl()->localBundleMid());
 	}
 }
 
+void PeerConnection::gatherLocalCandidates(std::vector<IceServer> additionalIceServers) {
+	auto iceTransport = impl()->initIceTransport();
```
paullouisageneau, line 155: Since the ICE transport is created here, why not simply add the additional ICE servers to the config just before? Then you wouldn't need any custom logic in `IceTransport`.

.gitmodules
```diff
@@ -8,7 +8,7 @@
 	shallow = true
 [submodule "deps/libjuice"]
 	path = deps/libjuice
-	url = https://github.com/paullouisageneau/libjuice.git
+	url = https://github.com/sean-der/libjuice.git
```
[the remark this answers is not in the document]
paullouisageneau: I've merged https://github.com/paullouisageneau/libjuice/pull/239 and released it in v1.4.0, but I'm not sure you actually need it for this PR, as it seems to me you could simply update the config before `initIceTransport()` is called.

src/peerconnection.cpp
paullouisageneau: Sorry, I was mislead by the call here, this should be `getIceTransport()` and you should throw a `logic_error` if null, as it means no local description has been set.

.gitmodules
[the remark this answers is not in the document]
Sean-Der: Done!

src/impl/icetransport.cpp
Sean-Der: I will go fix that. I was only thinking about TURN servers when I implemented, will implement and test

Sean-Der: @paullouisageneau I will need to add another API to libjuice for that.

I haven't had anyone request this/I don't need it myself. Would it be better for libdatachannel if I implemented it though?

Personally it feels like more LoC/more bugs for a feature that I don't need yet! I am happy to do it though, your call!

paullouisageneau: No need to implement it for now, but rather than silently ignoring the issue, you could rename the method `addIceServer()` and display a warning like "only TURN servers are supported as additional ICE servers".

src/impl/icetransport.cpp
```diff
@@ -210,9 +215,17 @@ bool IceTransport::addRemoteCandidate(const Candidate &candidate) {
 	return juice_add_remote_candidate(mAgent.get(), string(candidate).c_str()) >= 0;
 }
 
-void IceTransport::gatherLocalCandidates(string mid) {
+void IceTransport::gatherLocalCandidates(string mid, std::vector<IceServer> additionalIceServers) {
 	mMid = std::move(mid);
 
+	int k = 0;
+	for (auto &server : additionalIceServers) {
```
paullouisageneau, line 222: You should randomize `additionalIceServers` before adding, like it is done in the constructor.

src/impl/icetransport.cpp
```diff
@@ -210,9 +215,17 @@ bool IceTransport::addRemoteCandidate(const Candidate &candidate) {
...
+		addTURNServer(server);
+		if (++k >= MAX_TURN_SERVERS_COUNT) {
```
paullouisageneau, line 224: The counter should now be a class member otherwise it is not enforced properly anymore.

src/peerconnection.cpp
```diff
@@ -146,11 +146,22 @@ void PeerConnection::setLocalDescription(Description::Type type) {
...
 
+void PeerConnection::gatherLocalCandidates(std::vector<IceServer> additionalIceServers) {
+	auto iceTransport = impl()->getIceTransport();
+	if (!iceTransport) {
+		throw std::logic_error("No IceTransport. Local Description has not been set");
+	}
+
+	if (impl()->gatheringState == GatheringState::New) {
+		iceTransport->gatherLocalCandidates(impl()->localBundleMid(), additionalIceServers);
+	}
```
paullouisageneau, line 162: If the state is not `New` you could log a warning like "Candidates gathering already started".

src/impl/icetransport.cpp
```diff
@@ -154,6 +133,32 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
 	mAgent = decltype(mAgent)(juice_create(&jconfig), juice_destroy);
 	if (!mAgent)
 		throw std::runtime_error("Failed to create the ICE agent");
+
+	// Add TURN servers
+	int k = 0;
+	for (auto &server : servers) {
+		addTURNServer(server);
```
paullouisageneau, line 140: You should filter out STUN servers explicitly.

.gitmodules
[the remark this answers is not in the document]
paullouisageneau: Thanks!

src/impl/icetransport.cpp
Sean-Der: Done, we now log `Only TURN servers are supported as additional ICE servers`

src/impl/icetransport.cpp
Sean-Der: Done! New method returns early on STUN servers

src/peerconnection.cpp
Sean-Der: Done! Logs a message

src/impl/icetransport.cpp
Sean-Der: It is now a class member

src/impl/icetransport.cpp
Sean-Der: Done I do a shuffle

Sean-Der:
@paullouisageneau can I get another review please, thank you!

src/impl/icetransport.cpp
```diff
@@ -154,6 +133,39 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
 	mAgent = decltype(mAgent)(juice_create(&jconfig), juice_destroy);
 	if (!mAgent)
 		throw std::runtime_error("Failed to create the ICE agent");
+
+	// Add TURN servers
+	for (auto &server : servers) {
+		if (mTURNServersAdded++ >= MAX_TURN_SERVERS_COUNT) {
+			break;
+		}
```
paullouisageneau, line 141: The counter incorrectly counts STUN servers too, so you might end up with a single TURN server or none added if you set STUN servers. It should be done once in `addIceServer()` after ensuring the type is indeed TURN.

src/impl/icetransport.cpp
```diff
@@ -154,6 +133,39 @@ IceTransport::IceTransport(const Configuration &config, candidate_callback candi
...
+
+		addIceServer(server, false);
+	}
+}
+
+void IceTransport::addIceServer(IceServer server, bool warnOnSTUN) {
```
paullouisageneau, line 147: I'm not a fan of the boolean. Why not always logging the warning and filtering out STUN servers in the constructor for loop?

Sean-Der: Done! Not needed anymore because we filter.

src/impl/icetransport.cpp
Sean-Der: Done! Not needed anymore because we filter.

Sean-Der:
Ok last one I hope @paullouisageneau I think I got it this time :)

paullouisageneau (approved):
Thank you for the work, it will be great to introduce TURN support for WHIP/WHEP!

Sean-Der:
@paullouisageneau whats left for a new tag? I can take on the tasks!

paullouisageneau:
@Sean-Der I just released `v0.21.0`!

PR #924, opened 2023-07-11
Update Opus to have proper media format

Sean-Der:
Defined as `opus` in [rfc7587#section-6.1](https://datatracker.ietf.org/doc/html/rfc7587#section-6.1)

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -1127,7 +1127,7 @@ void Description::Audio::addAudioCodec(int payloadType, string codec, optional<s
 }
 
 void Description::Audio::addOpusCodec(int payloadType, optional<string> profile) {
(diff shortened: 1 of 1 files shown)


Sean-Der:
@paullouisageneau this caused negotiation to fail with AWS' Chime (just found via OBS testing with different services)

murillo128:
I don't mind doing the change, but being pedantic, encoding names are case insensitive: https://datatracker.ietf.org/doc/html/rfc4855#section-3

paullouisageneau:
Indeed, it should be case insensitive, but if the change increases compatibility it still looks good to me!

paullouisageneau (approved):

paullouisageneau:
@Sean-Der I tagged it as `v0.19.0-alpha.5` in case you need it.

PR #950, opened 2023-08-23
Add AAC support

troman123:
according to  rfc6416,  libdatachannel add aac rtp support
https://datatracker.ietf.org/doc/rfc6416/

### DIFF
--- a/src/capi.cpp
+++ b/src/capi.cpp
@@ -1026,6 +1026,7 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 			case RTC_CODEC_OPUS:
 			case RTC_CODEC_PCMU:
 			case RTC_CODEC_PCMA:
+			case RTC_CODEC_AAC:
 				mid = "audio";
 				break;
 			default:
@@ -1059,7 +1060,8 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 		}
 		case RTC_CODEC_OPUS:
 		case RTC_CODEC_PCMU:
-		case RTC_CODEC_PCMA: {
+		case RTC_CODEC_PCMA:
+		case RTC_CODEC_AAC: {
 			auto desc = Description::Audio(mid, direction);
 			switch (init->codec) {
 			case RTC_CODEC_OPUS:
@@ -1071,6 +1073,9 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 			case RTC_CODEC_PCMA:
 				desc.addPCMACodec(init->payloadType);
 				break;
+			case RTC_CODEC_AAC:
+					desc.addAacCodec(init->payloadType, init->profile ? std::make_optional(string(init->profile)) : nullopt);
+					break;
 			default:
 				break;
 			}
@@ -1225,6 +1230,23 @@ int rtcSetOpusPacketizationHandler(int tr, const rtcPacketizationHandlerInit *in
 		track->setMediaHandler(opusHandler);
 		return RTC_ERR_SUCCESS;
 	});
+}
+
+int rtcSetAACPacketizationHandler(int tr, const rtcPacketizationHandlerInit *init) {
+	return wrap([&] {
+		auto track = getTrack(tr);
+		// create RTP configuration
+		auto rtpConfig = createRtpPacketizationConfig(init);
+		// create packetizer
+		auto packetizer = std::make_shared<AACRtpPacketizer>(rtpConfig);
+		// create AAC handler
+		auto aacHandler = std::make_shared<AACPacketizationHandler>(packetizer);
+		emplaceMediaChainableHandler(aacHandler, tr);
+		emplaceRtpConfig(rtpConfig, tr);
+		// set handler
+		track->setMediaHandler(aacHandler);
+		return RTC_ERR_SUCCESS;
+  });
 }
 
 int rtcChainRtcpSrReporter(int tr) {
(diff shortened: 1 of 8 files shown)


troman123:
@paullouisageneau could you review this and provide some feedback of any change that may be required for merging?

src/capi.cpp
```diff
@@ -1098,6 +1098,74 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 	});
 }
 
+RTC_C_EXPORT int rtcAddAACTrack(int pc, const rtcTrackInit *init, uint8_t *latmMuxCfg, int size) {
```
paullouisageneau, line 1101: I would prefer that you add fields to `rtcTrackInit` rather than introducing a new function with duplicated code, just add a comment on them to specify that they are for AAC only.

src/capi.cpp
```diff
@@ -1227,6 +1295,23 @@ int rtcSetOpusPacketizationHandler(int tr, const rtcPacketizationHandlerInit *in
...
 }
 
+int rtcSetAACPacketizationHandler(int tr, const rtcPacketizationHandlerInit *init) {
+	return wrap([&] {
+		auto track = getTrack(tr);
+		// create RTP configuration
+		auto rtpConfig = createRtpPacketizationConfig(init);
+		// create packetizer
+		auto packetizer = std::make_shared<AACRtpPacketizer>(rtpConfig);
+		// create Opus handler
```
paullouisageneau, line 1305:
```suggestion
		// create AAC handler
```

src/capi.cpp
```diff
@@ -1098,6 +1098,74 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
...
+				mid = "audio";
+				break;
+			}
+		}
+
+		optional<Description::Media> optDescription = nullopt;
+		switch (init->codec) {
+		case RTC_CODEC_AAC: {
+			auto desc = Description::Audio(mid, direction);
+			switch (init->codec) {
```
paullouisageneau, line 1128: I don't get why there are two switch blocks here, is it a mistake?

src/capi.cpp
```diff
@@ -1098,6 +1098,74 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
...
+			case RTC_CODEC_AAC:
+					if (latmMuxCfg) {
+						std::string muxbuf="cpresent=0;config=";
+						for (int i = 0; i < size; i++) {
+							char buf[3] = {0};
+							snprintf(buf, 3, "%02x", latmMuxCfg[i]);
```
paullouisageneau, line 1134: This could be done more simply and idiomatically with an `std::ostringstream` rather than relying on a C-style `snprintf`.

src/capi.cpp
```diff
@@ -1098,6 +1098,74 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
...
+							muxbuf.append(buf, 2);
+						}
+						
+						desc.addAudioCodec(init->payloadType, "MP4A-LATM", muxbuf);
```
paullouisageneau, line 1138: Could you please create a dedicated method `Description::Media::addAacCodec()`? The logic to serialize `latmMuxCfg` could also be moved there.

paullouisageneau (commented):
Thank you for adding! If possible, I'd like to keep the logic in the C++ classes so the C API stays a simple wrapper.

src/capi.cpp
troman123: add a latmMuxCfg field for aac latm track.

src/capi.cpp
troman123: done

src/capi.cpp
troman123: done , delete rtcAddTrackEx

src/capi.cpp
troman123: done, add  Description::Media::addAacCodec method

src/capi.cpp
troman123: done

troman123:
All done! Can I get another review please @paullouisageneau

thanks

src/capi.cpp
```diff
@@ -1071,6 +1073,9 @@ int rtcAddTrackEx(int pc, const rtcTrackInit *init) {
 			case RTC_CODEC_PCMA:
 				desc.addPCMACodec(init->payloadType);
 				break;
+			case RTC_CODEC_AAC:
+					desc.addAacCodec(init->payloadType, init->latmMuxCfg);
```
paullouisageneau, line 1077: This will attempt to create a string from a null pointer if `init->latmMuxCfg` is not set.
```suggestion
					desc.addAacCodec(init->payloadType, init->latmMuxCfg ? std::make_optional(string(init->latmMuxCfg)) : nullopt);
```

src/description.cpp
```diff
@@ -1138,6 +1138,17 @@ void Description::Audio::addPCMUCodec(int payloadType, optional<string> profile)
 	addAudioCodec(payloadType, "PCMU", profile);
 }
 
+void Description::Audio::addAacCodec(int payloadType, optional<string> latmMuxCfg) {
+	if (latmMuxCfg) {
+		string profile = "cpresent=0;config=";
+		profile += *latmMuxCfg;
```
paullouisageneau, line 1144: If `latmMuxCfg` is just a string concatenated to the profile now, do you really need a different argument for AAC? Couldn't you pass `profile` like other codecs, and set it to `cpresent=1` by default if unset?

src/capi.cpp
troman123: Fixed!

troman123:
Thanks again for the review @paullouisageneau

src/description.cpp
troman123: set a common profile field in rtcTrackInit

paullouisageneau (approved):
Thanks for the changes!

PR #955, opened 2023-08-26
Use plog::UTF8Converter instead of std::codecvt_utf8

SergiusTheBest:
Plog has a built-in UTF-8 conversion, so `std::codecvt_utf8` is not required. This makes code simpler and gets rid of `std::codecvt_utf8` deprecation warnings. Also it fixes using a more recent plog version (>= 1.1.10) that supports utf8everywhere mode. That mode is triggered by the `/utf-8` switch in MSVC and is used by default in vcpkg.


paullouisageneau (approved):
That's very handy, thank you for making the change. And thank you for the plog library!

PR #978, opened 2023-09-13
Set the shared object version (SOVERSION) to only the major.minor version

Conan-Kudo:
This declares that the library ABI is intended to be backward compatible to older versions of the same major.minor version.


Conan-Kudo:
@paullouisageneau To explain this more concretely: if your intent is that you don't intend to regularly break the public API of this library, then this change is fine.

The Application Binary Interface (ABI) is constructed out of the compiler symbol generation based on the function signatures that exist for all publicly accessible classes and headers that define "plain" functions that aren't wrapped in a class.

So ultimately, if you don't intend to regularly make changes to the API where programs have to change how they use your library, this should be okay to land.

paullouisageneau:
@Conan-Kudo Thank you for the PR! I

 have a concern as libdatachannel is still in 0.X and ABI compatibility is not enforced between minor version numbers since the API can change (it is between build version numbers). Starting from 1.X, which hopefully will be released in a few months, I plan to enforce ABI compatibility between major versions. Isn't this an issue? Shouldn't shared object version use major and minor instead for now?

paullouisageneau:
@Conan-Kudo Thank you for the PR! 

I have a concern as libdatachannel is still in 0.X and ABI compatibility is not enforced between minor version numbers since the API can change (it is between build version numbers). Starting from 1.X, which hopefully will be released in a few months, I plan to enforce ABI compatibility between major versions. Isn't this an issue? Shouldn't shared object version use major and minor instead for now?

Conan-Kudo:
> Shouldn't shared object version use major and minor instead for now?

Yes, you're quite right! I've updated it accordingly.

paullouisageneau (approved):

PR #999, opened 2023-10-05
Add a media handler to respond to intra frame requests

kuzux:
This PR introduces a PLIResponder class, which can be used for responding back with an intra frame after the receiver requests one with an FIR (Full Intra Request) or a PLI (Picture Loss Indicator) message.

This is useful when a video is being streamed live and we can control the behavior of the video encoder. PLIResponder simply notifies its caller when a new intra frame is requested, which passes this request onto its video frame source.

I actually thought about contributing the generic PLIResponder thing we used upstream all the way back last year, but the implementation back then had some issues (later discovered that they were due to browsers grouping control messages)

Now, there are a couple things missing:
* The naming is temporary, since this responds to FIR messages as well as PLI
* I'm not sure about how to integrate this into the C API
* Also I'm not sure about how to integrate this into an example (This doesn't make sense in the existing examples since none of them can meaningfully respond to an intra frame request)

### DIFF
--- a/include/rtc/plihandler.hpp
+++ b/include/rtc/plihandler.hpp
@@ -20,12 +20,12 @@ namespace rtc {
 /// Responds to PLI and FIR messages sent by the receiver. The sender should respond to these
 /// messages by sending an intra. 
 class RTC_CPP_EXPORT PliHandler final : public MediaHandlerElement {
-    std::function<void(void)> mOnPli;
+    rtc::synchronized_callback<> mOnPli;
 
 public:
 	/// Constructs the PLIResponder object to notify whenever a new intra frame is requested
 	/// @param onPli The callback that gets called whenever an intra frame is requested by the receiver
-    PliHandler(rtc::synchronized_callback<> onPli);
+    PliHandler(std::function<void(void)> onPli);
     ChainedIncomingControlProduct processIncomingControlMessage(message_ptr) override;
 };
 
--- a/include/rtc/plihandler.hpp
+++ b/include/rtc/plihandler.hpp
@@ -18,13 +18,13 @@ namespace rtc {
 
 /// Responds to PLI and FIR messages sent by the receiver. The sender should respond to these
 /// messages by sending an intra. 
-class RTC_CPP_EXPORT PliResponder final : public MediaHandlerElement {
+class RTC_CPP_EXPORT PliHandler final : public MediaHandlerElement {
     std::function<void(void)> onPli;
 
 public:
 	/// Constructs the PLIResponder object to notify whenever a new intra frame is requested
 	/// @param onPli The callback that gets called whenever an intra frame is requested by the receiver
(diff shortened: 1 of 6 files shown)


kuzux:
Also noticed that #929 will definitely change things in relation to this :D

include/rtc/pliresponder.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "mediahandlerelement.hpp"
+#include <functional>
+
+namespace rtc {
+
+/// Responds to PLI and FIR messages sent by the receiver. The sender should respond to these
+/// messages by sending an intra. 
+class RTC_CPP_EXPORT PliResponder final : public MediaHandlerElement {
+    std::function<void(void)> onPli;
```
paullouisageneau, line 22: `onPli` should be a `Synchronized_callback` for thread safety. Also, private member variables should be prefixed by `m` in the library so the name should be `mOnPli`.

include/rtc/pliresponder.hpp
```diff
@@ -0,0 +1,35 @@
...
+#if RTC_ENABLE_MEDIA
+
+#include "mediahandlerelement.hpp"
+#include <functional>
+
+namespace rtc {
+
+/// Responds to PLI and FIR messages sent by the receiver. The sender should respond to these
+/// messages by sending an intra. 
+class RTC_CPP_EXPORT PliResponder final : public MediaHandlerElement {
```
paullouisageneau, line 21: I'm not a fan of the naming, `Pli` is fine by me but this media handler does not really respond to anything (meaning it doesn't send messages). You could call it something more generic like `PliHandler`.

include/rtc/plihandler.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include "mediahandlerelement.hpp"
+#include "utils.hpp"
+#include <functional>
+
+namespace rtc {
+
+/// Responds to PLI and FIR messages sent by the receiver. The sender should respond to these
+/// messages by sending an intra. 
+class RTC_CPP_EXPORT PliHandler final : public MediaHandlerElement {
+    std::function<void(void)> mOnPli;
```
paullouisageneau, line 23: This should be changed to `synchronized_callback` as it's the actual callback being called by this class.

src/plihandler.cpp
```diff
@@ -0,0 +1,44 @@
...
+				mOnPli();
+				break;
+			}
+		}
+		offset += header->lengthInBytes();
+	}
+	return { message, std::nullopt };
+}
+
+PliHandler::PliHandler(rtc::synchronized_callback<> onPli) : mOnPli(onPli) { }
```
paullouisageneau, line 40: This should stay a `function`.

paullouisageneau (approved):
Looking good, thanks.

PR #1009, opened 2023-10-21
Allow setting CA certificate during TLS connection

### LINKED ISSUE #1007: Unable to Connect to TLS when Using MbedTLS
When using MbedTLS, I cannot establish a TLS connection to servers that use CA certificates issued by Let's Encrypt. I suspect this is because MbedTLS does not load the root CA.

As a result, I'm seeking a feature that allows loading a specific CA certificate.

In OpenSSL, the paths retrieved via `X509_get_default_cert_file()` or `X509_get_default_cert_dir()` serve as the default root CA, enabling the avoidance of this issue.

In MbedTLS, by adding the following code to `src/impl/tlstransport.cpp`, I confirmed that TLS connections work on Ubuntu:

```cpp
mbedtls_x509_crt_init(&mCacert); // Defined as mbedtls_x509_crt mCacert; in tlstransport.h
mbedtls::check(mbedtls_x509_crt_parse_file(&mCacert, "/etc/ssl/certs/ca-certificates.crt"));
mbedtls_ssl_conf_ca_chain(&mConf, &mCacert, NULL);
```

I'm considering submitting a Pull Request to address this issue. In this context, I'd like to know whether the CA certificate should be passed as a file path or in PEM string format. Furthermore, I'm curious if there's a need to add the capability to load specific CA certificates for OpenSSL and GnuTLS as well.

melpon:
Fixes #1007.

### DIFF
--- a/include/rtc/websocket.hpp
+++ b/include/rtc/websocket.hpp
@@ -39,6 +39,7 @@ public:
 		optional<std::chrono::milliseconds> connectionTimeout; // zero to disable
 		optional<std::chrono::milliseconds> pingInterval;      // zero to disable
 		optional<int> maxOutstandingPings;
+		optional<string> caCertificatePemFile;
 	};
 
 	WebSocket();
--- a/src/impl/verifiedtlstransport.cpp
+++ b/src/impl/verifiedtlstransport.cpp
@@ -13,9 +13,11 @@
 
 namespace rtc::impl {
 
+static const string PemBeginCertificateTag = "-----BEGIN CERTIFICATE-----";
+
 VerifiedTlsTransport::VerifiedTlsTransport(
     variant<shared_ptr<TcpTransport>, shared_ptr<HttpProxyTransport>> lower, string host,
-    certificate_ptr certificate, state_callback callback)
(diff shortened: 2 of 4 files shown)


paullouisageneau (commented):
Thank you for the PR. It works, but architecturally, setting the CA chain in `TlsTransport` is quite weird, since `TlsTransport` by itself doesn't perform certificate verification. I think the CA chain should only be set in `VerifiedTlsTransport`, which would also be simpler as you wouldn't have to forward `cacert` all the way down to `TlsTransport`. What do you think?

Moreover, you need to free `mCaCert` in the destructor and constructor catch block. I'm just noticing now that the `TlsTransport` destructor is not implemented for MbedTLS... I'm pushing a fix. @Sean-Der if you use Websockets with MbedTLS somewhere, there must be a memory leak.

melpon:
> I think the CA chain should only be set in VerifiedTlsTransport

Indeed, you're right. I'll make the changes.

melpon:
I've made the changes and squashed the commits. Thank you for the advice!

paullouisageneau (approved):
Looks good, thanks.

PR #1013, opened 2023-10-23
Implement missing methods of RtcpReportBlock

ognkrmms:
This PR implements `setPacketsLost`, `getLossPercentage` and `getPacketLostCount` methods of `RtcpReportBlock` that we need for receiver reports.

I also added `extendedHighestSeqNo `method to return the extended sequence number of RtcpReportBlock in case one wants to access the the whole thing.

### DIFF
--- a/src/rtp.cpp
+++ b/src/rtp.cpp
@@ -169,18 +169,18 @@ void RtcpReportBlock::preparePacket(SSRC in_ssrc, [[maybe_unused]] unsigned int 
 
 void RtcpReportBlock::setSSRC(SSRC in_ssrc) { _ssrc = htonl(in_ssrc); }
 
-void RtcpReportBlock::setPacketsLost(uint8_t packetsLost,
-                                     unsigned int totalPackets) {
-	_fractionLostAndPacketsLost = ( (uint32_t)packetsLost << 24) && htonl(totalPackets);
+void RtcpReportBlock::setPacketsLost(uint8_t fractionLost,
+                                     unsigned int packetsLostCount) {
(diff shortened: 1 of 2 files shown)


src/rtp.cpp
```diff
@@ -169,26 +169,27 @@ void RtcpReportBlock::preparePacket(SSRC in_ssrc, [[maybe_unused]] unsigned int
...
+	_fractionLostAndPacketsLost = ( (uint32_t)packetsLost << 24) && htonl(totalPackets);
 }
 
-unsigned int RtcpReportBlock::getLossPercentage() const {
-	// TODO Implement loss percentages.
-	return 0;
+uint8_t RtcpReportBlock::getLossPercentage() const {
+	// Loss percentage is expressed as 8-bit fixed point number
+	// In order to get actual percentage divide the result by 256
+	return _fractionLostAndPacketsLost & 0xFF;
```
paullouisageneau, line 180: As the function just returns the fraction lost as fixed point, I think you should rename it `getFractionLost` otherwise it it can be quite confusing.

src/rtp.cpp
```diff
@@ -169,26 +169,27 @@ void RtcpReportBlock::preparePacket(SSRC in_ssrc, [[maybe_unused]] unsigned int
...
 }
 
 unsigned int RtcpReportBlock::getPacketLostCount() const {
```
paullouisageneau, line 183: Could you also rename this one in `getPacketsLostCount` for consistency?

paullouisageneau (commented):
Thank you for the PR. I think it's a good opportunity to change some of the naming.

paullouisageneau (approved):
Looks good, thanks!

PR #1015, opened 2023-10-24
Set SCTP zero checksum

melpon:
Based on the discussion at https://github.com/paullouisageneau/libdatachannel/discussions/1012, where it was concluded that enabling `SCTP_ACCEPT_ZERO_CHECKSUM` is not an issue, I have implemented zero checksum.

### DIFF
--- a/src/impl/sctptransport.cpp
+++ b/src/impl/sctptransport.cpp
@@ -102,7 +102,6 @@ SctpTransport::InstancesSet *SctpTransport::Instances = new InstancesSet;
 
 void SctpTransport::Init() {
 	usrsctp_init(0, SctpTransport::WriteCallback, SctpTransport::DebugCallback);
-	usrsctp_enable_crc32c_offload();       // We'll compute CRC32 only for outgoing packets
 	usrsctp_sysctl_set_sctp_pr_enable(1);  // Enable Partial Reliability Extension (RFC 3758)
 	usrsctp_sysctl_set_sctp_ecn_enable(0); // Disable Explicit Congestion Notification
 #ifdef SCTP_DEBUG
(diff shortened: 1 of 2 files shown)


src/impl/sctptransport.cpp
```diff
@@ -284,6 +284,16 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
 		throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
 		                         std::to_string(errno));
 
+	// When using SCTP over DTLS, the data integrity is ensured by DTLS, so there's no need to
+	// additionally compute CRC32c.
+	// See https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-sctp-zero-checksum
+	if (config.sctpZeroChecksum) {
+		int on = 1;
+		if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_ACCEPT_ZERO_CHECKSUM, &on, sizeof(on)))
```
paullouisageneau, line 292: I don't think the parameter does what you expect: it doesn't disable checksums for outgoing packets, it only tells the remote side that this side is willing to accept zero-checksum packets.

src/impl/sctptransport.cpp
```diff
@@ -284,6 +284,16 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
 		throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
 		                         std::to_string(errno));
 
+	// When using SCTP over DTLS, the data integrity is ensured by DTLS, so there's no need to
+	// additionally compute CRC32c.
+	// See https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-sctp-zero-checksum
+	if (config.sctpZeroChecksum) {
```
paullouisageneau, line 290: The feature should be enabled by default as it handles retro-compatibility automatically.

paullouisageneau (commented):
I don't think this works, since checksums for outgoing packets will always be computed here, even if the remote side is willing to accept zero-checksum packets: https://github.com/paullouisageneau/libdatachannel/blob/283488d78959ba211df71480cbe3734212209179/src/impl/sctptransport.cpp#L977

src/impl/sctptransport.cpp
paullouisageneau: Additionally, the expected parameter is not a boolean, the value should be instead set to `SCTP_EDMID_LOWER_LAYER_DTLS` as indicated [in the draft](https://datatracker.ietf.org/doc/html/draft-ietf-tsvwg-sctp-zero-checksum#section-7.1).

melpon:
I see, it seems I misunderstood the specification. Thank you for the comment. I'll make the corrections.

melpon:
I had been working on an implementation to eliminate the CRC calculation during transmission, but there were some issues.

The first issue is that it seems usrsctp users cannot be informed when `SCTP_ZERO_CHECKSUM_ACCEPTABLE` is received. As a result, it's not possible to determine within sctptransport.c whether there's a need to compute the CRC during transmission.

Thus, I decided to disable CRC offloading and let usrsctp handle the CRC computation. As long as we include `usrsctp_setsockopt(IPPROTO_SCTP, SCTP_ACCEPT_ZERO_CHECKSUM)`, the CRC check during reception should not be conducted, similar to when CRC offloading is enabled.

The second issue is that, when usrsctp is connected using the AF_CONN address family, it appears to compute the CRC even if `SCTP_ACCEPT_ZERO_CHECKSUM` has been received from the server. This seems like a bug, so I've submitted a Pull Request addressing this at https://github.com/sctplab/usrsctp/pull/690. However, it's uncertain if or when it will be merged.

If there's a better implementation approach, please let me know.

paullouisageneau (approved):
It looks good now, thank you! I hope the issue with AF_CONN sockets in usrsctp can be sorted out, thanks for submitting the PR.

PR #1019, opened 2023-10-28
Add PLI handler to C API

chobie:
regarding #999.  I've added missing Pli handler to C API.

I've confirmed up to the part where PliHandler is called in my local environment.


paullouisageneau (approved):
Thanks for adding!

PR #1023, opened 2023-10-29
Forward RTP/RTCP messages unconditionally if there is only one track

wusspuss:


src/impl/peerconnection.cpp
```diff
@@ -586,6 +586,9 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 		if (auto track = it->second.lock())
 			track->incoming(message);
 	} else {
+		if (mTrackLines.size()==1) {
+			mTrackLines.front().lock()->incoming(message);
```
paullouisageneau, line 590: `lock()` may return `nullptr` if the `Track` object has been destroyed. You need to check to check it otherwise it will crash as soon as the user releases the shared pointer:
```suggestion
			if (auto track = mTrackLines.front().lock())
			    track->incoming(message);
```

paullouisageneau (commented):
This implements forwarding for RTP packets but not for RTCP ones. I think the check for single-track case should be at the beginning of `dispatchMedia` to not even go through the SSRC lookup if there is only one track. In that case `mTracksMutex` should be locked for the entire method.

paullouisageneau (approved):
It looks good, thank you.

PR #1028, opened 2023-11-02
Correct NTP of Sender Report

tljdebrouwer:
Hello,

I had an issue when calling preparePacket() on a reportBlock and looking into the sender report results (arm 64-bit instance). The way the NTP timestamp was mangled wasn't correct.

As per spec it is doing the correct bit shift, however when doing the host to network byte conversion we only care about the result (which is an uint32_t). Currently it performed a host to network byte order on a uint64_t and then converted it into a uint32_t (the type of _lastReport) which, depending on the system, can provide an invalid result. 

In the updated code:
- input is an 64-bit value
- we shift it 16 right
- then take the present 32-bits (which are basically bits 16 to 48 of the original 64-bit value)
- convert it depending on the OS and store the value.

Thanks.


paullouisageneau (approved):
Good catch, thanks

PR #1035, opened 2023-11-15
Expose DTLS connection info to PeerConnection

gasperov:


src/impl/dtlstransport.cpp
```diff
@@ -179,6 +179,20 @@ void DtlsTransport::postHandshake() {
 	// Dummy
 }
 
+DtlsTransport::DtlsInfo DtlsTransport::getDtlsInfo() {
+	DtlsInfo info;
+	{
+		std::lock_guard lock(mSslMutex);
+		auto desc = gnutls_session_get_desc(mSession);
```
gasperov, line 186: No function to get version and suite that is similar to openssl ?

PR #1037, opened 2023-11-15
Fix the target location of the FU header pointer

melpon:
`fragmentIndicator()` returns a `H265NalUnitHeader*`, so `fragmentIndicator() + H265_NAL_HEADER_SIZE` results in a 4-byte offset.


paullouisageneau (approved):
Good catch, thank you for the fix

PR #1041, opened 2023-11-20
Set SNI for MbedTLS

melpon:
In OpenSSL, SNI was set, but it wasn't set in MbedTLS, so I have now configured it to be set in MbedTLS as well.


paullouisageneau (approved):
Thanks for adding!

PR #1055, opened 2023-11-24
Add #include <algorithm> to fix building with gcc 14

kostadinsh:
With gcc 14 some C++ Standard Library headers have been changed to no longer include other headers that were used internally by the library. In libdatachannel's case it is the `<algorithm>` header.

GCC 14 porting guide: https://gcc.gnu.org/gcc-14/porting_to.html#header-dep-changes


paullouisageneau (approved):
Thanks for the fix!

PR #1061, opened 2023-11-28
Ensure rtpExtHeaderSize is a multiple of 4

melpon:
As defined in RFC3550, headerLength counts the number of 32-bit words. Therefore, rtpExtHeaderSize must always be a multiple of 4.


paullouisageneau (approved):
Good catch, `rtpExtHeaderSize` must indeed be rounded up to a multiple of 4 before creating the message. Thank you for the fix.

Issue #1054, opened 2023-11-24
SSL client certificate in impl::TlsTransport


ilsalvopss:
While initializing the SSL context, `impl::TlsTransport` seems to assume that, if a certificate is provided, we're acting as a server. Indeed `SSL_CTX_set_default_verify_paths` is called in the else branch only if we don't pass a certificate pointer to the constructor.

https://github.com/paullouisageneau/libdatachannel/blob/cafd410f190dd98e4802ce3ec5e864b72ffd7755/src/impl/tlstransport.cpp#L595C1-L603

As this is used by the `WebSocket` implementation, I'd like to be able to provide a client certificate for mutual TLS authentication while still verifying server certificate using default paths.

I'd like to draft some changes to allow this but I'm not sure if I'm instead misinterpreting OpenSSL's API.

paullouisageneau:
Your understanding is correct, the current implementation of `impl::TlsTransport` assumes that we either set a certificate or verify the remote certificate. In practice it works as the client can't set a certificate. If you plan to add such possibility, you would need to change the logic to call `SSL_CTX_set_default_verify_paths` when `mIsClient` is `true`.

ilsalvopss:
Here is a first draft:

https://github.com/ilsalvopss/libdatachannel/commit/efe808764e948a17a585423e290656ea37e1473c
https://github.com/ilsalvopss/libdatachannel/commit/a568712ab89fb10f0d00312e42938a89452e2338

As we cannot access `Certificate` from outside `impl`, I introduced `struct rtc::WebSocket::ClientCertificateInfo` and delegated the certificate construction to an alternative `impl::WebSocket` constructor; this way `WebSocketServer` can continue working as before.

If this seems reasonable I'll make a PR.

paullouisageneau:
> As we cannot access `Certificate` from outside `impl`, I introduced `struct rtc::WebSocket::ClientCertificateInfo` and delegated the certificate construction to an alternative `impl::WebSocket` constructor; this way `WebSocketServer` can continue working as before.

[...]

PR #1062, opened 2023-11-29
WebSocket: allow clients to provide a TLS certificate

ilsalvopss:
As discussed in #1054

### DIFF
--- a/src/websocket.cpp
+++ b/src/websocket.cpp
@@ -13,13 +13,14 @@
 
 #include "impl/internals.hpp"
 #include "impl/websocket.hpp"
+#include "impl/certificate.hpp"
 
 namespace rtc {
 
 WebSocket::WebSocket() : WebSocket(Configuration()) {}
 
-WebSocket::WebSocket(Configuration config)
-    : CheshireCat<impl::WebSocket>(std::move(config)),
+WebSocket::WebSocket(const Configuration& config)
+    : CheshireCat<impl::WebSocket>(config, loadCertificate(config)),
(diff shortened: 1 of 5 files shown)


src/websocket.cpp
```diff
@@ -13,13 +13,14 @@
...
 #include "impl/websocket.hpp"
+#include "impl/certificate.hpp"
 
 namespace rtc {
 
 WebSocket::WebSocket() : WebSocket(Configuration()) {}
 
-WebSocket::WebSocket(Configuration config)
-    : CheshireCat<impl::WebSocket>(std::move(config)),
+WebSocket::WebSocket(const Configuration& config)
```
paullouisageneau, line 22: I see why you need such a change, but I'm not a fan of changing the signature here, because it breaks consistency with the rest of the API and prevents from moving the configuration all the way by forcing a copy when constructing `impl::WebSocket`.

Could you please move the content of `loadCertificate()` to the body of the constructor of `impl::WebSocket` instead? You can check if `certificate` is null there and attempt to load from `config` if so.

paullouisageneau (commented):
Thank you for the PR. I think it could make sense to move certificate loading to `impl::WebSocket::WebSocket()` so you don't have to change `config` to a reference.

src/websocket.cpp
ilsalvopss: Sure. I moved the whole helper to preserve `mCertificate` constness.

paullouisageneau (approved):
It looks good, thank you!

PR #1069, opened 2023-12-08
Fix RtcpReportBlock::setPacketsLost()

ognkrmms:
The previous version used logical and operator instead of bitwise.

### DIFF
--- a/src/rtp.cpp
+++ b/src/rtp.cpp
@@ -171,7 +171,7 @@ void RtcpReportBlock::setSSRC(SSRC in_ssrc) { _ssrc = htonl(in_ssrc); }
 
 void RtcpReportBlock::setPacketsLost(uint8_t fractionLost,
                                      unsigned int packetsLostCount) {
-	_fractionLostAndPacketsLost = ((uint32_t)fractionLost << 24) & htonl(packetsLostCount);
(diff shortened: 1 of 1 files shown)


src/rtp.cpp
```diff
@@ -171,7 +171,7 @@ void RtcpReportBlock::setSSRC(SSRC in_ssrc) { _ssrc = htonl(in_ssrc); }
 
 void RtcpReportBlock::setPacketsLost(uint8_t fractionLost,
                                      unsigned int packetsLostCount) {
-	_fractionLostAndPacketsLost = ((uint32_t)fractionLost << 24) && htonl(packetsLostCount);
+	_fractionLostAndPacketsLost = ((uint32_t)fractionLost << 24) & htonl(packetsLostCount);
```
paullouisageneau, line 174: Why is bitwise AND even used here? And isn't the htonl conversion actually wrong? I think it should be:
```suggestion
	_fractionLostAndPacketsLost = htonl((uint32_t(fractionLost) << 24) | (packetsLostCount & 0xFFFFFF));
```

ognkrmms: You are right, it has to be OR. When I saw the logical operator instead of bitwise, I thought changing the operator would suffice. 
For your second question: I was a little bit confused about endianness at the time of writing here therefore I thought this way circumvents the endiannes of the system but I see it is redundant now.

paullouisageneau (approved):

PR #1075, opened 2023-12-20
Use prime256v1 for EVP_EC_gen instead of P-256

Sean-Der:
The alias P-256 isn't available in all versions of OpenSSL


Sean-Der:
If it is ok @paullouisageneau could I also get a tag!

This is for a OBS bug fix. Bumping the dependency will require a tag.

paullouisageneau (approved):
Thanks for the fix, I'll release `0.19.5` with it.

Sean-Der:
Thank you @paullouisageneau !

Mind if I do my others also? I have to fix certificate issue and add H264 depackization. Only one libdatachannel upgrade will be easier for everyone 

thanks for fast review!

paullouisageneau:
@Sean-Der I released this fix in 0.19.5, others will probably introduce API changes so I'll tag them as 0.20.

Sean-Der:
👍 thank you so much!

PR #1086, opened 2024-01-04
Add maxMessageSize to Websocket configuration

Nyran:
The commits associated with this pull request add a member to the configuration struct of the websocket implementation-class.

If the option is set in the configuration struct being passed to the Websocket constructor, the configured value is used instead of the message size limit as defined in src/impl/internals.hpp.


paullouisageneau (approved):
Looks good, thanks.

PR #1091, opened 2024-01-11
Configure datachannel target_properties for APPLE

Sean-Der:
After 7591b9695c libdatachannel stopped working with OBS on macOS[0].
XCode only resolves one level of symlinks [1]. libdatachannel
is generating two levels.

This commit updates the shared object to only use only the major version
in the name.

[0] https://stackoverflow.com/questions/29946961/xcode-copy-files-build-phase-and-symlinks
[1] https://github.com/obsproject/obs-deps/pull/204#discussion_r1449200874

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,13 @@ set_target_properties(datachannel PROPERTIES
 	SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
 	CXX_STANDARD 17
 	CXX_VISIBILITY_PRESET default)
+
+if(APPLE)
+	set_target_properties(datachannel PROPERTIES
+		VERSION ${PROJECT_VERSION_MAJOR}
+		SOVERSION ${PROJECT_VERSION_MAJOR})
+endif()
+
 target_compile_definitions(datachannel PRIVATE RTC_EXPORTS)
 
 add_library(datachannel-static STATIC EXCLUDE_FROM_ALL


Sean-Der:
@PatTheMav does this look ok?

CMakeLists.txt
```diff
@@ -262,6 +262,14 @@ set_target_properties(datachannel PROPERTIES
 	SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
 	CXX_STANDARD 17
 	CXX_VISIBILITY_PRESET default)
+
+if(APPLE)
+	set_target_properties(datachannel PROPERTIES
+		SOVERSION ${PROJECT_VERSION}
+		MACHO_COMPATIBILITY_VERSION 0
```
PatTheMav, line 269: Compatibility version should be at least 1.0.0, but setting a default value would also remove any possibility to prohibit loading the library after a breaking ABI change.

Sean-Der: Changed to `1.0.0`

CMakeLists.txt
```diff
@@ -262,6 +262,14 @@ set_target_properties(datachannel PROPERTIES
 	SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
 	CXX_STANDARD 17
 	CXX_VISIBILITY_PRESET default)
+
+if(APPLE)
+	set_target_properties(datachannel PROPERTIES
+		SOVERSION ${PROJECT_VERSION}
```
RytoEX, line 268: Shouldn't you override VERSION to `0` here?

Sean-Der:
@PatTheMav I am reading the CMake docs and it seems to recommend NOT setting [NO_SONAME](https://cmake.org/cmake/help/latest/prop_tgt/NO_SONAME.html)

> use this property only for leaf private libraries or plugins. If you use it on normal shared libraries which other targets link against, on some platforms a linker will insert a full path to the library (as specified at link time) into the dynamic section of the dependent binary.

When I set `NO_SONAME` in libdatachannel's CMakeLists.txt it also breaks the test. I have no idea what other impact it will have on dependent projects. I don't understand the benefit of these changes, but I do see some negatives.

@paullouisageneau what do you think is best here?

CMakeLists.txt
Sean-Der: 👍 

So this PR doesn't follow the suggestions from [here](https://github.com/obsproject/obs-deps/pull/204#discussion_r1449211126). When I apply those changes it breaks the tests. I worry it will break others that depend on libdatachannel also.

The PR in its current from is just reverting 7591b9695c partially.

PatTheMav:
> @PatTheMav I am reading the CMake docs and it seems to recommend NOT setting [NO_SONAME](https://cmake.org/cmake/help/latest/prop_tgt/NO_SONAME.html)
> 
> > use this property only for leaf private libraries or plugins. If you use it on normal shared libraries which other targets link against, on some platforms a linker will insert a full path to the library (as specified at link time) into the dynamic section of the dependent binary.
> 
> When I set `NO_SONAME` in libdatachannel's CMakeLists.txt it also breaks the test. I have no idea what other impact it will have on dependent projects. I don't understand the benefit of these changes, but I do see some negatives.
> 
> @paullouisageneau what do you think is best here?

The `SONAME` is only used to set the `install_name` on the generated library by CMake, which it prepends with the `@rpath/` prefix. It is expected that an app binary has a set of rpath entries it uses to replace the prefix with when resolving library paths.

Xcode does that by itself, but if you do not use Xcode to build the library for macOS, you need to add that manually: `target_link_options(datachannel PRIVATE -install_name "@rpath/libdatachannel.dylib")`. We dropped all support for non-Xcode generators on macOS for OBS Studio, which is why we can omit that step.

FWIW, omitting the `NO_SONAME` property and ensuring a single version token is used (e.g. `libdatachannel.5.dylib`) is fine and should work as long as there is no additional symlink redirection. Setting the MACHO properties is still recommended though (as those are the actual properties checked by the library loader).

CMakeLists.txt
```diff
@@ -262,6 +262,14 @@ set_target_properties(datachannel PROPERTIES
...
+		MACHO_COMPATIBILITY_VERSION 1.0.0
+		MACHO_CURRENT_VERSION ${PROJECT_VERSION})
```
paullouisageneau, line 270: `MACHO_COMPATIBILITY_VERSION` and `MACHO_CURRENT_VERSION` were introduced in CMake 3.17, so you should bump `cmake_minimum_required` to 3.17.

CMakeLists.txt
paullouisageneau: @PatTheMav Does this mean that `MACHO_COMPATIBILITY_VERSION` should now be manually increased to track ABI compatibility specifically for MacOS? I'd prefer not to have another version to update on each release, would it be possible to set it to something like `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.0`? Otherwise, it is a bit issue to set a dummy value here in practice?

PatTheMav: I was under the impression that it could be set to `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.0` or whichever other version number is used for ABI compatibility can be used. That's up to the project.

PatTheMav: > I was under the impression that it could be set to `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.0` or whichever other version number is used for ABI compatibility can be used. That's up to the project.

`${PROJECT_VERSION_MAJOR}.0.0` is also possible if ABI compatibility is ensured within the same major version.

The requirements are outlined on the associated manage:

> Specifies the compatibility version number of the library.  When a library is loaded by dyld, the compatibility version is checked and if the program's version is greater that the library's version, it is an error.
>
> If the compatibility version number is not specified, it has a value of 0 and no checking is done when the library is used.

That's also why it should not default to `0.0.0`.

PatTheMav:
I took a look at the `dyld` source code (https://opensource.apple.com/source/dyld/dyld-852.2/src/ImageLoader.cpp.auto.html) and it seems that contrary to my understanding (and what the manpage states), checks of the compatibility version are skipped since macOS 10.14 (macOS Mojave).

As such I stand corrected that using a single version-token SONAME (e.g. `libdatachannel.1.dylib`) and incrementing that single number on an ABI break is the correct approach (alternatively A/B/C can be used similar to how Frameworks are versioned, but that's entirely semantics). The MACHO ELF headers will still work on macOS 10.13 and older.

Annoying, but the more you know...

CMakeLists.txt
paullouisageneau: Great, so `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.0` sounds good to me. @Sean-Der does the value work?

paullouisageneau:
@PatTheMav MacOS 10.13 is legacy now, right? Is there still a point in setting the `MACHO_` properties?

CMakeLists.txt
PatTheMav: Alright I ran a few more tests on the change - the root issue is CMake creating 2 layers of symlinks, which is the main thing we attempt to fix:

* Setting `SOVERSION` and `VERSION` to the _same_ value avoids the 2 layers of symlinks.
* As such we should set _both_ to `${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}` (e.g. 0.18) which generates a single dynamic library file (`libdatachannel.0.18.dylib`) and a single symlink (`libdatachannel.dylib`)
* This ensures that when the minor version changes, the link requirement is not resolved anymore
* The MACHO headers can stay as is if the _full_ version strings are desired in the headers, otherwise they inherit the generic version parameters (albeit truncated to the minor version, so 0.18.0 in this example)

PatTheMav:
> @PatTheMav MacOS 10.13 is legacy now, right? Is there still a point in setting the `MACHO_` properties?

See https://github.com/paullouisageneau/libdatachannel/pull/1091#discussion_r1450975892 - the `MACHO_` properties can be kept if the _exact_ version numbers are desired to be visible via `otool` on the library and consumers of the library, but as far as breaking the link on ABI change is required, it doesn't seem necessary anymore.

paullouisageneau:
@PatTheMav Thank you for testing and clarifying!

@Sean-Der If you want to keep the `MACHO_`  properties even if they are not strictly necessary, they were introduced in CMake 3.17, so you must bump `cmake_minimum_required` to 3.17.

Sean-Der:
@PatTheMav can/should we drop the MACHO values?

I don’t care what we do, my hope/goal is
* Works with OBS
* Doesn’t impact other libdatachannel users

@paullouisageneau would you be ok with a new  var? `MACOS_SO_VERSION` and we start it at `0`. I would put it next to VERSION in CmakeLists.txt. From this thread that seems idiomatic?

PatTheMav:
> @PatTheMav can/should we drop the MACHO values?
> 
> I don’t care what we do, my hope/goal is
> 
> * Works with OBS
> * Doesn’t impact other libdatachannel users
> 
> @paullouisageneau would you be ok with a new var? `MACOS_SO_VERSION` and we start it at `0`. I would put it next to VERSION in CmakeLists.txt. From this thread that seems idiomatic?
> 
> When a v0.20.0 is ready will then bump OBS and resolve issues.

You can drop them, the values will be inferred from `VERSION`/`SOVERSION` then, so they'll always be limited to <Major>.<Minor>.0 but that shouldn't introduce major issues.

Sean-Der:
@PatTheMav @paullouisageneau @RytoEX How does this look? I think it hits everyones concerns.

* It doesn't require setting/maintaining a new variable
* We only have one symlink
* Just puts the major version number in the name

paullouisageneau (approved):
@Sean-Der It looks good to me. Would you like the change on 0.19? If so, could you please rebase on the `v0.19` branch?

Sean-Der:
@paullouisageneau no thanks! v0.20 only (if that is ok)

That way I can fix both bugs

PR #1112, opened 2024-02-16
Add version macros

### LINKED ISSUE #1111: Add version macros
Requested by OBS

I will add major/minor/patch and have it be generated by CMake (so we don't need to update multiple things)

Sean-Der:
Add version.h with the following values. This values are managed/updated via CMake

```
#pragma once

#define RTC_VERSION_MAJOR 0
#define RTC_VERSION_MINOR 20
#define RTC_VERSION_PATCH 1
#define RTC_VERSION "0.20.1"
```
Resolves #1111


paullouisageneau (commented):
It looks like the `version.h` header is only put in place at installation, so the `RTC_VERSION` defines are absent when adding the library to a project directly with `add_subdirectory()`. There are also not available to the library itself and examples. This can probably be fixed by adding something like `target_include_directories(libdatachannel[-static] PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>/include/rtc)` but relying on include directory search order seems a bit hacky. 

Another issue is that the defines are also absent if the library is built with the Makefile or the Jamfile. It seems to me that common practice to support multiple build systems is a `config.h` file that the different build systems configure but I would prefer not to go there, I like that the code compiles as-is for the sake of simplicity.

IMHO it would be better to hardcode the version in `include/rtc/version.h` so it just works (current defines look good). I'll update them when releasing, and possibly read them from CMakeLists later so the version number is in a single place. What do you think?

Sean-Der:
Done! @paullouisageneau can I get another review?

paullouisageneau (approved):
Looks good, thank you!

PR #1121, opened 2024-02-29
Add FrameInfo

Sean-Der:
Depacketizers will include metadata about assembled frame.

PR #1123, opened 2024-02-29
Add OpusRtpDepacketizer

Sean-Der:
Inverse of OpusRtpPacketizer. Takes incoming Opus packets and emits frames.

### DIFF
--- a/include/rtc/rtpdepacketizer.hpp
+++ b/include/rtc/rtpdepacketizer.hpp
@@ -1,0 +1,34 @@
+/**
+ * Copyright (c) 2024 Paul-Louis Ageneau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef RTC_RTP_DEPACKETIZER_H
+#define RTC_RTP_DEPACKETIZER_H
+
+#if RTC_ENABLE_MEDIA
+
+#include "mediahandler.hpp"
+#include "message.hpp"
+
+namespace rtc {
+
+class RTC_CPP_EXPORT RtpDepacketizer : public MediaHandler {
+public:
+	RtpDepacketizer() = default;
+	virtual ~RtpDepacketizer() = default;
+
+	virtual void incoming(message_vector &messages, const message_callback &send) override;
+};
+
+using OpusRtpDepacketizer = RtpDepacketizer;
+using AACRtpDepacketizer = RtpDepacketizer;
+
+} // namespace rtc
+
+#endif /* RTC_ENABLE_MEDIA */
+
+#endif /* RTC_RTP_DEPACKETIZER_H */
--- a/src/rtpdepacketizer.cpp
+++ b/src/rtpdepacketizer.cpp
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) 2024 Paul-Louis Ageneau
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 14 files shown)


include/rtc/rtpdepacketizer.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#if RTC_ENABLE_MEDIA
+
+#include "mediahandler.hpp"
+#include "message.hpp"
+
+namespace rtc {
+
+/// RTP depacketization for Opus
+class RTC_CPP_EXPORT OpusRtpDepacketizer : public MediaHandler {
```
paullouisageneau, line 20: You can actually rename it `AudioRtpDepacketizer` and define depacketizers for both Opus and AAC with:
```
using OpusRtpDepacketizer = AudioRtpDepacketizer;
using AACRtpDepacketizer = AudioRtpDepacketizer;
```
Packetizers are identical too and defined this way.

src/rtpdepacketizer.cpp
```diff
@@ -0,0 +1,50 @@
...
+		                              }
+
+		                              return false;
+	                              }),
+	               messages.end());
+
+	for (auto &message : messages) {
+		auto pkt = reinterpret_cast<const rtc::RtpHeader *>(message->data());
+		auto headerSize = sizeof(rtc::RtpHeader) + pkt->csrcCount() + pkt->getExtensionHeaderSize();
+		message->erase(message->begin(), message->begin() + headerSize);
```
paullouisageneau, line 44: I think it's safer to create a new message here. Also, it's consistent with how packetizers operate.
```suggestion
		message = make_message(message->begin() + headerSize, message->end());
```

paullouisageneau (commented):
It looks good, thanks for adding! Since Opus and AAC depacketizers are identical, do you want to define both here as suggested?

include/rtc/rtpdepacketizer.hpp
Nemirtingas: Any reason why you want to prefix Audio while the RtcPacketizer doesn't ?

paullouisageneau: Good question, I guess it could be named `RtpDepacketizer` too. Opus and AAC packetizers are defined as `AudioRtpPacketizer`, so it made sense to do something symmetrical. An `RtpDepacketizer` base class could make sense especially if there is common logic in the future (possibly stuff related to frame info).

paullouisageneau: Good question, I guess it could be named `RtpDepacketizer` too. Opus and AAC packetizers are defined as `AudioRtpPacketizer`, so it made sense to do something symmetrical.
https://github.com/paullouisageneau/libdatachannel/blob/437a758ae89e5342940c139b8a647bf1bb1c0ee5/include/rtc/rtppacketizer.hpp#L61

An `RtpDepacketizer` base class could make sense especially if there is common logic in the future (possibly stuff related to frame info).

Sean-Der:
Thanks for the feedback @Nemirtingas @paullouisageneau I have updated the PR. Could I get another review?

Sean-Der:
@paullouisageneau mind waiting to merge this until #1121 is done? I will update this to emit timestamps properly then

Sean-Der:
I pulled `FrameInfo` in here as well. I am hoping after the last review I got it right :)

Sean-Der:
@paullouisageneau I updated with the two changes from other PR

* Filter bad RTP packets
* Rename `frame_info` to `frameInfo`

ready for a review!

Sean-Der:
Build failure is homebrew HTTP failed respond, nothing to do with code changes

paullouisageneau (approved):
It looks good, thank you!

PR #1127, opened 2024-03-05
Support dependency descriptor and two-byte header for RTP header extension

melpon:
I implemented the [Dependency Descriptor RTP Header Extension](https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension) in `AV1RtpPacketizer`. This allows SFUs to determine the dependencies between frames and the presence of keyframes without inspecting the RTP payload, enabling them to appropriately forward video frames.

Additionally, since the body size of the DD could become larger than 16 bytes, I also implemented the [Two-Byte Header](https://datatracker.ietf.org/doc/html/rfc8285#section-4.3).

### DIFF
--- a/src/dependencydescriptor.cpp
+++ b/src/dependencydescriptor.cpp
@@ -1,0 +1,407 @@
+/**
+ * Copyright (c) 2024 Shigemasa Watanabe (Wandbox)
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#include "dependencydescriptor.hpp"
+
+#include <algorithm>
+#include <functional>
+#include <stdexcept>
+
+namespace rtc {
+
+BitWriter BitWriter::fromSizeBits(std::byte *buf, size_t offsetBits, size_t sizeBits) {
+	BitWriter writer;
+	writer.mBuf = buf;
+	writer.mInitialOffset = offsetBits;
+	writer.mOffset = offsetBits;
+	writer.mSize = sizeBits;
+	return writer;
+}
+BitWriter BitWriter::fromNull() {
+	BitWriter writer;
+	writer.mSize = SIZE_MAX;
+	return writer;
+}
+
+size_t BitWriter::getWrittenBits() const { return mOffset - mInitialOffset; }
+
+bool BitWriter::write(uint64_t v, size_t bits) {
+	if (mOffset + bits > mSize) {
+		return false;
+	}
+	uint8_t *p = mBuf == nullptr ? nullptr : reinterpret_cast<uint8_t *>(mBuf + mOffset / 8);
+	// First, write up to the 8-bit boundary
+	size_t written_bits = writePartialByte(p, mOffset % 8, v, bits);
+
+	if (p != nullptr) {
+		p++;
+	}
+	bits -= written_bits;
+	mOffset += written_bits;
+
+	if (bits == 0) {
+		return true;
+	}
+
+	// Write 8 bits at a time
+	while (bits >= 8) {
+		if (p != nullptr) {
+			*p = (v >> (bits - 8)) & 0xff;
+			p++;
+		}
+		bits -= 8;
+		mOffset += 8;
+	}
+
+	// Write the remaining bits
+	written_bits = writePartialByte(p, 0, v, bits);
+	bits -= written_bits;
+	mOffset += written_bits;
+
+	assert(bits == 0);
+
+	return true;
+}
+
+bool BitWriter::writeNonSymmetric(uint64_t v, uint64_t n) {
+	if (n == 1) {
+		return true;
+	}
+	size_t w = 0;
+	uint64_t x = n;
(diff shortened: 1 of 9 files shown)


src/rtppacketizer.cpp
```diff
@@ -21,26 +21,48 @@ RtpPacketizer::~RtpPacketizer() {}
...
+	bool twoByteHeader = false;
+
+	const bool setVideoRotation =
+	    (rtpConfig->videoOrientationId != 0) && mark && (rtpConfig->videoOrientation != 0);
+
+	// Determine if a two-byte header is necessary
+	// Check for dependency descriptor extension
+	if (rtpConfig->dependencyDescriptorContext.has_value()) {
+		auto &ctx = *rtpConfig->dependencyDescriptorContext;
+		DependencyDescriptorWriter writer(ctx.structure, ctx.activeChains, ctx.descriptor);
```
paullouisageneau, line 33: Since you pass every member of the context to the writer, why not referencing the entire context in the writer and constructing it with `DependencyDescriptorWriter(DependencyDescriptorContext *ctx)`?

src/rtppacketizer.cpp
```diff
@@ -21,26 +21,48 @@ RtpPacketizer::~RtpPacketizer() {}
...
-		rtpExtHeaderSize += (1 + rtpConfig->mid->length());
+		rtpExtHeaderSize += headerSize + rtpConfig->mid->length();
 
 	if (rtpConfig->rid.has_value())
-		rtpExtHeaderSize += (1 + rtpConfig->rid->length());
+		rtpExtHeaderSize += headerSize + rtpConfig->rid->length();
+
+	if (rtpConfig->dependencyDescriptorContext.has_value()) {
+		auto &ctx = *rtpConfig->dependencyDescriptorContext;
+		DependencyDescriptorWriter writer(ctx.structure, ctx.activeChains, ctx.descriptor);
```
paullouisageneau, line 63: The writer should be created only once in the method.

src/rtppacketizer.cpp
```diff
@@ -71,24 +93,33 @@ message_ptr RtpPacketizer::packetize(shared_ptr<binary> payload, bool mark) {
...
+			    extHeader->writeHeader(twoByteHeader, offset, rtpConfig->ridId,
+			                           reinterpret_cast<const std::byte *>(rtpConfig->rid->c_str()),
+			                           rtpConfig->rid->length());
+		}
+
+		if (rtpConfig->dependencyDescriptorContext.has_value()) {
+			auto &ctx = *rtpConfig->dependencyDescriptorContext;
+			DependencyDescriptorWriter writer(ctx.structure, ctx.activeChains, ctx.descriptor);
+			auto sizeBits = writer.getSizeBits();
+			size_t sizeBytes = (sizeBits + 7) / 8;
```
paullouisageneau, line 118: There should be a getter `getSize()` returning size in bytes on the writer as it's used multiple times.

src/rtppacketizer.cpp
```diff
@@ -71,24 +93,33 @@ message_ptr RtpPacketizer::packetize(shared_ptr<binary> payload, bool mark) {
...
+			std::unique_ptr<std::byte[]> buf(new std::byte[sizeBytes]);
```
paullouisageneau, line 119: This should be a `std::vector<byte>`.

src/dependencydescriptor.cpp
```diff
@@ -0,0 +1,407 @@
...
+	BitWriter writer;
+	writer.mBuf = buf;
+	writer.mInitialOffset = offsetBits;
+	writer.mOffset = offsetBits;
+	writer.mSize = sizeBits;
+	return writer;
+}
+BitWriter BitWriter::fromNull() {
+	BitWriter writer;
+	writer.mSize = SIZE_MAX;
```
paullouisageneau, line 27:
```suggestion
	writer.mSize = std::numeric_limits<size_t>::max();
```

src/dependencydescriptor.cpp
```diff
@@ -0,0 +1,407 @@
...
+    : mStructure(structure), mActiveChains(activeChains), mDescriptor(descriptor) {}
+
+size_t DependencyDescriptorWriter::getSizeBits() const {
+	auto writer = rtc::BitWriter::fromNull();
+	doWriteTo(writer);
+	return writer.getWrittenBits();
+}
+
+void DependencyDescriptorWriter::writeTo(std::byte *buf, size_t sizeBits) const {
+	auto writer = BitWriter::fromSizeBits(buf, 0, (sizeBits + 7) / 8 * 8);
```
paullouisageneau, line 219: This method should take size in bytes, because `buf` is in bytes, and also because you convert it anyway.

paullouisageneau (commented):
I'm really sorry for the late review, it got buried and I forgot about it. Thank you for adding the feature, I have a few comments but I can address them if you are not working on this anymore.

melpon:
Thank you for the code review. I have fixed all the pointed-out issues.
However, since this PR has already merged the changes from https://github.com/paullouisageneau/libdatachannel/pull/1364, this one needs to be merged first.

paullouisageneau (approved):
It looks good, thank you!

PR #1129, opened 2024-03-09
Fix rtcSetSignalingStateChangeCallback reset

gasperov:
Fixed wrong function called in rtcSetSignalingStateChangeCallback.


paullouisageneau (approved):
Thank you for the fix!

Issue #1130, opened 2024-03-09
Callback for rtcSetGatheringStateChangeCallback sometimes not firing for RTC_GATHERING_COMPLETE


gasperov:
I'm encountering an issue in my test based on capi connectivity (with libjuice), where callback (set with rtcSetGatheringStateChangeCallback) function isn't consistently being invoked for the RTC_GATHERING_COMPLETE status.

This behavior is observed when both sides of the connection are using TURN servers.

What could be the root cause for this issue? How can I debug it?
Could there be a problem with calling rtcSetRemoteDescription before callback from rtcSetLocalDescriptionCallback is finished?

I also noticed that juice_set_remote_gathering_done is not somehow exposed in libdatachannel. Should there be a call to juice_set_remote_gathering_done on remote side after local gets RTC_GATHERING_COMPLETE ?

gasperov:
Related to this issue I experimented with https://github.com/paullouisageneau/libdatachannel/pull/1131.

paullouisageneau:
> What could be the root cause for this issue? How can I debug it?

Is the callback never called or called only after a long time (like 1min)? If the callback is never called, you should ensure you set it early enough and you don't block any other callback from the library. If it is called after a long time, the probable cause is an offline ICE server.

The gathering complete state is only reached when every candidate has been gathered. If an ICE server does not answer, for instance because it is unreachable or overloaded, then the state transition can only happen after the client times out, which can take a relatively long time (more than 30s). This is why it is recommended to use candidates trickling instead of sending full descriptions on gathering done when using ICE servers, in particular TURN servers.

For more information, you should look at the verbose log to investigate what is blocking.

> Could there be a problem with calling rtcSetRemoteDescription before callback from rtcSetLocalDescriptionCallback is finished?

It should be fine, but it sounds like you block a callback, which is something you must not do. Could you please elaborate on the scenario where you end up with such a call order?

> I also noticed that juice_set_remote_gathering_done is not somehow exposed in libdatachannel. Should there be a call to juice_set_remote_gathering_done on remote side after local gets RTC_GATHERING_COMPLETE ?

It's not exposed because it's mostly useless. It's part of the API only for the sake of completeness, libjuice do not use the information for anything apart from adding the `end-of-candidates` marker to the remote description and preventing more remote candidates from being added.

gasperov:
@paullouisageneau thank you for your answer!

[...]

PR #1131, opened 2024-03-09
RFC: introduce function SetRemoteGatherDone

gasperov:
I was experimenting gathering complete signal. It seems juice_set_remote_gathering_done should be called at the end of candidate gathering phase.

The reason for this PR is issue https://github.com/paullouisageneau/libdatachannel/issues/1130. At this time I have no data to confirm that this PR fixes anything.

### DIFF
--- a/src/impl/icetransport.cpp
+++ b/src/impl/icetransport.cpp
@@ -202,6 +202,8 @@ void IceTransport::setRemoteDescription(const Description &description) {
 		throw std::invalid_argument("Invalid ICE settings from remote SDP");
 }
 
+void IceTransport::setRemoteGatherDone() { juice_set_remote_gathering_done(mAgent.get()); }
+
 bool IceTransport::addRemoteCandidate(const Candidate &candidate) {
 	// Don't try to pass unresolved candidates for more safety
 	if (!candidate.isResolved())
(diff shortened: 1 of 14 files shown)


src/impl/icetransport.cpp
```diff
@@ -664,6 +666,8 @@ void IceTransport::setRemoteDescription(const Description &description) {
 		throw std::invalid_argument("Invalid ICE settings from remote SDP");
 }
 
+void IceTransport::setRemoteGatherDone() {}
```
gasperov, line 669: Quick search didn't find similar function.

paullouisageneau:
There are a few issues with this:
- Calling `juice_set_remote_gathering_done()`  does not make any difference. Recent versions of libjuice do not use the information for anything apart from adding the end-of-candidates marker to the remote description and preventing more remote candidates from being added.
- To properly signal end-of-candidates, the mechanism would also need to update the libdatachannel state, it would at least need to update the remote description.
- The reference WebRTC API uses null candidate to signal end-of-candidates instead of relying on gathering state change, so it might be better to do something similar (signaling end of candidates makes sense only when trickling candidates, if you wait for gathering complete with description already has an end-of-candidate marker).
- This is unrelated to #1130

gasperov:
@paullouisageneau thank you for your review and the comment!

I think we can abandon this PR.

PR #1134, opened 2024-03-18
Add H265RtpDepacketizer

edmonds:
This commit adds an H265 depacketizer which takes incoming H265 RTP packets and emits H265 access units. It is closely based on the `H264RtpDepacketizer` added by @Sean-Der in https://github.com/paullouisageneau/libdatachannel/pull/1082.

I originally started with a version of this commit that was closer to the `H264RtpDepacketizer` and which emitted individual H265 NALUs in `H265RtpDepacketizer::buildFrames()`. This resulted in calling my `Track::onFrame()` callback for each NALU, which did not work well with the decoder that I'm using which wants to see the VPS/SPS/PPS NALUs as a unit before initializing the decoder (https://intel.github.io/libvpl/v2.10/API_ref/VPL_func_vid_decode.html#mfxvideodecode-decodeheader).

So for the `H265RtpDepacketizer` I've tried to make it emit access units rather than NALUs. An "access unit" is (RFC 7798):

> A set of NAL units that are associated with each other according to a specified classification rule, that are consecutive in decoding order, *and that contain exactly one coded picture.*

"Exactly one coded picture" seems to correspond with what a caller might expect an "onFrame" callback to do. Maybe the `H264RtpDepacketizer` should be revised to similarly emit H264 access units rather than NALUs, too. At least, I could not find a way to receive individual NALUs from the depacketizer and run the VPL decoder without needing to do my own buffering/copying of the NALUs.

With this commit I can now do the following:

* Generate encoded bitstream output from the Intel VPL encoder.
* Pass the output of the encoder one frame at a time to libdatachannel's `Track::send()` on a track with an `H265RtpPacketizer` media handler.
* Transport the video track over a WebRTC connection to a libdatachannel peer.
* Depacketize it with the `H265RtpDepacketizer` media handler in this commit.
* Pass the depacketized output via my `Track::onFrame()` callback to the Intel VPL decoder in "complete frame" mode (https://intel.github.io/libvpl/v2.10/API_ref/VPL_enums.html#_CPPv428MFX_BITSTREAM_COMPLETE_FRAME). Each "onFrame" callback corresponds to a single call to the decoder API to decode a frame.

### DIFF
--- a/src/h265rtpdepacketizer.cpp
+++ b/src/h265rtpdepacketizer.cpp
@@ -62,7 +62,6 @@ message_vector H265RtpDepacketizer::buildFrames(message_vector::iterator begin,
 	message_vector out = {};
 	auto accessUnit = binary{};
 	auto frameInfo = std::make_shared<FrameInfo>(payloadType, timestamp);
-	auto nFrags = 0;
 
 	for (auto it = begin; it != end; ++it) {
 		auto pkt = it->get();
@@ -87,7 +86,11 @@ message_vector H265RtpDepacketizer::buildFrames(message_vector::iterator begin,
 			auto nalUnitFragmentHeader = H265NalUnitFragmentHeader{
 			    std::to_integer<uint8_t>(pkt->at(rtpHeaderSize + sizeof(H265NalUnitHeader)))};
 
-			if (nFrags++ == 0) {
+			// RFC 7798: "When set to 1, the S bit indicates the start of a fragmented
+			// NAL unit, i.e., the first byte of the FU payload is also the first byte of
+			// the payload of the fragmented NAL unit. When the FU payload is not the start
+			// of the fragmented NAL unit payload, the S bit MUST be set to 0."
+			if (nalUnitFragmentHeader.isStart() || accessUnit.empty()) {
 				addSeparator(accessUnit);
 				nalUnitHeader.setUnitType(nalUnitFragmentHeader.unitType());
 				accessUnit.emplace_back(byte(nalUnitHeader._first));
--- a/src/h265rtpdepacketizer.cpp
+++ b/src/h265rtpdepacketizer.cpp
@@ -32,6 +32,17 @@ message_vector H265RtpDepacketizer::buildFrames(message_vector::iterator begin,
 		auto pkt = it->get();
 		auto pktParsed = reinterpret_cast<const rtc::RtpHeader *>(pkt->data());
 		auto rtpHeaderSize = pktParsed->getSize() + pktParsed->getExtensionHeaderSize();
+		auto rtpPaddingSize = 0;
+
+		if (pktParsed->padding()) {
+			rtpPaddingSize = std::to_integer<uint8_t>(pkt->at(pkt->size() - 1));
+		}
+
+		if (pkt->size() == rtpHeaderSize + rtpPaddingSize) {
+			PLOG_VERBOSE << "H.265 RTP packet has empty payload";
+			continue;
+		}
+
 		auto nalUnitHeader =
 		    H265NalUnitHeader{std::to_integer<uint8_t>(pkt->at(rtpHeaderSize)),
 		                      std::to_integer<uint8_t>(pkt->at(rtpHeaderSize + 1))};
--- a/src/h265rtpdepacketizer.cpp
+++ b/src/h265rtpdepacketizer.cpp
@@ -57,10 +57,11 @@ void H265RtpDepacketizer::addSeparator(binary& accessUnit)
 }
 
 message_vector H265RtpDepacketizer::buildFrames(message_vector::iterator begin,
-                                                message_vector::iterator end, uint32_t timestamp) {
+                                                message_vector::iterator end,
+                                                uint8_t payloadType, uint32_t timestamp) {
 	message_vector out = {};
 	auto accessUnit = binary{};
-	auto frameInfo = std::make_shared<FrameInfo>(timestamp);
+	auto frameInfo = std::make_shared<FrameInfo>(payloadType, timestamp);
 	auto nFrags = 0;
 
 	for (auto it = begin; it != end; ++it) {
@@ -154,6 +155,7 @@ void H265RtpDepacketizer::incoming(message_vector &messages, const message_callb
 	               messages.end());
 
 	while (mRtpBuffer.size() != 0) {
+		uint8_t payload_type = 0;
 		uint32_t current_timestamp = 0;
 		size_t packets_in_timestamp = 0;
 
@@ -162,6 +164,7 @@ void H265RtpDepacketizer::incoming(message_vector &messages, const message_callb
 
 			if (current_timestamp == 0) {
 				current_timestamp = p->timestamp();
+				payload_type = p->payloadType(); // should all be the same for data of the same codec
 			} else if (current_timestamp != p->timestamp()) {
 				break;
 			}
@@ -176,7 +179,7 @@ void H265RtpDepacketizer::incoming(message_vector &messages, const message_callb
 		auto begin = mRtpBuffer.begin();
 		auto end = mRtpBuffer.begin() + (packets_in_timestamp - 1);
 
-		auto frames = buildFrames(begin, end + 1, current_timestamp);
+		auto frames = buildFrames(begin, end + 1, payload_type, current_timestamp);
 		messages.insert(messages.end(), frames.begin(), frames.end());
 		mRtpBuffer.erase(mRtpBuffer.begin(), mRtpBuffer.begin() + packets_in_timestamp);
 	}
--- a/src/h265rtpdepacketizer.cpp
+++ b/src/h265rtpdepacketizer.cpp
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2023-2024 Paul-Louis Ageneau
+ * Copyright (c) 2024 Robert Edmonds
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#if RTC_ENABLE_MEDIA
+
+#include "h265rtpdepacketizer.hpp"
+#include "h265nalunit.hpp"
+
+#include "impl/internals.hpp"
+
+namespace rtc {
+
+const binary naluStartCode = {byte{0}, byte{0}, byte{0}, byte{1}};
+
+const uint8_t naluTypeAP = 48;
+const uint8_t naluTypeFU = 49;
+
+message_vector H265RtpDepacketizer::buildFrames(message_vector::iterator begin,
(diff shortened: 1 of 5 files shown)


edmonds:
By the way, producing access units rather than NALUs seems to be what FFmpeg's H265 RTP depacketizer does. It's [kind of awkward](https://github.com/edmonds/vacon/blob/c015350d6f3ac370fe00ccec67059d56fb5c60d3/src/rtp_depacketizer.cpp#L53-L184) to set up, though!

Some open items:

- Is it always correct to build an access unit by simply prepending the start code { 0x00, 0x00, 0x00, 0x01 } to each NALU?
- `H265RtpDepacketizer::incoming()` is identical to `H264RtpDepacketizer::incoming()` and doesn't even look all that specific to H264/H265. Should that function be factored out somehow?

Sean-Der:
Agree on NALU API being awkward!

do you have bandwidth to update H264 depacketizer I can test + approve? We should make that change ASAP before people depend on it.

or maybe this is controlled via constructor

edmonds:
> do you have bandwidth to update H264 depacketizer I can test + approve? We should make that change ASAP before people depend on it.

Yeah, I should be able to take a look later today.

paullouisageneau:
> "Exactly one coded picture" seems to correspond with what a caller might expect an "onFrame" callback to do. Maybe the `H264RtpDepacketizer` should be revised to similarly emit H264 access units rather than NALUs, too. At least, I could not find a way to receive individual NALUs from the depacketizer and run the VPL decoder without needing to do my own buffering/copying of the NALUs.

I think you are right, it corresponds better to what the user would expect the callback to do. Additionally, it makes the API more symmetrical with packetizers.

> Is it always correct to build an access unit by simply prepending the start code { 0x00, 0x00, 0x00, 0x01 } to each NALU?

This is the most common, but a short start sequence { 0x00, 0x00, 0x01 } also exists. Packetizers have an `NalUnit::Separator` constructor parameter to specify what they should expect, maybe depacketizers could have the same.

> H265RtpDepacketizer::incoming() is identical to H264RtpDepacketizer::incoming() and doesn't even look all that specific to H264/H265. Should that function be factored out somehow?

Yes it should be, but I can do it in a second time.

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,143 @@
...
+#if RTC_ENABLE_MEDIA
+
+#include "h265rtpdepacketizer.hpp"
+#include "h265nalunit.hpp"
+
+#include "impl/internals.hpp"
+
+namespace rtc {
+
+const binary naluStartCode = {byte{0}, byte{0}, byte{0}, byte{1}};
```
paullouisageneau, line 18: The start code could be selectable with a `NalUnit::Separator` parameter like with the packetizer.

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,143 @@
...
+		    H265NalUnitHeader{std::to_integer<uint8_t>(pkt->at(rtpHeaderSize)),
+		                      std::to_integer<uint8_t>(pkt->at(rtpHeaderSize + 1))};
+
+		if (nalUnitHeader.unitType() == naluTypeFU) {
+			auto nalUnitFragmentHeader = H265NalUnitFragmentHeader{
+			    std::to_integer<uint8_t>(pkt->at(rtpHeaderSize + sizeof(H265NalUnitHeader)))};
+
+			if (nFrags++ == 0) {
+				std::copy(naluStartCode.begin(), naluStartCode.end(),
+				          std::back_inserter(accessUnit));
```
paullouisageneau, line 44: `back_inserter` is handy but also [inefficient](https://thewayofc.blogspot.com/2014/05/copying-memory-from-c-to-c-using.html) as it ends up inserting elements (here, bytes) one by one. IMHO you should use `insert` for better performance here.

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+				nalUnitHeader.setUnitType(nalUnitFragmentHeader.unitType());
+				accessUnit.emplace_back(byte(nalUnitHeader._first));
+				accessUnit.emplace_back(byte(nalUnitHeader._second));
+			}
+
+			std::copy(pkt->begin() + rtpHeaderSize + sizeof(H265NalUnitHeader) +
+			              sizeof(H265NalUnitFragmentHeader),
+			          pkt->end(), std::back_inserter(accessUnit));
```
paullouisageneau, line 53: Same here

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+				if (pkt->size() < currOffset + naluSize) {
+					throw std::runtime_error("H265 AP declared size is larger than buffer");
+				}
+
+				std::copy(naluStartCode.begin(), naluStartCode.end(),
+				          std::back_inserter(accessUnit));
+
+				std::copy(pkt->begin() + currOffset, pkt->begin() + currOffset + naluSize,
+				          std::back_inserter(accessUnit));
```
paullouisageneau, line 71: Same here

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+				currOffset += naluSize;
+			}
+		} else if (nalUnitHeader.unitType() < naluTypeAP) {
+			// "NAL units with NAL unit type values in the range of 0 to 47, inclusive, may be
+			// passed to the decoder."
+			std::copy(naluStartCode.begin(), naluStartCode.end(), std::back_inserter(accessUnit));
+			std::copy(pkt->begin() + rtpHeaderSize, pkt->end(), std::back_inserter(accessUnit));
```
paullouisageneau, line 79: Same here

paullouisageneau (commented):
Thank you for adding, I have only a couple comments and suggestions.

src/h265rtpdepacketizer.cpp
edmonds: a0e2e06c97d4fd4eada5540fd07881f4f794c7a4

src/h265rtpdepacketizer.cpp
edmonds: a0e2e06c97d4fd4eada5540fd07881f4f794c7a4

src/h265rtpdepacketizer.cpp
edmonds: a0e2e06c97d4fd4eada5540fd07881f4f794c7a4

src/h265rtpdepacketizer.cpp
edmonds: a0e2e06c97d4fd4eada5540fd07881f4f794c7a4

src/h265rtpdepacketizer.cpp
edmonds: Sounds good. Just to clarify:

```cpp
/// NAL unit separator
enum class Separator {
        Length = RTC_NAL_SEPARATOR_LENGTH, // first 4 bytes are NAL unit length
        LongStartSequence = RTC_NAL_SEPARATOR_LONG_START_SEQUENCE,   // 0x00, 0x00, 0x00, 0x01
        ShortStartSequence = RTC_NAL_SEPARATOR_SHORT_START_SEQUENCE, // 0x00, 0x00, 0x01
        StartSequence = RTC_NAL_SEPARATOR_START_SEQUENCE, // LongStartSequence or ShortStartSequence
};
```

`LongStartSequence` and `ShortStartSequence` are easy, just write 00 00 00 01 or 00 00 01 into the buffer.

Should `StartSequence` just mean the same thing as `LongStartSequence`? Or `ShortStartSequence`?

`Length` is a little more complicated, we'd need to insert a placeholder for where the length should go and then fill it in later once all of the reassembled fragments have been written into the buffer and the total length can be calculated.

Or could we specify that only `LongStartSequence` and `ShortStartSequence` are supported for the value of this parameter?

paullouisageneau: > Should `StartSequence` just mean the same thing as `LongStartSequence`? Or `ShortStartSequence`?

`StartSequence` for packetizers means the parser can expect both the long and the short start sequence, and it will effectively use both as separators. Here it should mean the same as `LongStartSequence`.

> Or could we specify that only `LongStartSequence` and `ShortStartSequence` are supported for the value of this parameter?

Yes, I think you can just throw an `invalid_argument` exception if another value is specified so you don't have to implement `Length`. I'm not sure it would be useful apart maybe for some niche use cases.

src/h265rtpdepacketizer.cpp
```diff
@@ -0,0 +1,142 @@
...
+	auto accessUnit = binary{};
+	auto frameInfo = std::make_shared<FrameInfo>(timestamp);
+	auto nFrags = 0;
+
+	for (auto it = begin; it != end; ++it) {
+		auto pkt = it->get();
+		auto pktParsed = reinterpret_cast<const rtc::RtpHeader *>(pkt->data());
+		auto rtpHeaderSize = pktParsed->getSize() + pktParsed->getExtensionHeaderSize();
+		auto nalUnitHeader =
+		    H265NalUnitHeader{std::to_integer<uint8_t>(pkt->at(rtpHeaderSize)),
```
paullouisageneau, line 35: This assumes that there is something else than padding in the packet, and it breaks if the packet is empty. See the corresponding fix for `H264RtpDepacketizer` in https://github.com/paullouisageneau/libdatachannel/pull/1140.

edmonds: Thanks, I should have some time this weekend to fix that and implement the other change to make the start code configurable.

edmonds: c14469432a6405d6f9c146714c490c69587844f4

src/h265rtpdepacketizer.cpp
edmonds: ee96c7ade195ed0909c2909f0e843d13b8500adf

edmonds:
Re-based on master, added fix for empty RTP packet payloads, made the start sequence configurable.

paullouisageneau (approved):
Thanks, it looks good, feel free to add you name in new files headers.

edmonds:
> Thanks, it looks good, feel free to add you name in new files headers.

Done.

paullouisageneau:
@edmonds Sorry, you need to make a couple changes after rebasing as [FrameInfo now requires the payload type](https://github.com/paullouisageneau/libdatachannel/actions/runs/8723442203/job/23931675842?pr=1134#step:6:99) after https://github.com/paullouisageneau/libdatachannel/pull/1156 was merged.

edmonds:
OK, no problem. Should be able to take a look in a couple days.

edmonds:
Sorry for the delay getting back to this!

I rebased on master, added the missing `<algorithm>` include, made the FrameInfo changes (due to #1156), and made a fix similar to #1167 to handle multiple NALUs.

paullouisageneau (approved):
Great, thank you for the work!

PR #1135, opened 2024-03-19
H264RtpDepacketizer: De-packetize access units rather than individual NALUs

edmonds:
This commit updates the `H264RtpDepacketizer` to accumulate the NALUs for a particular RTP timestamp into a single output message, rather than returning each NALU as an individual message. This helps decoders which may want to see the non-VCL SPS/PPS/etc. NALUs in the same access unit as a VCL NALU rather than as standalone messages.

Each NALU in the access unit buffer is prepended with an H.264 Annex B start code 0x00, 0x00, 0x01.

### DIFF
--- a/src/h264rtpdepacketizer.cpp
+++ b/src/h264rtpdepacketizer.cpp
@@ -10,94 +10,103 @@
 
 #include "h264rtpdepacketizer.hpp"
 #include "nalunit.hpp"
-#include "track.hpp"
 
-#include "impl/logcounter.hpp"
-
-#include <cmath>
-#include <utility>
-
-#ifdef _WIN32
-#include <winsock2.h>
-#else
-#include <arpa/inet.h>
-#endif
+#include "impl/internals.hpp"
 
 namespace rtc {
 
-const unsigned long stapaHeaderSize = 1;
-const auto fuaHeaderSize = 2;
+const binary naluLongStartCode = {byte{0}, byte{0}, byte{0}, byte{1}};
+const binary naluShortStartCode = {byte{0}, byte{0}, byte{1}};
 
 const uint8_t naluTypeSTAPA = 24;
 const uint8_t naluTypeFUA = 28;
 
-message_vector H264RtpDepacketizer::buildFrames(message_vector::iterator begin,
-                                                message_vector::iterator end, uint8_t payloadType, uint32_t timestamp) {
-	message_vector out = {};
-	auto fua_buffer = std::vector<std::byte>{};
-	auto frameInfo = std::make_shared<FrameInfo>(payloadType, timestamp);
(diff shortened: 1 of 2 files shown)


edmonds:
@Sean-Der This should implement the change to the H264 de-packetizer you were asking for in https://github.com/paullouisageneau/libdatachannel/pull/1134#issuecomment-2003706014. I tested it with the Intel VPL encoder/decoder and packets produced with `rtc::H264RtpPacketizer` and it worked for me.

Sean-Der:
@paullouisageneau what do you think of this change? I am in big support of it!

paullouisageneau:
> @paullouisageneau what do you think of this change? I am in big support of it!

I think this is great, the change totally makes sense.

src/h264rtpdepacketizer.cpp
```diff
@@ -10,85 +10,76 @@
...
-#else
-#include <arpa/inet.h>
-#endif
+#include "impl/internals.hpp"
 
 namespace rtc {
 
-const unsigned long stapaHeaderSize = 1;
-const auto fuaHeaderSize = 2;
+const binary naluStartCode = {byte{0}, byte{0}, byte{1}};
```
paullouisageneau, line 18: What is the rationale for using the 3-byte start code here instead of the 4-byte one?

edmonds: I was wondering if there were any compatibility issues with the 3 vs. 4 byte start sequence, so I googled it and the 3 byte start sequence seemed to be more common for H264, at least on the pages that I looked at :-)

If there are no real compatibility concerns it seems like it would be better to make the H264/H265 de-packetizers consistent.

edmonds: Made it configurable in 27f238fc3c3513cdbbe63fde64beeb18ccf5a326 like https://github.com/paullouisageneau/libdatachannel/pull/1134.

edmonds:
Re-based on master, made the start sequence configurable.

paullouisageneau (approved):
It looks good, thanks!

paullouisageneau:
@edmonds Could you please address conflicts here too?

Sean-Der:
@edmonds I would love to fix these merge conflicts for you! If you add me as a contributor to your fork I can fix them right now :)

edmonds:
> @edmonds I would love to fix these merge conflicts for you! If you add me as a contributor to your fork I can fix them right now :)

OK, just sent you an invite.

Sean-Der:
@paullouisageneau done! can I get a review please?

@edmonds I changed the name of the member variable from `seperator` -> `mSeperator` and ran clang-format, beyond that no changes!

paullouisageneau (approved):
It looks good, thank you.

PR #1144, opened 2024-03-28
Preserve codecs priority

wusspuss:


src/description.cpp
```diff
@@ -885,7 +885,9 @@ void Description::Application::parseSdpLine(string_view line) {
 Description::Media::Media(const string &sdp) : Entry(get_first_line(sdp), "", Direction::Unknown) {
 	string line;
 	std::istringstream ss(sdp);
-	std::getline(ss, line); // discard first line
+	std::getline(ss, line); 
+	parseFirstLine(line); // save codecs order
```
paullouisageneau, line 889: This method parses the first line a second time, which is not ideal.

src/description.cpp
```diff
@@ -1051,6 +1054,17 @@ string Description::Media::generateSdpLines(string_view eol) const {
 	return sdp.str();
 }
 
+void Description::Media::parseFirstLine(string line) {
+  std::istringstream ss(line);
+  string word;
+  std::getline(ss, word, ' '); // audio
+  std::getline(ss, word, ' '); // 9
+  std::getline(ss, word, ' '); // UDP/TLS/RTP/SAVPF
```
paullouisageneau, line 1062: The logic here is different from the one in `Entry()`, meaning the behavior can be inconsistent, for instance if there are two consecutive spaces.

paullouisageneau (commented):
The fix makes perfect sense but the implementation does not work as-is. For instance, when RTP maps are added or removed with `addRtpMap()` or `removeRtpMap()` you must update the ordered payload types, otherwise descriptions created this way are broken. For consistency, you should remove the payload types that are not in the map after parsing the media description (statically-assigned codecs are not supported), and `payloadTypes()` should also return the ordered payload types.

paullouisageneau:
Any news on this?

paullouisageneau:
Closing due to inactivity.

PR #1152, opened 2024-04-01
Added playout delay extension support

sbarrac:
Adds support for the [playout delay extension](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/playout-delay)

### DIFF
--- a/src/rtppacketizer.cpp
+++ b/src/rtppacketizer.cpp
@@ -34,7 +34,7 @@ message_ptr RtpPacketizer::packetize(shared_ptr<binary> payload, bool mark) {
 	const bool setPlayoutDelay = (rtpConfig->playoutDelayId > 0 && rtpConfig->playoutDelayId < 15);
 
 	if (setPlayoutDelay)
-		rtpExtHeaderSize += 1;
+		rtpExtHeaderSize += 4;
 
 	if (rtpConfig->mid.has_value())
 		rtpExtHeaderSize += (1 + rtpConfig->mid->length());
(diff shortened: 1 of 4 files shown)


sbarrac:
Just thought I'd add some context.

This extension is really useful for cloud gaming scenarios. I was having an issue where the browser (chrome) would try to make the video smoother by delaying the rendering of a H264 stream. Even with ~10ms latency and zero packet loss, it was delaying the rendering by ~40MS, which is a huge delay in the context of playing a streamed game. 

The playout delay extension with minimum & maximum playout set to zero was the only way I found to remove this delay.

src/rtppacketizer.cpp
```diff
@@ -85,6 +90,21 @@ message_ptr RtpPacketizer::packetize(shared_ptr<binary> payload, bool mark) {
...
+			uint16_t max = rtpConfig->playoutDelayMax & 0xFFF;
+
+			// 12 bits for min + 12 bits for max
+			char data[] = {
+				static_cast<char>((min >> 4) & 0xFF), 
+				static_cast<char>(((min & 0xF) << 4) | ((max >> 8) & 0xF)),
+				static_cast<char>(max & 0xFF)
+			};
+
+			extHeader->writeOneByteHeader(offset, rtpConfig->playoutDelayId, (byte *)data, 3);
```
paullouisageneau, line 105: You should use `byte` directly so you don't need casts:
```suggestion
    byte data[] = {
				byte((min >> 4) & 0xFF), 
			    byte(((min & 0xF) << 4) | ((max >> 8) & 0xF)),
				byte(max & 0xFF)
			};

			extHeader->writeOneByteHeader(offset, rtpConfig->playoutDelayId, data, 3);
```

paullouisageneau (commented):
It looks good, I'm only concerned about the header size in the packetizer.

sbarrac:
I think this is ready.

I've tested with different min/max values and it's working as expected. Debug output also shows no issues.

paullouisageneau (approved):
It looks good, thank you!

PR #1154, opened 2024-04-08
Fix MbedTLS usage bugs and allow cmake to use imported mbedtls library

dmllr:
# Changes

## Fixes
* in `examples/streamer/h264fileparser.cpp` replaced `vector<byte>` to `vector<std::byte>` to avoid ambiguity
* in `src/impl/tls.cpp` added `MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET` as allowed return codes to support TLS1.3
* in `src/impl/tlstransport.cpp` added call to `psa_crypto_init()` to allow PSA Cryptographic operations that are actually used
* in `src/impl/verifiedtlstransport.cpp` function `mbedtls_x509_crt_parse` assumes that a certificate in PEM format contains tailing 0, so it should be counted when passing string size, changed `cacert->size()` to `cacert->size() + 1`

## Features
`find_library` in `CMake` always tries to find the library installed, but in the case when the same library was build from source as a CMake dependency, find_library would fail. To fix the issue, `CMakeLists.txt` is changed to call `find_library` for `mbedtls` and `srtp2` only if related targets are not exist.


paullouisageneau (approved):
Thanks for the fixes!

PR #1156, opened 2024-04-10
Include payloadType in FrameInfo

singpolyma:
So that consumers of onFrame can tell what codec the frame is in and know what decoder to use.

### DIFF
--- a/include/rtc/frameinfo.hpp
+++ b/include/rtc/frameinfo.hpp
@@ -14,7 +14,8 @@
 namespace rtc {
 
 struct RTC_CPP_EXPORT FrameInfo {
-	FrameInfo(uint32_t timestamp) : timestamp(timestamp){};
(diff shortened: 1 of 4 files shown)


paullouisageneau (approved):
It looks OK but examining the payload type after depacketization only works for audio because packetization is typically the same irrelevant of the codec. The trick does not work for video.

singpolyma:
I'm only using it for audio right now, but I'm not sure why it would be different for video? It's still going to be RtpHeader + payload, no? Or is video packets totally different RtpHeader sizes etc?

paullouisageneau:
> I'm only using it for audio right now, but I'm not sure why it would be different for video? It's still going to be RtpHeader + payload, no? Or is video packets totally different RtpHeader sizes etc?

It's fundamentally different for video. Of course RTP headers are the same, but a single video frame may be packetized over several RTP packets, and the method to fragment and reconstruct a frame requires parsing payloads which is codec-dependent.

PR #1157, opened 2024-04-10
Fix misaligned dereference in H264 and H265 packetizers

ray1422:
Use memcpy instead of direct dereference to avoid
misaligned access.

### DIFF
--- a/examples/streamer/h264fileparser.cpp
+++ b/examples/streamer/h264fileparser.cpp
@@ -11,6 +11,7 @@
 #include "rtc/rtc.hpp"
 
 #include <fstream>
+#include <cstring>
 
 #ifdef _WIN32
 #include <winsock2.h>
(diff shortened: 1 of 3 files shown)


paullouisageneau (commented):
Thank you for the fix!

There is the exact same issue in `H265RtpPacketizer`, would you mind applying the same fix there too please?
https://github.com/paullouisageneau/libdatachannel/blob/5064b0bf7f095e15848a028d37633890fc6196a1/src/h265rtppacketizer.cpp#L36

ray1422:
Fixed

ray1422:
> Thank you for the fix!
> 
> There is the exact same issue in `H265RtpPacketizer`, would you mind applying the same fix there too please?
> 
> https://github.com/paullouisageneau/libdatachannel/blob/5064b0bf7f095e15848a028d37633890fc6196a1/src/h265rtppacketizer.cpp#L36

Fixed

paullouisageneau (approved):

PR #1171, opened 2024-05-10
Add option to disable fingerprint validation

xicilion:
In Peer B of the WebRTC Direct protocol, the certificate fingerprint of Peer A needs to be ignored in order to establish a handshake with Peer A when the real sdp is not known.

### DIFF
--- a/include/rtc/configuration.hpp
+++ b/include/rtc/configuration.hpp
@@ -77,7 +77,7 @@ struct RTC_CPP_EXPORT Configuration {
 	bool disableAutoNegotiation = false;
 	bool disableAutoGathering = false;
 	bool forceMediaTransport = false;
-	bool skipCheckFingerprint = false;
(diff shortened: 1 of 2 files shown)


include/rtc/configuration.hpp
```diff
@@ -77,6 +77,7 @@ struct RTC_CPP_EXPORT Configuration {
 	bool disableAutoNegotiation = false;
 	bool disableAutoGathering = false;
 	bool forceMediaTransport = false;
+	bool skipCheckFingerprint = false;
```
paullouisageneau, line 80: You could call the flag `disableFingerprintVerification` for consistency with `WebSocketConfiguration`, otherwise it looks good.

xicilion: make sense.

paullouisageneau (approved):
It looks good, thank you.

PR #1172, opened 2024-05-10
Remove checking remote description has different ICE credentials

xicilion:
In the WebRTC Direct protocol, we need to allow local and remote to use the same ufrag and pwd.

### DIFF
--- a/include/rtc/description.hpp
+++ b/include/rtc/description.hpp
@@ -59,7 +59,6 @@ public:
 	string typeString() const;
 	Role role() const;
 	string bundleMid() const;
-	string sessionId() const;
 	std::vector<string> iceOptions() const;
 	optional<string> iceUfrag() const;
 	optional<string> icePwd() const;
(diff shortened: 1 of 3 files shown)


paullouisageneau:
I get the idea, but this disables a useful check (it's common for users to mess up signaling and send back the description into the same peer connection). I wonder if there is a way to check for such a scenario that doesn't rely on ICE attributes.

xicilion:
Or we can leave that choice to the developers and provide an option to disable it.

xicilion:
mSessionId is randomly generated each time, the probability of duplication is very small, is it possible to check if mSessionId is equal?

xicilion:
@paullouisageneau how about check mSessionId?

paullouisageneau:
It makes sense, the problem is that while session id is currently randomly generated in libdatachannel, it is not guaranteed to be unique between endpoints (other implementations typically use a timestamp), so it feels a bit clumsy. Since I've no better idea, I think you can remove the check.

paullouisageneau (approved):

PR #1198, opened 2024-05-29
Fix a bug where packets retransmitted for NACK are encrypted with SRTP twice

sile:
Let me report a NACK handling bug and submit a PR to fix it.

I used OBS Studio (30.1.2) WHIP to publish a video stream to a WebRTC SFU server. 
When the network became unstable and NACK packets were sent from the server, the video stream appeared corrupted on the viewer’s side. 
After some debugging, it was discovered that the packet binaries stored in `RtcpNackResponder` were unintentionally modified by the `srtp_protect()` function when the messages were first sent. As a result, when the packets were re-sent to the server for NACK, they were encrypted twice.

This PR fixes the issue by deep copying messages before they are stored in `RtcpNackResponder`.

### DIFF
--- a/src/rtcpnackresponder.cpp
+++ b/src/rtcpnackresponder.cpp
@@ -79,9 +79,13 @@ optional<binary_ptr> RtcpNackResponder::Storage::get(uint16_t sequenceNumber) {
 	                                 : nullopt;
 }
 
-void RtcpNackResponder::Storage::store(binary_ptr packet) {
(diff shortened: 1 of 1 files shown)


paullouisageneau:
Thank you for submitting a PR, but this is already fixed by https://github.com/paullouisageneau/libdatachannel/pull/1118

sile:
Oops, sorry for missing the PR.

PR #1204, opened 2024-06-07
feat: update current remote description with fingerprint

### LINKED ISSUE #1203: Update the remote description after SDP negotiation?
This is related to #1166

For libp2p WebRTC-Direct after the DTLS handshake has taken place and the connection becomes ready to use, we open a datachannel and use it to [verify](https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md#connection-security) the local and remote certificate fingerprints and the peer identifier so we can assert that the remote peer is the peer we think they are.

During connection we set the remote certificate fingerprint to a dummy value and disable fingerprint validation, since we are going to do it later.

When we come to do it we need access to the fingerprint of the remote certificate.

The browser `RTCPeerConnection` has a [.currentRemoteDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/currentRemoteDescription) property - this returns:

> an object describing the remote end of the connection as it was most recently successfully negotiated

and includes the fingerprint of the remote connection.

The node.js polyfill [returns the remoteDescription](https://github.com/murat-dogan/node-datachannel/blob/master/polyfill/RTCPeerConnection.js#L168) when this property is accessed, the same as `.remoteDescription` - this is not quite right as the `.remoteDescription` property should return what was passed to [.setRemoteDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setRemoteDescription) - this can be different to the currently negotiated state of remote end of the connection.

The Description class [parses the fingerprint](https://github.com/paullouisageneau/libdatachannel/blob/master/src/description.cpp#L108-L144) out of the description passed to `.setRemoteDescription`, later on that fingerprint is [verified or not, depending on config](https://github.com/paullouisageneau/libdatachannel/blob/master/src/impl/peerconnection.cpp#L447-L448) - would it be possible to have a copy of the local Description that gets updated with the actual fingerprint the connection uses?

If this is done along with the final ICE candidates etc it would make it possible to implement `.currentRemoteDescription` on the polyfill correctly and also for me to implement the final piece of the WebRTC-Direct connection listener.

I can try to implement this if it's acceptable though it's really stretching my C++!

achingbrain:
Adds a `currentRemoteDescription` field to `PeerConnection` which can be updated independently of `remoteDescription`, the idea being the `remoteDescription` is the offer received from the remote peer and the `currentRemoteDescription` is what has actually been negotiated with them.

The `currentRemoteDescription` is initialised when the `remoteDescription` is set.

Currently the `checkFingerprint` method updates the certificate fingerprint used by the connection to the remote peer but I guess other modifications may be useful.

Exposes the `currentRemoteDescription` via a getter similar to `remoteDescription`.

Closes #1203
Refs #1166

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -278,6 +278,24 @@ Return value: the length of the string copied in buffer (including the terminati
 
 If `buffer` is `NULL`, the description is not copied but the size is still returned.
 
+#### rtcGetCurrentRemoteDescription
+
+```
+int rtcGetCurrentRemoteDescription(int pc, char *buffer, int size)
+```
+
+Retrieves the currently negotiated remote description in SDP format.
+
+Arguments:
+
+- `pc`: the Peer Connection identifier
+- `buffer`: a user-supplied buffer to store the description
+- `size`: the size of `buffer`
+
+Return value: the length of the string copied in buffer (including the terminating null character) or a negative error code
+
+If `buffer` is `NULL`, the description is not copied but the size is still returned.
+
 #### rtcGetLocalDescriptionType
 
 ```
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -84,6 +84,7 @@ public:
 	bool hasMedia() const;
 	optional<Description> localDescription() const;
 	optional<Description> remoteDescription() const;
+	optional<Description> currentRemoteDescription() const;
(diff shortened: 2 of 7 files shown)


achingbrain:
Superseded by #1206

PR #1206, opened 2024-06-11
feat: add remoteFingerprints method to PeerConnection

### LINKED ISSUE #1203: Update the remote description after SDP negotiation?

achingbrain:
Returns a vector that contains the certificate fingerprints used by the connection to the remote peer.

Fingerprints are deduplicated before adding to the vector as the verifyFingerprint method can be invoked multiple times during the connection lifecycle.

Closes #1203
Refs #1166

### DIFF
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -35,6 +35,11 @@ struct RTC_CPP_EXPORT DataChannelInit {
 	string protocol = "";
 };
 
+struct RTC_CPP_EXPORT RemoteFingerprint {
+	string value;
+	CertificateFingerprint::Algorithm algorithm;
+};
+
 class RTC_CPP_EXPORT PeerConnection final : CheshireCat<impl::PeerConnection> {
 public:
 	enum class State : int {
@@ -113,6 +118,7 @@ public:
 	void onSignalingStateChange(std::function<void(SignalingState state)> callback);
 
 	void resetCallbacks();
+	std::vector<struct RemoteFingerprint> remoteFingerprints();
 
 	// Stats
 	void clearStats();
--- a/include/rtc/peerconnection.hpp
+++ b/include/rtc/peerconnection.hpp
@@ -35,11 +35,6 @@ struct RTC_CPP_EXPORT DataChannelInit {
 	string protocol = "";
 };
 
-struct RTC_CPP_EXPORT RemoteFingerprint {
-	string value;
-	CertificateFingerprint::Algorithm algorithm;
-};
-
 class RTC_CPP_EXPORT PeerConnection final : CheshireCat<impl::PeerConnection> {
 public:
 	enum class State : int {
@@ -118,7 +113,7 @@ public:
 	void onSignalingStateChange(std::function<void(SignalingState state)> callback);
 
 	void resetCallbacks();
-	std::vector<struct RemoteFingerprint> remoteFingerprints();
+	CertificateFingerprint remoteFingerprint();
 
 	// Stats
 	void clearStats();
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -239,7 +239,7 @@ shared_ptr<DtlsTransport> PeerConnection::initDtlsTransport() {
 			throw std::logic_error("No underlying ICE transport for DTLS transport");
 
 		auto certificate = mCertificate.get();
-		auto verifierCallback = weak_bind(&PeerConnection::checkFingerprint, this, _1);
+		auto verifierCallback = weak_bind(&PeerConnection::checkFingerprint, this, _1, fingerprintAlgorithm);
 		auto dtlsStateChangeCallback =
 		    [this, weak_this = weak_from_this()](DtlsTransport::State transportState) {
 			    auto shared_this = weak_this.lock();
@@ -439,22 +439,34 @@ void PeerConnection::rollbackLocalDescription() {
 	}
 }
 
-bool PeerConnection::checkFingerprint(const std::string &fingerprint) const {
+bool PeerConnection::checkFingerprint(const std::string &fingerprint, const CertificateFingerprint::Algorithm &algorithm) {
 	std::lock_guard lock(mRemoteDescriptionMutex);
 	if (!mRemoteDescription || !mRemoteDescription->fingerprint())
 		return false;
 
-	if (config.disableFingerprintVerification)
-		return true;
-
 	auto expectedFingerprint = mRemoteDescription->fingerprint()->value;
-	if (expectedFingerprint  == fingerprint) {
+	if (config.disableFingerprintVerification || expectedFingerprint == fingerprint) {
 		PLOG_VERBOSE << "Valid fingerprint \"" << fingerprint << "\"";
+		storeRemoteFingerprint(fingerprint, algorithm);
 		return true;
 	}
 
 	PLOG_ERROR << "Invalid fingerprint \"" << fingerprint << "\", expected \"" << expectedFingerprint << "\"";
 	return false;
+}
+
+void PeerConnection::storeRemoteFingerprint(const std::string &value, const CertificateFingerprint::Algorithm &algorithm) {
+	auto iter = std::find_if(rFingerprints.begin(), rFingerprints.end(), [&](const RemoteFingerprint& existing){return existing.value == value;});
+  bool seenPreviously = iter != rFingerprints.end();
+
+	if (seenPreviously) {
+		return;
+	}
+
+	rFingerprints.push_back({
+		value,
+		algorithm
+	});
 }
 
 void PeerConnection::forwardMessage(message_ptr message) {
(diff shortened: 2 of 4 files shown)


src/impl/peerconnection.cpp
```diff
@@ -1301,6 +1313,13 @@ void PeerConnection::resetCallbacks() {
 	trackCallback = nullptr;
 }
 
+std::vector<struct RemoteFingerprint> PeerConnection::remoteFingerprints() {
+	std::vector<struct RemoteFingerprint> ret;
+	ret = rFingerprints;
```
achingbrain, line 1318: I'm trying to return a copy of the vector here, I'm not sure if this is sufficient or even necessary?

include/rtc/peerconnection.hpp
```diff
@@ -35,6 +35,11 @@ struct RTC_CPP_EXPORT DataChannelInit {
 	string protocol = "";
 };
 
+struct RTC_CPP_EXPORT RemoteFingerprint {
+	string value;
+	CertificateFingerprint::Algorithm algorithm;
+};
```
paullouisageneau, line 41: `struct CertificateFingerprint` implements [exactly this](https://github.com/paullouisageneau/libdatachannel/blob/eec049f6ddb3d2937fca8fe21443a87ba40e5225/include/rtc/description.hpp#L31), you shouldn't create a new struct.

src/impl/peerconnection.cpp
```diff
@@ -439,24 +439,36 @@ void PeerConnection::rollbackLocalDescription() {
...
+
+	if (seenPreviously) {
+		return;
+	}
+
+	rFingerprints.push_back({
+		value,
+		algorithm
+	});
+}
```
paullouisageneau, line 470: This is not needed as the verifier callback will be called for a single algorithm.

src/impl/peerconnection.hpp
```diff
@@ -157,6 +159,8 @@ struct PeerConnection : std::enable_shared_from_this<PeerConnection> {
 
 	Queue<shared_ptr<DataChannel>> mPendingDataChannels;
 	Queue<shared_ptr<Track>> mPendingTracks;
+
+	std::vector<struct RemoteFingerprint> rFingerprints;
```
paullouisageneau, line 163: Private member variables should be prefixed by `m`. Actually, since a single fingerprint is passed to the verifier callback, you can store only the one fingerprint:
```suggestion
    CertificateFingerprint::Algorithm mRemoteFingerprintAlgorithm = CertificateFingerprint::Algorithm::Sha256;
	optional<string> mRemoteFingerprint;
```

src/impl/peerconnection.cpp
```diff
@@ -439,24 +439,36 @@ void PeerConnection::rollbackLocalDescription() {
 	}
 }
 
-bool PeerConnection::checkFingerprint(const std::string &fingerprint) const {
+bool PeerConnection::checkFingerprint(const std::string &fingerprint, const CertificateFingerprint::Algorithm &algorithm) {
```
paullouisageneau, line 442: You don't need to add the fingerprint to the callback, you can simply set the member variable `mRemoteFingerprintAlgorithm` in `initDtlsTransport` instead of the local one `fingerprintAlgorithm`.

src/impl/peerconnection.cpp
```diff
@@ -439,24 +439,36 @@ void PeerConnection::rollbackLocalDescription() {
...
 		return false;
 
-	if (config.disableFingerprintVerification)
-		return true;
-
 	auto expectedFingerprint = mRemoteDescription->fingerprint()->value;
-	if (expectedFingerprint  == fingerprint) {
+	if (config.disableFingerprintVerification || expectedFingerprint == fingerprint) {
 		PLOG_VERBOSE << "Valid fingerprint \"" << fingerprint << "\"";
+		storeRemoteFingerprint(fingerprint, algorithm);
```
paullouisageneau, line 450:
```suggestion
		mRemoteFingerprint = fingerprint;
```

include/rtc/peerconnection.hpp
```diff
@@ -113,6 +118,7 @@ class RTC_CPP_EXPORT PeerConnection final : CheshireCat<impl::PeerConnection> {
 	void onSignalingStateChange(std::function<void(SignalingState state)> callback);
 
 	void resetCallbacks();
+	std::vector<struct RemoteFingerprint> remoteFingerprints();
```
paullouisageneau, line 121:
```suggestion
	std::vector<CertificateFingerprint> remoteFingerprints();
```

src/peerconnection.cpp
```diff
@@ -367,6 +367,10 @@ optional<std::chrono::milliseconds> PeerConnection::rtt() {
 	return sctpTransport ? sctpTransport->rtt() : nullopt;
 }
 
+std::vector<struct RemoteFingerprint> PeerConnection::remoteFingerprints() {
+	return impl()->remoteFingerprints();
+}
```
paullouisageneau, line 372:
```suggestion
std::vector<CertificateFingerprint> PeerConnection::remoteFingerprints() {
    if (mRemoteFingerprint)
	    return {CertificateFingerprint{mRemoteFingerprintAlgorithm, *mRemoteFingerprint}};
	else
	    return {};
}
```

src/impl/peerconnection.cpp
achingbrain: That's interesting, I was getting duplicate fingerprints as the callback was being invoked multiple times.

include/rtc/peerconnection.hpp
achingbrain: Given there's only every going to be one of these we should probably just make this:

```c++
CertificateFingerprint remoteFingerprint();
```

src/impl/peerconnection.cpp
paullouisageneau: It's called only once in the certificate callback of the cryptographic library, and always uses the fingerprint algorithm that was passed on DTLS transport initialization.

src/impl/peerconnection.cpp
```diff
@@ -439,24 +439,36 @@ void PeerConnection::rollbackLocalDescription() {
...
 	if (!mRemoteDescription || !mRemoteDescription->fingerprint())
 		return false;
 
-	if (config.disableFingerprintVerification)
-		return true;
-
 	auto expectedFingerprint = mRemoteDescription->fingerprint()->value;
-	if (expectedFingerprint  == fingerprint) {
+	if (config.disableFingerprintVerification || expectedFingerprint == fingerprint) {
 		PLOG_VERBOSE << "Valid fingerprint \"" << fingerprint << "\"";
```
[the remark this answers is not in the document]
achingbrain: Done!

include/rtc/peerconnection.hpp
achingbrain: Done!

src/impl/peerconnection.hpp
achingbrain: Done!

include/rtc/peerconnection.hpp
achingbrain: Done!

src/peerconnection.cpp
achingbrain: I think this wanted to be in `/src/impl/peerconnection.cpp` as that's where `mRemoteFingerprint` is declared?

src/impl/peerconnection.cpp
achingbrain: Done!

paullouisageneau (approved):
It looks good, thank you!

PR #1211, opened 2024-06-14
feat: add listener for unhandled STUN requests with ICE UDP mux

achingbrain:
Calls the functions added to libjuice in https://github.com/paullouisageneau/libjuice/pull/248 (this needs to be merged & released before CI will pass here).

Exports a `OnUnhandledStunRequest` function that can be passed a callback that will be invoked when an incoming STUN message is received that has no corresponding agent for the ICE ufrag.

Refs #1166

### DIFF
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -35,15 +35,20 @@ RTC_CPP_EXPORT void Preload();
 RTC_CPP_EXPORT std::shared_future<void> Cleanup();
 
 struct UnhandledStunRequest {
-	optional<std::string> localUfrag;
-	optional<std::string> remoteUfrag;
-	std::string address;
-	uint16_t port;
+	std::string localUfrag;
+	std::string remoteUfrag;
+	std::string remoteHost;
+	uint16_t remotePort;
 };
 
-RTC_CPP_EXPORT typedef std::function<void(UnhandledStunRequest request)> UnhandledStunRequestCallback;
+RTC_CPP_EXPORT typedef std::function<void(UnhandledStunRequest request, void* userPtr)> UnhandledStunRequestCallback;
 
-RTC_CPP_EXPORT void OnUnhandledStunRequest(std::string host, int port, UnhandledStunRequestCallback callback = nullptr);
+struct UnhandledStunRequestHandler {
+	UnhandledStunRequestCallback callback;
+	void *userPtr;
+};
+
+RTC_CPP_EXPORT void OnUnhandledStunRequest(std::string host, int port, UnhandledStunRequestCallback callback = nullptr, void *userPtr = nullptr);
 
 struct SctpSettings {
 	// For the following settings, not set means optimized default
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -34,17 +34,6 @@ RTC_CPP_EXPORT void InitLogger(LogLevel level, LogCallback callback = nullptr);
 RTC_CPP_EXPORT void Preload();
 RTC_CPP_EXPORT std::shared_future<void> Cleanup();
 
-struct IceUdpMuxRequest {
-	std::string localUfrag;
-	std::string remoteUfrag;
-	std::string remoteHost;
-	uint16_t remotePort;
-};
-
-RTC_CPP_EXPORT typedef std::function<void(IceUdpMuxRequest request)> IceUdpMuxCallback;
-
-RTC_CPP_EXPORT void ListenIceUdpMux (int port, IceUdpMuxCallback *callback, std::optional<std::string> bindAddress = nullptr);
-
 struct SctpSettings {
 	// For the following settings, not set means optimized default
 	optional<size_t> recvBufferSize;                // in bytes
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -34,21 +34,16 @@ RTC_CPP_EXPORT void InitLogger(LogLevel level, LogCallback callback = nullptr);
 RTC_CPP_EXPORT void Preload();
 RTC_CPP_EXPORT std::shared_future<void> Cleanup();
 
-struct UnhandledStunRequest {
+struct IceUdpMuxRequest {
 	std::string localUfrag;
 	std::string remoteUfrag;
 	std::string remoteHost;
 	uint16_t remotePort;
 };
 
-RTC_CPP_EXPORT typedef std::function<void(UnhandledStunRequest request, void* userPtr)> UnhandledStunRequestCallback;
+RTC_CPP_EXPORT typedef std::function<void(IceUdpMuxRequest request)> IceUdpMuxCallback;
 
-struct UnhandledStunRequestHandler {
-	UnhandledStunRequestCallback callback;
-	void *userPtr;
-};
-
-RTC_CPP_EXPORT void OnUnhandledStunRequest(std::string host, int port, UnhandledStunRequestCallback callback = nullptr, void *userPtr = nullptr);
+RTC_CPP_EXPORT void ListenIceUdpMux (int port, IceUdpMuxCallback *callback, std::optional<std::string> bindAddress = nullptr);
 
 struct SctpSettings {
 	// For the following settings, not set means optimized default
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -34,6 +34,17 @@ RTC_CPP_EXPORT void InitLogger(LogLevel level, LogCallback callback = nullptr);
 RTC_CPP_EXPORT void Preload();
 RTC_CPP_EXPORT std::shared_future<void> Cleanup();
 
+struct UnhandledStunRequest {
+	optional<std::string> localUfrag;
+	optional<std::string> remoteUfrag;
+	std::string address;
+	uint16_t port;
+};
+
+RTC_CPP_EXPORT typedef std::function<void(UnhandledStunRequest request)> UnhandledStunRequestCallback;
+
+RTC_CPP_EXPORT void OnUnhandledStunRequest(std::string host, int port, UnhandledStunRequestCallback callback = nullptr);
+
 struct SctpSettings {
 	// For the following settings, not set means optimized default
 	optional<size_t> recvBufferSize;                // in bytes
--- a/include/rtc/rtc.h
+++ b/include/rtc/rtc.h
@@ -172,6 +172,20 @@ typedef void(RTC_API *rtcAvailableCallbackFunc)(int id, void *ptr);
 typedef void(RTC_API *rtcPliHandlerCallbackFunc)(int tr, void *ptr);
 typedef void(RTC_API *rtcRembHandlerCallbackFunc)(int tr, unsigned int bitrate, void *ptr);
 
+// Handle STUN requests with unexpected ufrags
+
+typedef struct {
+	const char * ufrag;
+	const char * pwd;
+	uint8_t family;
+	const char * address;
+	uint16_t port;
+} rtcUnhandledStunRequest;
+
+typedef void(RTC_API *rtcUnhandledStunRequestCallbackFunc)(rtcUnhandledStunRequest request);
+
+RTC_C_EXPORT void rtcOnUnhandledStunRequest(const char *host, int port, rtcUnhandledStunRequestCallbackFunc callback);
+
 // Log
 
 // NULL cb on the first call will log to stdout
--- a/include/rtc/rtc.h
+++ b/include/rtc/rtc.h
@@ -172,20 +172,6 @@ typedef void(RTC_API *rtcAvailableCallbackFunc)(int id, void *ptr);
 typedef void(RTC_API *rtcPliHandlerCallbackFunc)(int tr, void *ptr);
 typedef void(RTC_API *rtcRembHandlerCallbackFunc)(int tr, unsigned int bitrate, void *ptr);
 
-// Handle STUN requests with unexpected ufrags
-
-typedef struct {
-	const char * ufrag;
-	const char * pwd;
-	uint8_t family;
-	const char * address;
-	uint16_t port;
-} rtcUnhandledStunRequest;
-
-typedef void(RTC_API *rtcUnhandledStunRequestCallbackFunc)(rtcUnhandledStunRequest request);
-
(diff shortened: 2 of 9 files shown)


achingbrain:
When implementing this I hadn't realised that libjuice will [only ever allow one UDP socket to be used](https://github.com/paullouisageneau/libjuice/issues/247) so unregistering the handler takes a host/port pair in order to unregister the handler only for that pair.

I hope this can be worked around as it's quite common to create two listeners during a test run and they'll not be able to co-exist in the same process like this.

We could, for example [invalidate the module cache](https://github.com/nodejs/help/issues/2806#issuecomment-2107762489) when loading the native part of `node-datachannel` to get a separate copy of `libdatachannel` that would run in it's own memory space but it's a bit of a hack.

Also if the application has already created a peer connection outside of libp2p's control, setting the host/port will have no effect as it would have already opened the port specified in the first invocation, and now we'd be setting the handler for that port which may not be what the user expects.

Anyway, if you'd prefer a separate function to remove the handler please shout.

It may not need the `void *user_ptr` arg to `juice_bind_stun` then since it can just invoke the global reference stored in `unboundStunCallback`.

xicilion:
> Also if the application has already created a peer connection outside of libp2p's control, setting the host/port will have no effect as it would have already opened the port specified in the first invocation, and now we'd be setting the handler for that port which may not be what the user expects.

This should not happen. In order to avoid such hidden errors, I added a check in `juice_bind_stun` to prevent binding when mux connections are detected, so as to avoid binding to ports that users do not expect.

xicilion:
The Mux mode is bound to a unique socket, which has no problem at all in a normal WebRTC application, after all, you can't even set the port in the browser.

But in the Peer B application of WebRTC Direct, it is restricted by this. One process can only bind to one port, which greatly limits its usage.

achingbrain:
Needs the changes in https://github.com/paullouisageneau/libjuice/pull/292 before this will pass CI.

src/global.cpp
```diff
@@ -88,6 +93,61 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
...
+			info->port
+		}, handler->userPtr);
+	} else {
+		PLOG_DEBUG << "No unhandled STUN request callback configured for port " << info->port;
+	}
+}
+
+#endif
+
+void OnUnhandledStunRequest ([[maybe_unused]] std::string host, [[maybe_unused]] int port, [[maybe_unused]] UnhandledStunRequestCallback callback, [[maybe_unused]] void *userPtr) {
```
paullouisageneau, line 119: The name should reflect that this listens only on mux mode, and that it attempts to listen to a specific port, you could rename it to `ListenIceUdpMux()` for instance, since the feature is called `enableIceUdpMux` in `rtc::Configuration`.

`host` should also be optional to allow any address (it would pass NULL to libjuice), and you could remame it to `bindAddress` as it is the name in the configuration. If it's not part of the mapping, I think it should be moved at the end and made an optional parameter.

src/global.cpp
```diff
@@ -88,6 +93,61 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
...
+	#if USE_NICE
+		PLOG_WARNING << "BindStunListener is not supported with libnice, please use libjuice";
+	#else
+	if (callback == NULL) {
+		PLOG_DEBUG << "Removing unhandled STUN request listener";
+
+		free(unboundStunCallbacks.at(port));
```
paullouisageneau, line 126: This causes a race condition: If the callback is called in parallel it will cause a use after free.

Also, if no callback is set for the port it will throw an `std::out_of_range` exception which is confusing for the user so you should use `find` instead.

include/rtc/global.hpp
```diff
@@ -34,6 +34,22 @@ RTC_CPP_EXPORT void InitLogger(LogLevel level, LogCallback callback = nullptr);
...
 RTC_CPP_EXPORT std::shared_future<void> Cleanup();
 
+struct UnhandledStunRequest {
+	std::string localUfrag;
+	std::string remoteUfrag;
+	std::string remoteHost;
+	uint16_t remotePort;
+};
+
+RTC_CPP_EXPORT typedef std::function<void(UnhandledStunRequest request, void* userPtr)> UnhandledStunRequestCallback;
```
paullouisageneau, line 44: An `std::function` should not have a user pointer as it has its own context. The user can use bind, lambda capture, or a callable object.

Additionally, all callbacks in the library should be implemented with `synchronized_callback` to prevent race conditions between calling and setting/resetting.

include/rtc/global.hpp
```diff
@@ -34,6 +34,22 @@ RTC_CPP_EXPORT void InitLogger(LogLevel level, LogCallback callback = nullptr);
...
+
+struct UnhandledStunRequestHandler {
+	UnhandledStunRequestCallback callback;
+	void *userPtr;
+};
```
paullouisageneau, line 49: I think you don't need this since the user pointer is useless in C++.

include/rtc/rtc.h
```diff
@@ -172,6 +172,20 @@ typedef void(RTC_API *rtcAvailableCallbackFunc)(int id, void *ptr);
...
+
+typedef struct {
+	const char * ufrag;
+	const char * pwd;
+	uint8_t family;
+	const char * address;
+	uint16_t port;
+} rtcUnhandledStunRequest;
+
+typedef void(RTC_API *rtcUnhandledStunRequestCallbackFunc)(rtcUnhandledStunRequest request);
```
paullouisageneau, line 185: On the contrary, the C callback should have a user pointer as there is no way to pass context otherwise. It should be stored by capturing it in the C++ callback.

src/global.cpp
```diff
@@ -88,6 +93,61 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
...
+		if (juice_mux_listen(host.c_str(), port, NULL, NULL) < 0) {
+			throw std::runtime_error("Could not unbind STUN listener");
+		}
+
+		return;
+	}
+
+	PLOG_DEBUG << "Adding listener for unhandled STUN requests";
+
+	UnhandledStunRequestHandler *handler = (UnhandledStunRequestHandler*)calloc(1, sizeof(UnhandledStunRequestHandler));
```
paullouisageneau, line 139: Please don't use C-style functions in C++. Here, you should use a `unique_ptr` or directly set the object as map value. However, this should not be necessary anymore if you remove the user pointer.

src/global.cpp
```diff
@@ -88,6 +93,61 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
...
+	handler->callback = std::move(callback);
+	handler->userPtr = userPtr;
+
+	unboundStunCallbacks[port] = handler;
```
paullouisageneau, line 143: There should be a mutex protecting the map otherwise the function is not thread-safe.

src/global.cpp
```diff
@@ -88,6 +93,61 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
 
 void SetSctpSettings(SctpSettings s) { impl::Init::Instance().setSctpSettings(std::move(s)); }
 
+std::map<int, UnhandledStunRequestHandler *> unboundStunCallbacks;
```
paullouisageneau, line 96: Should callbacks be specified by port or by bind address? Just asking because it needs to be consistent with libjuice. I think port makes sense, since the user is typically going to set the same `bindAddress` everytime, and the any address is quite tedious to handle.

achingbrain: The [libjuice PR](https://github.com/paullouisageneau/libjuice/pull/292) uses `${host}:${port}` as the key.

Since the `conn_mux.c` uses `udp_socket_config_t` to bind a UDP port, I wanted to make sure the user could explicitly bind both IPv4 and IPv6 addresses with:

```c
juice_mux_listen("0.0.0.0", port, &cb, NULL);
juice_mux_listen("::", port, &cb, NULL);
```

I could be misreading it but I think [these lines](https://github.com/paullouisageneau/libjuice/blob/master/src/udp.c#L161-L172) mean you need to pass one address family or the other.  For example passing `NULL` as the `bind_address` will not cause it to bind to both.

src/global.cpp
achingbrain: Fixed!

src/global.cpp
achingbrain: I've removed the map - `ListenIceUdpMux` now accepts a pointer to a callback and the caller of `ListenIceUdpMux` is responsible for it's memory management.

include/rtc/global.hpp
achingbrain: Fixed the user pointer part.

I'm not sure I follow about `synchronized_callback` - where it's used elsewhere here a callback has been passed in that's stored as a class member.  That member can get overwritten so I can see why we need to synchronize on it's access.

Here we're passing a reference to a function to `libjuice` and all state is passed along with the function invocation - do we still need `synchronized_callback`?

include/rtc/global.hpp
achingbrain: Fixed!

include/rtc/rtc.h
achingbrain: Hmm, I think this def might be left over from a previous attempt.  I've removed it.

src/global.cpp
achingbrain: Fixed!

src/global.cpp
achingbrain: I've removed the map.

src/global.cpp
achingbrain: > `host` should also be optional to allow any address (it would pass NULL to libjuice)

I guess the only problem with this is the invocation to unbind a listener with a host now looks a bit awkward, certainly in js:

```js
listenIceUdpMux(port, null, host)
```

How about passing a `bind_address` struct with `host`/`port` fields as the first argument instead?  That way the host field can be omitted if it should bind to all available addresses.

include/rtc/global.hpp
paullouisageneau: You are right that it's not needed if you store only a pointer to an external function object, since the concern is pushed to the user. However, I don't think it's a good idea, as stated in https://github.com/paullouisageneau/libdatachannel/pull/1211#discussion_r1922347908.

src/global.cpp
paullouisageneau: > Actually I'm not sure that's correct, passing `NULL` seems to bind all IPv4 and IPv6 interfaces.

Correct, `NULL` binds on IPv6 and IPv4. That's what you want in most cases.

src/global.cpp
paullouisageneau: > How about passing a `bind_address` struct with `host`/`port` fields as the first argument instead? That way the host field can be omitted if it should bind to all available addresses.

Good idea, it could be something like that, on the model of `WebSocketServerConfiguration`:
```
struct IceUdpMuxListenerConfiguration {
    uint16_t port;
    optional<string> bindAddress;
}
```

paullouisageneau:
I think letting the user manage the callback object is not idiomatic design. Instead, you should introduce an object to hold the state, let's say `IceUdpMuxListener`, which would have a `synchronized_callback` member. The would also decouple the callback from listening on the port. For instance:
```cpp
class IceUdpMuxListener {
    IceUdpMuxListener(uint16_t port, optional<string> bindAddress = nullopt); // calls juice_mux_listen
    ~IceUdpMuxListener(); // calls juice_mux_listen to unset
    void OnUnhandledStunRequest(std::function<void(IceUdpMuxRequest)>); // sets or unsets the synchronized_callback
    [...]
};
```
This way you don't have to manage the objects but it is still safe and usable from a user perspective. You can do it by looking at `WebSocketServer` for instance. I can help if necessary.

paullouisageneau:
Thank you for addressing the comments! However, I think letting the user manage the callback object is not idiomatic design. Instead, you should introduce an object to hold the state, let's say `IceUdpMuxListener`, which would have a `synchronized_callback` member. The would also decouple the callback from listening on the port. For instance:
```cpp
class IceUdpMuxListener {
    IceUdpMuxListener(uint16_t port, optional<string> bindAddress = nullopt); // calls juice_mux_listen
    ~IceUdpMuxListener(); // calls juice_mux_listen to unset
    void OnUnhandledStunRequest(std::function<void(IceUdpMuxRequest)>); // sets or unsets the synchronized_callback
    [...]
};
```
This way you still don't have to manage the objects but it is still safe and usable from a user perspective. You can do it by looking at `WebSocketServer` for instance. I can help if necessary.

achingbrain:
Thank you for your patience and all the feedback!

> I can help if necessary

Yes please!  I can take a stab at this but it seems like you have something specific in mind and I'm not very good with C++ so I don't feel like I'm going to get it right first time - please do feel free to push some commits here.

src/global.cpp
```diff
@@ -93,54 +88,7 @@ std::shared_future<void> Cleanup() { return impl::Init::Instance().cleanup(); }
...
-	PLOG_DEBUG << "Adding ICE UDP mux callback for port " << port;
-
-	if (juice_mux_listen(host, port, &InvokeIceUdpMuxCallback, callback) < 0) {
-		throw std::invalid_argument("Could not add ICE UDP mux callback");
-	}
-	#endif
-}
-
-RTC_CPP_EXPORT std::ostream &operator<<(std::ostream &out, LogLevel level) {
+std::ostream &operator<<(std::ostream &out, LogLevel level) {
```
achingbrain, line 91: The `RTC_CPP_EXPORT` has been removed here, not sure if that was intentional.

achingbrain:
Looks good to me 👍

src/impl/iceudpmuxlistener.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <juice/juice.h>
+#endif
+
+#include <atomic>
+
+namespace rtc::impl {
+
+struct IceUdpMuxListener final {
+	IceUdpMuxListener(uint16_t port, optional<string> bindAddress = nullopt);
+    ~IceUdpMuxListener();
```
achingbrain, line 27: Whitespace

```suggestion
	~IceUdpMuxListener();
```

achingbrain:
@paullouisageneau are there any further changes you'd like to see here or is this ready to go in?

paullouisageneau (approved):

PR #1214, opened 2024-06-17
add simple signaling-server implemented using libdatachannel cpp-api

Tim-S:
added simple signaling-server implemented using libdatachannel cpp-api


examples/signaling-server-libdatachannel-ws/signaling-server.cpp
```diff
@@ -0,0 +1,162 @@
...
+	config.enableTls = false;
+	config.certificatePemFile = std::nullopt;
+	config.keyPemFile = std::nullopt;
+	config.keyPemPass = std::nullopt;
+	config.bindAddress = std::nullopt;
+	config.connectionTimeout = std::nullopt;
+
+	// Check command line arguments.
+	for (int i = 1; i < argc; i++) {
+		if (strcmp(argv[i], "--help") == 0) {
```
paullouisageneau, line 53: You should not mix C functions in C++ when it is not necessary. Use `std::string`.

examples/signaling-server-libdatachannel-ws/signaling-server.cpp
```diff
@@ -0,0 +1,162 @@
...
+			const size_t len = strlen(argv[0]);
+			char *path = (char *)malloc(len + 1);
```
paullouisageneau, line 55: Same here, you should use `new[]`, or better way, `std::string`.

examples/signaling-server-libdatachannel-ws/signaling-server.cpp
```diff
@@ -0,0 +1,162 @@
...
+			strcpy(path, argv[0]);
+			path[len] = 0;
+
+			char *app_name = NULL;
+			// app_name = last_path_segment(path, "\\/");
+			fprintf(stderr,
```
paullouisageneau, line 61: Same here, you should use `std::cout`.

Sum1cares (approved):

Sum1cares (approved):

PR #1231, opened 2024-07-14
Disable the version.h update by default

### LINKED ISSUE #1227: Source directory requires read/write filesystem during build
Hi! I'm currently working in Java bindings of libdatachannel for a project of mine. The project will be cross-compiled using dockcross a for a set of architectures. To speed up the build process in these JNI-binding projects I build all target architectures in parallel. To ensure this is safe, I'm mounting the sources as read-only into the dockcross container and only the architecture-specific output directory is mounted read/write.

"Unfortunately" libdatachannel uses cmake's configure_file to render the version.h.in template into the source directory, which breaks the assumption "source directories won't be written during build, only output directories".

I don't have a ton of experience with cmake, but I see 2 general ways around this:

1. instead of having a header file with the 4 defines, the defines could be added using cmake's [add_definitions](https://cmake.org/cmake/help/latest/command/add_definitions.html) (or the newer alternatives)
2. A folder in the output directory could be setup and added to the includes, then the configure_file command could place the output there

I'd be willing to make a PR if either of the options are acceptable or if we come up with another solution that fits better. I'd also be fine if this is not something you want to do.

pschichtel:
closes #1227

### DIFF
--- a/.github/workflows/check-version.yml
+++ b/.github/workflows/check-version.yml
@@ -14,7 +14,7 @@ jobs:
     - name: submodules
       run: git submodule update --init --recursive --depth 1
     - name: cmake
-      run: cmake -B build -DUSE_GNUTLS=0 -DUSE_SYSTEM_SRTP=1 -DWARNINGS_AS_ERRORS=1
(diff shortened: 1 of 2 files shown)


pschichtel:
Not sure why that job would fail given my changes.

paullouisageneau (approved):
It looks god, that you!

paullouisageneau:
> Not sure why that job would fail given my changes.

A connection sometimes fail during the Windows CI tests, it need to look into that.

PR #1245, opened 2024-08-29
Fix SctpTransport::Instances leak

les-sosna:
SctpTransport::Instances was not being freed causing memory leak warnings


paullouisageneau (approved):
Looks good, thanks.

PR #1247, opened 2024-09-03
This callback doesn't do anything, and it's difficult to understand 

sbkyy:
This callback doesn't do anything, and it's difficult to understand


paullouisageneau:
This callback allows media handlers to send RTCP packets without going through the rest of the chain. It is used by `RtcpSrReporter` for instance, it must not be removed.

Issue #1260, opened 2024-09-26
Websocket response + disconnect race condition?


SeanMollet:
I'm shipping video via WebRTC to a "server" with a unique websocket behavior:

Connect, send SDP, it immediately responds with an answer and then disconnects.

Most of the time, onClosed ends up getting called and onMessage never is. Wireshark shows that the answer was received. Digging into PollService::runLoop, it seems like this is caused by a race. If the close is executed before the runLoop comes back around, the message ends up dropped on the floor.

Am I reading the code correctly?

From onClose, is there any easy way to get the data? receive() gives me nothing.

Happy to dig in and solve this, but wanted to get some thoughts from others on the best approach and/or validation that I'm reading all of this correctly.

Thanks

SeanMollet:
I'm using TLS and it seems the problem is that the socket close triggers TlsTransport::stop(), which unregisters the incoming callback., thus eating the received message.

paullouisageneau:
Thank you for reporting, you are right, data can still be waiting for TLS decryption when TCP disconnect. Could you please try with https://github.com/paullouisageneau/libdatachannel/pull/1266 ?

SeanMollet:
@paullouisageneau Testing now, will report back in a few minutes

SeanMollet:
Tested with #1266, fixed.

PR #1263, opened 2024-09-27
Shutdown race condition work-around

SeanMollet:
I doubt this is the best possible solution to the race condition, but I couldn't find anything else that was really viable, since we're waiting for the callback to get called by the main loop.

Adding a small delay (100ms) to TcpTransport::process allows the packets to get processed before we tear down the callbacks, etc.

### DIFF
--- a/src/impl/tcptransport.cpp
+++ b/src/impl/tcptransport.cpp
@@ -462,6 +462,11 @@ void TcpTransport::process(PollService::Event event) {
 		PLOG_ERROR << e.what();
 	}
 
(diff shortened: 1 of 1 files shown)


SeanMollet:
Fixes #1260

paullouisageneau (commented):
Thank you but adding a delay is definitely not a solution, especially because this blocks the entire poll service, meaning it blocks reception for all WebSockets.

SeanMollet:
Better fix provided by Paul-Louis. closing.

PR #1266, opened 2024-10-01
Fix race condition on WebSocket disconnect

paullouisageneau:
Should fix https://github.com/paullouisageneau/libdatachannel/issues/1260

### DIFF
--- a/src/impl/websocket.cpp
+++ b/src/impl/websocket.cpp
@@ -252,7 +252,8 @@ shared_ptr<TcpTransport> WebSocket::setTcpTransport(shared_ptr<TcpTransport> tra
 				remoteClose();
 				break;
 			case State::Disconnected:
-				remoteClose();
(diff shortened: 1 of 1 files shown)


SeanMollet:
@paullouisageneau Yes, this fixes it. Thank you.

Am I correct in my reading that this ultimately results in the timeout cleaning things up later and that's why it's safe to skip the earlier remoteClose?

SeanMollet:
Fixes #1260

paullouisageneau:
> Am I correct in my reading that this ultimately results in the timeout cleaning things up later and that's why it's safe to skip the earlier remoteClose?

It cleans up later but not on a timeout. Now, only `WsTransport` changing state can trigger close after the initial connection, so the EOF event has to travel trough the TCP and TLS layers, ensuring messages are delivered first.

PR #1297, opened 2024-11-26
libjuice: cmake find does not use correct names on imported target

metalMajor:
The plural names are used but it should be e.g JUICE_INCLUDE_DIR instead of JUICE_INCLUDE_DIRS

This problem only occurs when you use an external system libjuice.


metalMajor:
FYI, I discovered this while trying to make a conan package for libdatachannel. Others from conan are helping with the packaging, so it's almost ready: https://github.com/conan-io/conan-center-index/pull/26013
libjuice has already been added in another PR in conan.

metalMajor:
By the way, you can find the conan package here: https://conan.io/center/recipes/libdatachannel?version=0.22.2

Maybe it could be mentioned in the readme?

paullouisageneau:
Yes, I added a link in the readme. Thank you for your work!

PR #1315, opened 2025-01-10
typo fix: SHA-384

ArtemMartus:
There was a misspelling sha-384 with sha-256 typo which effectively lead to broken further fingerprint parsing.  This bug currently affects OBS 30.2.3 
Was reproduced with libwebrtc server


paullouisageneau (approved):
Thank you for the fix!

paullouisageneau:
Released in `v0.22.4`

PR #1325, opened 2025-01-24
feat: allow specifying initial ice mode

achingbrain:
Exposes the new parameter introduced in https://github.com/paullouisageneau/libjuice/pull/294

Please see that PR for rationale.

The value is passed as a string, I guess it could be an enum or something instead?


achingbrain:
Closing as not needed

PR #1348, opened 2025-03-08
Added VP8Rtpdecpacketizer and VP8naluit

InMDev:


mpietras:
Great stuff, thank you!  I tested it out and it's working fine for me.  A few thoughts unrelated to this specific change though:
- it would be handy to have in `onFrame` in `FrameInfo` an indication of `bool is_key` and possibly also the frame dimensions (if available).  It seems like an obvious thing to have that's specific to each codec (and depacketizer), and helpful when writing a video file for example.
- all the depacketizers steal the messages so we either get `onFrame` or `onMessage` but not _both_... it would be helpful in forward AND logging situations for example.  I get that it doesn't make sense to always copy around messages, but maybe a standard _optional_ depacketizer function would be to not remove packets in the `std::remove_if` and instead copy them e.g. `mRtpBuffer.push_back(std::make_shared<Message>(*message));` or maybe there's a better way...

paullouisageneau:
@mpietras 

> it would be handy to have in `onFrame` in `FrameInfo` an indication of `bool is_key` and possibly also the frame dimensions (if available).  It seems like an obvious thing to have that's specific to each codec (and depacketizer), and helpful when writing a video file for example.

An `isKey` flag is definitely something that could be added, but dimensions are lower-level information and would require parsing the actual bitstream. I'm not sure the library should be responsible for this.

> all the depacketizers steal the messages so we either get `onFrame` or `onMessage` but not _both_... it would be helpful in forward AND logging situations for example.  I get that it doesn't make sense to always copy around messages, but maybe a standard _optional_ depacketizer function would be to not remove packets in the `std::remove_if` and instead copy them e.g. `mRtpBuffer.push_back(std::make_shared<Message>(*message));` or maybe there's a better way...

Good point, but this is besides the scope of this PR.

mpietras:
@paullouisageneau 

> but dimensions are lower-level information and would require parsing the actual bitstream

It's true, but the depacketizer is already intimately knowledgeable about the codec packetization (clearly) and the frame size of a key-frame is readily available in the first ~20 octets of the frame headers for the codecs I know about at least so it's not much of a stretch...

paullouisageneau:
@InMDev Any news on this?

petersomers:
@InMDev, this would be really nice to have. Thanks for putting it together! Is it close to merging?

longnguyen2004:
This kinda looks like AI-generated code to me, at least some parts of it. Even if it's not, shoehorning VP8 into `NalUnit` is very weird to me, because NAL units are fundamentally an H26x thing only. Functions like `forbiddenBit` and `nri` will return bogus values if given a VP8 frame.

paullouisageneau:
@InMDev In the absence of follow-up, I'm going to close this in favor of a rewritten implementation not copying or relying on H264 stuff.

PR #1364, opened 2025-04-05
Fix the vector issue in AV1 fragment function

melpon:
It seems that an incorrect porting was done during the refactoring of https://github.com/paullouisageneau/libdatachannel/commit/4c3fb5a592af0988c2c64682aa50fb71e7d0280b


paullouisageneau:
Thank you for the fix!

Issue #1368, opened 2025-04-09
Allow zero byte messages


AltayAkkus:
https://github.com/paullouisageneau/libdatachannel/blob/32a15328a4a6109c389782914ac8e2239c2cd091/src/impl/dtlssrtptransport.cpp#L102C2-L103C57

Nice idea, and a very good warning, but [RFC 6263](https://datatracker.ietf.org/doc/html/rfc6263) says

> [4](https://datatracker.ietf.org/doc/html/rfc6263#section-4).  List of Alternatives for Performing RTP Keepalive
> 
>    This section lists, in no particular order, some alternatives that
>    can be used to perform a keepalive message within RTP media streams.
> 
> [4.1](https://datatracker.ietf.org/doc/html/rfc6263#section-4.1).  Empty (0-Byte) Transport Packet
> 
>    The application sends an empty transport packet (e.g., UDP packet,
>    Datagram Congestion Control Protocol (DCCP) packet).

So there is a valid use-case for sending messages which are < 8 bytes.

PR #1369, opened 2025-04-09
add zero byte msgs

AltayAkkus:
As mentioned in #1368 

By the way, the `size < 8` check in the `recvMedia` is completely fine, this is the exact mechanism which RFC6263 'exploits'. 

The receiver just truncates these packets, but the session is kept alive.


paullouisageneau (commented):
LGTM, thanks. Note that with libjuice as ICE transport, 0-byte packets won't reach libdatachannel as they are ignored at reception.

paullouisageneau (approved):

PR #1370, opened 2025-04-10
Fix not setting rtp config for all packetizers

parallelcc:


paullouisageneau (approved):
Thank you for the fix!

PR #1373, opened 2025-04-14
Add G722 codec and packetizer

parallelcc:


paullouisageneau (approved):
Thank you for the PR!

PR #1376, opened 2025-04-19
TWCC extension

ognkrmms:
Hi Paul,

This is the first half of the congestion control (CC) algorithm. This PR adds two new packet types: RtpTwccExt, which will be inserted into the RTP packets and RtcpTwcc, which is the type of the report we will get back. 

TwccHandler is responsible for inserting the RtpTwccExt header. TWCC sequence numbers will be recorded via a callback inside TwccHandler using ChainInterop. It will add new entries to the ChainInterop.

ChainInterop will be consumed by CC as the TWCC reports arrive.

Looking forward to your review.

### DIFF
--- a/include/rtc/chaininterop.hpp
+++ b/include/rtc/chaininterop.hpp
@@ -1,0 +1,82 @@
+#ifndef RTC_CHAIN_INTEROP_H
+#define RTC_CHAIN_INTEROP_H
+
+#if RTC_ENABLE_MEDIA
+#include "common.hpp"
+#include <map>
+#include <deque>
+#include <vector>
+#include <chrono>
+// ChainInterop will be used for storing info about TWCC packets
+// RtpTwccHandler produces new entries inside ChainInterop as it is processing RTP packets.
+// GCC algorithm consumes it.
+namespace rtc {
+	struct RTC_CPP_EXPORT BitrateStats {
+		double txBitsPerSecond, rxBitsPerSecond, packetLoss;
+		BitrateStats() : txBitsPerSecond(0), rxBitsPerSecond(0), packetLoss(0) {}
+		BitrateStats(double tx, double rx, double loss) : txBitsPerSecond(tx), rxBitsPerSecond(rx), packetLoss(loss) {};
+	};
+	
+	struct PacketInfo {
+		bool isReceived;
+		bool isSent;
+		uint16_t numBytes;
+		std::chrono::microseconds arrivalDuration;
+		std::chrono::steady_clock::time_point departureTime;
+		PacketInfo(uint16_t numBytes);
+	};
+	// WholeFrameInfo is an internal representation for a video frame
+	class WholeFrameInfo {
+		std::chrono::steady_clock::time_point time;
+		uint16_t seqNumStart, seqNumEnd;
+	
+	public:
+		WholeFrameInfo(std::chrono::steady_clock::time_point time, uint16_t seqNumStart, uint16_t seqNumEnd);
+		std::chrono::steady_clock::time_point getTime() const;
+		uint16_t getSeqNumStart() const;
+		uint16_t getSeqNumEnd() const;
+	};
+	
+	struct RTC_CPP_EXPORT ArrivalGroup {
+		std::vector<PacketInfo> packets;
+		std::chrono::microseconds arrivalTime;
+		std::chrono::steady_clock::time_point departureTime;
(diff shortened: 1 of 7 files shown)


src/rtptwcchandler.cpp
```diff
@@ -0,0 +1,49 @@
...
+	// Add TWCC exstension to RTP packets
+	for (unsigned int i = 0; i < messages.size(); i++) {
+		auto packet = messages[i];		
+		auto rtpHeader = reinterpret_cast<RtpHeader *>(packet->data());
+		outgoing.push_back(make_message(packet->size() + TWCC_EXT_HEADER_SIZE));
+		uint8_t *src = reinterpret_cast<uint8_t *>(packet->data());
+		uint8_t *dst = reinterpret_cast<uint8_t *>(outgoing.back()->data());
+		std::memcpy(dst, src, rtpHeader->getSize());
+		// set extension bit 1
+		dst[0] = (dst[0] & ~0x10) | (0x01 << 4);
```
paullouisageneau, line 30: You should use existing methods like `RtpHeader::setExtension()`.

include/rtc/rtp.hpp
```diff
@@ -356,6 +356,33 @@ struct RTC_CPP_EXPORT RtpRtx {
...
+
+	[[nodiscard]] uint16_t getBaseSeqNum() const;
+	[[nodiscard]] uint16_t getPacketStatusCount() const;
+	[[nodiscard]] uint32_t getReferenceTime() const;
+	[[nodiscard]] uint8_t getFbPacketCount() const;
+	[[nodiscard]] char* getBody();
+};
+
+struct RTC_CPP_EXPORT RtpTwccExt {
+	RtpExtensionHeader header;
```
paullouisageneau, line 374: You should not include the extension header here and handle it separately.

src/rtptwcchandler.cpp
```diff
@@ -0,0 +1,49 @@
...
+		src += rtpHeader->getSize();
+		dst += rtpHeader->getSize();
+
+		twccHeader.setTwccSeqNum(twccSeqNum++);
+		std::memcpy(dst, &twccHeader, TWCC_EXT_HEADER_SIZE);
```
paullouisageneau, line 35: What if the RTP packet already has extensions? It looks like this would override them, and since the size of the existing extension header might be different from the one you add with only TWCC, it could create a corrupted payload.

include/rtc/rtptwcchandler.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "mediahandler.hpp"
+#include "rtp.hpp"
+namespace rtc {
+
+class RTC_CPP_EXPORT TwccHandler final : public MediaHandler {
+	RtpTwccExt twccHeader;
+	uint16_t twccSeqNum;
+	// Callback below is for recording TWCC seqNums
+	std::function<void(message_vector &)> processPacketsCallback;
```
paullouisageneau, line 14: Why do you need a callback? The user should not be responsible for this, if you want to record the sequence number you should pass the recording object and call the corresponding method directly from the handler.

include/rtc/chaininterop.hpp
```diff
@@ -0,0 +1,82 @@
...
+		ArrivalGroup() = default;
+		ArrivalGroup(const ArrivalGroup &other);
+		ArrivalGroup(ArrivalGroup&& other) noexcept;
+		ArrivalGroup& operator=(const ArrivalGroup& other);
+		ArrivalGroup& operator=(ArrivalGroup&& other) noexcept;
+		void add(PacketInfo packet);	
+		void reset();
+	};
+	
+	class RTC_CPP_EXPORT ChainInterop {
```
paullouisageneau, line 54: I'm not a fan of the `ChainInterop` name as it doesn't really say anything about what state the class keeps or what it does.

Also, is this an API you want to expose to the user? It seems to me the user should only access a high-level rate controller interface and the internal logic and helper objects should be somewhere in `impl`.

paullouisageneau (commented):
Thank you for the contribution. I'm a bit puzzled by the whole architecture, do you run the congestion control algorithm from user code? It seems to me only a high-level rate controller interface should be exposed to the user, they should not be responsible for passing sequence numbers around.

Sum1cares (approved):

Sum1cares (approved):

Issue #1375, opened 2025-04-17
mSendMutex deadlock?


bobk-rey:
We've been trying to track down buffer corruption/issues when sending data via a websocket (eg 10MB with enough loss/latency to let the socket buffer actually fill up).

Commit https://github.com/paullouisageneau/libdatachannel/commit/b39a35e6fde9b3d45c09873ee37cf090706a750a started grabbing the mSendMutex from the polling thread, which seemed like a good step for preventing out-of-order issues with the application thread pushing more data.

Unfortunately it seems to have introduced a deadlock inside TcpTransport::setPoll aka PollService::Instance().add (with mMutex?)

bobk-rey:
Related to https://github.com/paullouisageneau/libdatachannel/issues/1293 ?

bobk-rey:
![Image](https://github.com/user-attachments/assets/3293e889-ae0a-46e5-8d68-5d8e5e0b0db4)

bobk-rey:
Having PollService::process unlock mMutex before calling callbacks seems to do the trick - at least stopped breaking or deadlocking when making a bunch of transfers.

Callback functions directly passing the 'this' pointer in seems a bit sketchy though. 
https://github.com/paullouisageneau/libdatachannel/blob/456c2151b7a1568b3805397a2d016e905c18a42e/src/impl/tcptransport.cpp#L329

Seems like you'd need to create the weak_ptr before deleting the object - not sure that calling this with a deleted transport and then having it check itself is the right move?
https://github.com/paullouisageneau/libdatachannel/blob/456c2151b7a1568b3805397a2d016e905c18a42e/src/impl/tcptransport.cpp#L413

PR #1379, opened 2025-04-24
Fix TCP mutex deadlock

bobk-rey:
Resolves paullouisageneau/libdatachannel#1375

Switches PollService::process to using a todo list for the callbacks so they can be called without mMutex locked.
Updates TcpTransport::setPoll to grab a weak pointer ahead of time.

### DIFF
--- a/src/impl/tcptransport.cpp
+++ b/src/impl/tcptransport.cpp
@@ -324,9 +324,10 @@ void TcpTransport::configureSocket() {
 }
 
 void TcpTransport::setPoll(PollService::Direction direction) {
+	auto weakSelf = weak_from_this();
 	PollService::Instance().add(
 	    mSock, {direction, direction == PollService::Direction::In ? mReadTimeout : nullopt,
-	            std::bind(&TcpTransport::process, this, _1)});
+	    [weakSelf](PollService::Event event) { if (auto self = weakSelf.lock()) { self->process(event); } } });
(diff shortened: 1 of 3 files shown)


src/impl/tcptransport.cpp
```diff
@@ -324,9 +324,10 @@ void TcpTransport::configureSocket() {
 }
 
 void TcpTransport::setPoll(PollService::Direction direction) {
+	auto weakSelf = weak_from_this();
 	PollService::Instance().add(
 	    mSock, {direction, direction == PollService::Direction::In ? mReadTimeout : nullopt,
-	            std::bind(&TcpTransport::process, this, _1)});
+	    [weakSelf](PollService::Event event) { if (auto self = weakSelf.lock()) { self->process(event); } } });
```
paullouisageneau, line 330: You should use the `weak_bind` helper.
```suggestion
        weak_bind(&TcpTransport::process, this, _1)});
```
You also need to change the other call to `PollService::Instance().add()`.

paullouisageneau (commented):
Thank you for the fix. This could also be achieved by delegating tasks to a  `Processor`, but the todo list is fine. You should use the `weak_bind` helper in `TcpTransport` for the sake of consistency.

paullouisageneau (approved):
Looking good, thanks for the fix!

PR #1394, opened 2025-05-18
remove a compile-time flag that completely disables media tracks

aumfer:
Hi Paul what is the reasoning behind this as it prevents me from receiving media packets.

### DIFF
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -1091,13 +1091,6 @@ void PeerConnection::processRemoteDescription(Description description) {
 			PLOG_DEBUG << "New remote track, mid=\"" << remoteMedia->mid() << "\"";
 
 			auto reciprocated = remoteMedia->reciprocate();
-#if !RTC_ENABLE_MEDIA
-			if (!reciprocated.isRemoved()) {
-				// No media support, mark as removed
-				PLOG_WARNING << "Rejecting track (not compiled with media support)";
-				reciprocated.markRemoved();
-			}
-#endif
 
 			// Create incoming track
 			auto track = std::make_shared<Track>(weak_from_this(), std::move(reciprocated));


paullouisageneau:
The flag `RTC_ENABLE_MEDIA` allows you to build the library with or without media support. If you need only data channels, setting `RTC_ENABLE_MEDIA=0` removes the dependency on libsrtp and makes the library lighter. Note [the flag defaults to 1](https://github.com/paullouisageneau/libdatachannel/blob/0e40aeb058b947014a918a448ce2d346e6ab14fe/include/rtc/rtc.h#L40), so if tracks are disabled, it's because you have set the flag to 0 somehow.

Removing the check here will enable tracks during negotiation but won't magically enable media reception. Also, other users rely on that flag.

aumfer:
You may as well throw an exception when the user calls rtc::PeerConnection::onTrack, or rtc::PeerConnection::addTrack. As it is it is every confusion to receive tracks from the client and have them immediately removed. Also I don't believe the change here would require bringing in the libsrtp dependency.

For context, I'm looking to receive messages from tracks and I'm hoping to use the version of libdatachannel available on vcpkg, which doesn't compile with RTC_ENABLE_MEDIA.

paullouisageneau:
> You may as well throw an exception when the user calls rtc::PeerConnection::onTrack, or rtc::PeerConnection::addTrack. As it is it is every confusion to receive tracks from the client and have them immediately removed. Also I don't believe the change here would require bringing in the libsrtp dependency.

The change doesn't require bringing libsrtp, but the library is needed to actually receive media. libsrtp, along with the whole media transport implementation, will be absent if `RTC_ENABLE_MEDIA=0`. Arguably, not marking tracks as rejected while they are not functional would be even more confusing.

> For context, I'm looking to receive messages from tracks and I'm hoping to use the version of libdatachannel available on vcpkg, which doesn't compile with RTC_ENABLE_MEDIA.

I'm not responsible for the vcpkg port, but it has a `srtp` [feature](https://vcpkg.link/ports/libdatachannel#features) that you must enable to receive media transport.

PR #1405, opened 2025-05-28
example:  whip whep

stephanrotolante:
a whip whep example using libdatachannel


paullouisageneau (commented):
Thank you for submitting the example. Currently there are issues with format negotiation, I might work only by chance depending on payload IDs picked by the browser.

examples/whip-whep/whep.cpp
```diff
@@ -0,0 +1,118 @@
...
+		int len;
+		while ((len = recv(sock, buffer, BUFFER_SIZE, 0)) >= 0) {
+
+			if (len < sizeof(rtc::RtpHeader) || !track->isOpen()) {
+				continue;
+			}
+
+			rtc::RtpHeader *rtp = reinterpret_cast<rtc::RtpHeader *>(buffer);
+			rtp->setSsrc(ssrc);
+			rtp->setPayloadType(payloadType);
```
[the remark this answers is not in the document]
paullouisageneau: It means there are more than one incoming RTP streams, overriding the payload type and SSRC on both of them creates a broken stream, in particular SRTP complains about inconsistent sequence numbers. It looks like your ffmpeg command sends an AAC audio stream, you should remove it.

examples/whip-whep/whip.cpp
```diff
@@ -0,0 +1,123 @@
...
+				track->setMediaHandler(session);
+
+				// forward rtp data
+				track->onMessage(
+				    [&sock, &addr](rtc::binary message) {
+					    if (message.size() == 0) {
+						    return;
+					    }
+
+					    rtc::RtpHeader *rtp = reinterpret_cast<rtc::RtpHeader *>(message.data());
```
paullouisageneau, line 83: You must check the size of `message` before casting to `rtc:RtpHeader`.

examples/whip-whep/whep.cpp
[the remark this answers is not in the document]
stephanrotolante: yes, there was an issue ffmpeg where the ssrc and sequence number would reset during the stream

HR1025:
This example looks great, but there are still some aspects that don't fully comply with the specification. For example:

- The **POST** request should return `201 Created`.
- The **POST** request should include a **Location header**, representing the resource (so that it can be deleted later via `DELETE`).
- A **DELETE operation** needs to be implemented to release the PeerConnection resource (which means the PeerConnection should be dynamically created during the POST request).
- The SDP answer should be properly negotiated—after receiving the remote offer, only retain and modify the necessary parts.

Could you improve these points? Or I can try making some modifications myself—I'm really interested in this WHIP server example!

HR1025:
RFC 9725 WebRTC-HTTP Ingestion Protocol (WHIP)

image

image

HR1025:
I also think that creating an integrated example combining both **WHEP** and **WHIP** would be a great choice. You can use **WHIP** for publishing (pushing) streams and **WHEP** for playback (pulling) streams.

This way, you don't need to prepare any media stream data in advance—it's very convenient for self-testing.

stephanrotolante:
@HR1025 

After this PR gets merged,  I will open up another implementing some of your suggestions. I will respond back with 201 status code, add the url path in the response, and consolidated the `whip.cpp`/`whp.cpp` into a single file

examples/whip-whep/README.md
```diff
@@ -0,0 +1,52 @@
...
+    x264enc speed-preset=ultrafast tune=zerolatency ! queue ! \
+    mp4mux name=mux ! \
+    filesink location=dummy_video.mp4
+```
+
+Open index.html in the browser.
+
+Run the `whep` executable to start the http server. You should be prompted with a message saying the application is running on port 8080.
+
+Once the http server is up, the the "Subscribe" button the begin the signaling.
```
nab-os: "the the" misstype

examples/whip-whep/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Now you should see video on the video player in the browser
+
+## Using the WHIP example
+
+Open index.html in the browser.
+
+Run the `whip` executable to start the http server. You should be prompted with a message saying the application is running on port 8080.
+
+Once the http server is up, the the "Publish" button the begin the signaling.
```
nab-os: "the the" misstype here too

nab-os (commented):
First review here,
That's a nice example, thank you !

Sum1cares (approved):

Sum1cares (approved):

PR #1413, opened 2025-06-15
Adds fraction and packet lost calculation in RR packets

dledanseur:
This PR adds the calculation of the fraction lost and total packet lost values in RR packets. The calculation algorithm is based on the algorithm from the RFC3550 (see [Appendix A.1 RTP Data Header Validity Check](https://datatracker.ietf.org/doc/html/rfc3550#appendix-A.1)).


paullouisageneau (approved):
Thank you for implementing!

Issue #1342, opened 2025-02-20
Custom headers for websocket


HugoMPaz:
I am trying to customize a header to send a websession token through for authentication purposes but currently it seems to be impossible. Is there any way that i can work around for it? 

Sending this token back in header is a must for my use case.

paullouisageneau:
There is currently no way to insert custom headers in the API, but in could be added, for instance with an optional second argument to `open()` which would be passed to the `WsHandshake`.

Jeremy-Boyle:
I'll do a MR for this :)

Jeremy-Boyle:
@paullouisageneau mind taking a look at the PR when you get a chance thank you !

PR #1424, opened 2025-07-03
Feature: Allow custom headers to be passed via WebSocket

Jeremy-Boyle:
Implement functionality to send and access custom headers during WebSocket handshake and server communication. This enhancement improves authentication and client-server interactions.

Closes: #1342

To Do:
- [x] Update readme with new examples
- [x] Update websocket api to allow optional headers
- [x] Update websocket server to be able to handle headers
  - [x] Implement way to get the headers
- [x] Update Tests
- [ ] Update c-api (Can you please help implement this, not sure what would be best here, for the new functions / names)

### DIFF
--- a/src/impl/wshandshake.cpp
+++ b/src/impl/wshandshake.cpp
@@ -30,8 +30,8 @@ using std::chrono::system_clock;
 
 WsHandshake::WsHandshake() {}
 
-WsHandshake::WsHandshake(string host, string path, std::vector<string> protocols)
-    : mHost(std::move(host)), mPath(std::move(path)), mProtocols(std::move(protocols)) {
+WsHandshake::WsHandshake(string host, string path, std::vector<string> protocols, std::map<string, string> headers)
+    : mHost(std::move(host)), mPath(std::move(path)), mProtocols(std::move(protocols)), mCustomHeaders(std::move(headers)) {
 
 	if (mHost.empty())
 		throw std::invalid_argument("WebSocket HTTP host cannot be empty");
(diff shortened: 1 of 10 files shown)


src/impl/wshandshake.cpp
```diff
@@ -73,6 +78,11 @@ string WsHandshake::generateHttpRequest() {
 	if (!mProtocols.empty())
 		out += "Sec-WebSocket-Protocol: " + utils::implode(mProtocols, ',') + "\r\n";
 
+	// Add custom headers
+	for (const auto& [headerName, headerValue] : mCustomHeaders) {
+		out += headerName + ": " + headerValue + "\r\n";
+	}
```
paullouisageneau, line 84: I think you should at least sanitize `headerValue`:
```suggestion
	for (auto [headerName, headerValue] : mCustomHeaders) {
	    headerValue.erase(std::remove(headerValue.begin(), headerValue.end(), '\r'), headerValue.end());
	    std::replace(headerValue.begin(), headerValue.end(), '\n', ' ');
		out += headerName + ": " + headerValue + "\r\n";
	}
```

src/impl/wshandshake.hpp
```diff
@@ -57,6 +58,8 @@ class WsHandshake final {
 	string mHost;
 	string mPath;
 	std::vector<string> mProtocols;
+	std::map<string, string> mCustomHeaders;
+	std::multimap<string, string> mRequestHeaders;
```
paullouisageneau, line 62: Is there a reason to use two different maps here? You could use only one multimap.

paullouisageneau (commented):
It looks good overall, thank you. I just have a few comments, the main one being about introducing two maps whereas it seems to me that one would do the job.

src/impl/wshandshake.hpp
Jeremy-Boyle: Hey ! It's to add support for the websocket server. 

But good suggestion. Rather than custom and request we can just rename them to request headers or just headers up to you.

ungive:
Hi, just FYI, I've implemented this too and made a PR: https://github.com/paullouisageneau/libdatachannel/pull/1488

These two PRs clash of course, so they should probably both be reviewed before merging one of them.

paullouisageneau:
Since there has been no activity for a while, I'm closing this in favor of https://github.com/paullouisageneau/libdatachannel/pull/1488

PR #1429, opened 2025-07-15
Add getter for sync NTPTS and RTPTS values in RtcpReceivingSession

petersomers:


include/rtc/rtcpreceivingsession.hpp
```diff
@@ -58,9 +66,10 @@ class RTC_CPP_EXPORT RtcpReceivingSession : public MediaHandler {
 	uint32_t mTransit = 0;			// relative trans time for prev pkt
 	uint32_t mJitter = 0;	
 	
-	uint64_t mSyncRTPTS, mSyncNTPTS;
+	SyncPTS mSyncPTS{0,0};
 
 	std::atomic<unsigned int> mRequestedBitrate = 0;
+	std::mutex mSyncMutex;
```
Nemirtingas, line 72: I'm not a maintainer, but creating a mutex only for 2 longs seems a bit overkill. They can still be stored as 2 uint64_t, or rather std::atomic<uint64_t>, then you can build your structure in an atomic way in the getSyncPTS without a lock.

petersomers:
@paullouisageneau Could I get some eyes on this? Or the opinion not to add it?

paullouisageneau (commented):
Thank you for the PR. It looks good to me, but I'd suggest renaming the getter and fields for consistency.

PR #1441, opened 2025-08-08
Added color space extension support

LeeTaekLim:
Hello!

I had been using your codes well but I had a problem with handling h.264 video color. The color of decoded video was not the same as the original video. So I searched a lot to fix this problem and finally I found this.
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/color-space/
I implemented codes based on the doc and the color issue was resolved! Since my video has no HDR metadata, I made **only** the RTP extension header without HDR metadata.

I think other people would need this color related feature so please review this PR.


paullouisageneau (approved):
It looks good, thank you for adding the extension.

PR #1458, opened 2025-09-20
Document proxyServer field in rtcWsConfiguration

egnima:
The proxyServer field passed to rtcCreateWebSocket was previously undocumented, making the rest of the fields' indexes inaccurate


paullouisageneau (approved):
Thanks for adding.

PR #1488, opened 2025-12-09
Implement custom WebSocket HTTP request headers

### LINKED ISSUE #1342: Custom headers for websocket

ungive:
There already is a PR for this (#1424), but since there hasn't been any activity and I need this for my project, I went ahead and implemented custom HTTP headers myself.

Changes:
- Added an optional `headers` parameter to the `rtc::WebSocket::open` method
- Changed `rtc::impl::parseHttpHeaders` to return an `std::multimap` with a comparator that ignores case-sensitivity. The casing of the headers is kept intact (the string is not transformed). The behaviour should be the same as before
- Added the `rtc::WebSocket::requestHeaders` method which returns the headers that were configured for the WebSocket client OR the headers that were sent by the client and received by the server
- Adapted the WebSocketServer test to check that headers are sent and received and properly sanitized (as mentioned in the other PR)
- Added a `rtc::Websocket::Headers` type alias to make it more convenient to build a map of headers and then pass it to `rtc::WebSocket::open`

Closes #1342

TODO:
- [ ] Implement HTTP authentication via URL (could be done in a separate PR): https://github.com/paullouisageneau/libdatachannel/blob/faed3cf9fbf6a530b273439f56eeccce2ba226ed/src/impl/websocket.cpp#L98-L100
- [ ] Update documentation
- [ ] Update C API

### DIFF
--- a/src/impl/websocket.cpp
+++ b/src/impl/websocket.cpp
@@ -68,7 +68,7 @@ WebSocket::WebSocket(optional<Configuration> optConfig, certificate_ptr certific
 
 WebSocket::~WebSocket() { PLOG_VERBOSE << "Destroying WebSocket"; }
 
-void WebSocket::open(const string &url, const ::rtc::WebSocket::Headers &headers) {
+void WebSocket::open(const string &url, const rtc::WebSocket::Headers &headers) {
 	PLOG_VERBOSE << "Opening WebSocket to URL: " << url;
 
 	if (state != State::Closed)
@@ -236,7 +236,7 @@ shared_ptr<TcpTransport> WebSocket::setTcpTransport(shared_ptr<TcpTransport> tra
 		transport->onBufferedAmount(weak_bind(&WebSocket::triggerBufferedAmount, this, _1));
 
 		transport->onStateChange([this, weak_this = weak_from_this()](State transportState) {
-			if(auto locked = weak_this.lock())
+			if (auto locked = weak_this.lock())
 				std::invoke([=]() {
 					switch (transportState) {
 					case State::Connected:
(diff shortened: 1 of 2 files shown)


src/impl/websocket.cpp
```diff
@@ -68,7 +68,7 @@ WebSocket::WebSocket(optional<Configuration> optConfig, certificate_ptr certific
 
 WebSocket::~WebSocket() { PLOG_VERBOSE << "Destroying WebSocket"; }
 
-void WebSocket::open(const string &url) {
+void WebSocket::open(const string &url, const ::rtc::WebSocket::Headers &headers) {
```
paullouisageneau:
```suggestion
void WebSocket::open(const string &url, const rtc::WebSocket::Headers &headers) {
```

src/impl/websocket.hpp
```diff
@@ -35,7 +35,7 @@ struct WebSocket final : public Channel, public std::enable_shared_from_this<Web
 	WebSocket(optional<Configuration> optConfig = nullopt, certificate_ptr certificate = nullptr);
 	~WebSocket();
 
-	void open(const string &url);
+	void open(const string &url, const std::map<string, string, case_insensitive_less> &headers);
```
paullouisageneau:
```suggestion
	void open(const rtc::WebSocket::Headers &headers);
```

paullouisageneau (approved):
It looks good, thank you for implementing. I only have a couple nitpicks about types.

ungive:
@paullouisageneau Should be ready to merge now!

PR #1492, opened 2025-12-11
Return whether preload performed initialization

ungive:
I think it would be a good addition to return a boolean from the global `Preload` method that indicates whether or not the function actually performed the initialization of the RTC library. This will give certainty that configuration through functions like `rtc::SetThreadPoolSize(n)` was actually applied. In the case of this function the thread pool size is not actually set, if initialization was already performed beforehand, which can be the case when creating an `rtc::WebSocket` instance, which silently calls `Init::Instance().token()` (which in turn triggers global library initialization).

Here's a usage example: https://github.com/ungive/loon/commit/2a7427a57ed08dcf0686eb9d5b7dbd5dc20943d9

### DIFF
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -52,7 +52,7 @@ struct SctpSettings {
 RTC_CPP_EXPORT void SetSctpSettings(SctpSettings s);
 
 // Optional global preload and cleanup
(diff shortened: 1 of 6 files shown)


Sum1cares (approved):

paullouisageneau (approved):
I think this makes sense, thanks.

PR #1499, opened 2025-12-17
Improve PacingHandler

Polusummator:
Changes:
* Add `setBitrate` setter for dynamic bitrate limit change
* Add `setMaxQueueSize` setter (also size can be set in ctor). There is a problem that buffer is unlimited and can grow indefinitely leading to long send delays. Now you can set queue size limit and drop queue if it is full. Also there is an option to set `onOverflowCallback`. By default queue is unlimited
* Change send interval calculation. Sending packets takes time, but this is not considered when scheduling further sends, although it is included in budget calculation

### DIFF
--- a/include/rtc/pacinghandler.hpp
+++ b/include/rtc/pacinghandler.hpp
@@ -23,11 +23,13 @@ namespace rtc {
 // and delivers them in a smoother manner by sending a fixed size of them on an interval
 class RTC_CPP_EXPORT PacingHandler : public MediaHandler {
 public:
-	PacingHandler(double bitsPerSecond, std::chrono::milliseconds sendInterval, size_t maxQueueSize=0, std::function<void(void)> onOverflowCallback=nullptr);
+	PacingHandler(double bitsPerSecond, std::chrono::milliseconds sendInterval, size_t maxQueueSize=0);
 
 	void setBitrate(double bitsPerSecond);
 
-	void setMaxQueueSize(size_t maxQueueSize);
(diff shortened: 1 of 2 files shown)


include/rtc/pacinghandler.hpp
```diff
@@ -23,7 +23,11 @@ namespace rtc {
 // and delivers them in a smoother manner by sending a fixed size of them on an interval
 class RTC_CPP_EXPORT PacingHandler : public MediaHandler {
 public:
-	PacingHandler(double bitsPerSecond, std::chrono::milliseconds sendInterval);
+	PacingHandler(double bitsPerSecond, std::chrono::milliseconds sendInterval, size_t maxQueueSize=0, std::function<void(void)> onOverflowCallback=nullptr);
```
paullouisageneau: For consistency with the rest of the library's interface, there should be a `onOverflow` setter which would set a `synchronized_callback mOverflowCallback`.

paullouisageneau (commented):
Thank you for the improvements. I suggest adding a `onOverflow()` method instead and changing the overflow logic. Also, for a pacer, I think the size should be in bytes, since it would correspond precisely to a send duration.

longnguyen2004:
@Polusummator Are you going to continue working on this? This is important for pacing VBR video streams

Polusummator:
> @Polusummator Are you going to continue working on this? This is important for pacing VBR video streams

Oh, I forgot about this PR. Yes, I will fix it in a few days

PR #1500, opened 2025-12-17
Fix REMB SSRC

Polusummator:
According to [REMB RFC](https://datatracker.ietf.org/doc/html/draft-alvestrand-rmcat-remb-03):

> SSRC of media source (32 bits): Always 0; this is the same convention as in [RFC5104] section 4.2.2.2 (TMMBN).
> ...
> SSRC feedback (32 bits) Consists of one or more SSRC entries which this feedback message applies to.

So `SSRC feedback` should be taken, not `SSRC of media source`
Also the number of `SSRC feedback` entries can be greater than 1

### DIFF
--- a/include/rtc/rtp.hpp
+++ b/include/rtc/rtp.hpp
@@ -269,7 +269,7 @@ struct RTC_CPP_EXPORT RtcpRemb {
 
 	char _id[4];       // Unique identifier ('R' 'E' 'M' 'B')
 	uint32_t _bitrate; // Num SSRC, Br Exp, Br Mantissa (bit mask)
-	SSRC _ssrcs;
+	SSRC _ssrcs[1];
 
 	[[nodiscard]] static size_t SizeWithSSRCs(int count);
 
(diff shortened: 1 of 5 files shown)


include/rtc/rtp.hpp
```diff
@@ -269,15 +269,16 @@ struct RTC_CPP_EXPORT RtcpRemb {
 
 	char _id[4];       // Unique identifier ('R' 'E' 'M' 'B')
 	uint32_t _bitrate; // Num SSRC, Br Exp, Br Mantissa (bit mask)
-	SSRC _ssrc[1];
+	SSRC _ssrcs;
```
paullouisageneau: You should keep the array here, to indicate there can be multiple SSRCs (at least one) and `_ssrcs` is the pointer on the array:
```suggestion
	SSRC _ssrcs[1];
```

paullouisageneau (commented):
Good catch, thank you for fixing.

l1tsolaiki:
@paullouisageneau Hi! Any ETA on merge of this PR?

paullouisageneau (approved):

PR #1503, opened 2026-01-04
General API improvement

orange-cpp:
-added nodiscard to all methods that can return something
-marked some unchanged arguments const
-improved CMakeLists.txt (removed hardcoded file paths, added version parsing from VERSION file with automatic version definitions)
-removed version.h since it not used anymore


include/rtc/version.h
```diff
@@ -1,9 +0,0 @@
-#ifndef RTC_VERSION_H
-#define RTC_VERSION_H
-
-#define RTC_VERSION_MAJOR 0
-#define RTC_VERSION_MINOR 24
-#define RTC_VERSION_PATCH 0
-#define RTC_VERSION "0.24.0"
```
paullouisageneau, line 7: Removing the file would result in missing version defines in a number of cases. You replaced them with public target compile definitions, which are only present when libdatachannel is imported from CMake. For instance, if someone installs the library then includes it manually, `RTC_VERSION` won't be defined anymore. This would probably break the setup of a lot of users. It also breaks version defines when building is built from the bare Makefile.

include/rtc/channel.hpp
```diff
@@ -25,13 +25,13 @@ class RTC_CPP_EXPORT Channel : private CheshireCat<impl::Channel> {
 	virtual ~Channel();
 
 	virtual void close() = 0;
-	virtual bool send(message_variant data) = 0; // returns false if buffered
-	virtual bool send(const byte *data, size_t size) = 0;
+	[[nodiscard]] virtual bool send(message_variant data) = 0; // returns false if buffered
+	[[nodiscard]] virtual bool send(const byte *data, size_t size) = 0;
```
paullouisageneau, line 29: It's perfectly fine to ignore the return value, it only indicates if sending happened synchronously or if the message was buffered.

include/rtc/description.hpp
```diff
@@ -272,32 +272,32 @@ class RTC_CPP_EXPORT Description {
...
-	int addMedia(Media media);
-	int addMedia(Application application);
-	int addApplication(string mid = "data");
-	int addVideo(string mid = "video", Direction dir = Direction::SendOnly);
-	int addAudio(string mid = "audio", Direction dir = Direction::SendOnly);
+	[[nodiscard]] int addMedia(Media media);
+	[[nodiscard]] int addMedia(Application application);
+	[[nodiscard]] int addApplication(string mid = "data");
+	[[nodiscard]] int addVideo(string mid = "video", Direction dir = Direction::SendOnly);
+	[[nodiscard]] int addAudio(string mid = "audio", Direction dir = Direction::SendOnly);
```
paullouisageneau, line 283: Same here, it's fine to ignore the returned index.

include/rtc/utils.hpp
```diff
@@ -76,19 +76,19 @@ template <typename... Args> class synchronized_callback {
 		return *this;
 	}
 
-	bool operator()(Args... args) const {
+	[[nodiscard]] bool operator()(Args... args) const {
```
paullouisageneau, line 79: Same here, it's fine to ignore the returned value.

include/rtc/utils.hpp
```diff
@@ -76,19 +76,19 @@ template <typename... Args> class synchronized_callback {
...
-	operator bool() const {
+	[[nodiscard]] operator bool() const {
 		std::lock_guard lock(mutex);
 		return callback ? true : false;
 	}
 
 protected:
 	virtual void set(std::function<void(Args...)> func) { callback = std::move(func); }
-	virtual bool call(Args... args) const {
+	[[nodiscard]] virtual bool call(Args... args) const {
```
paullouisageneau, line 91: Same here, it's fine to ignore the returned value. Additionally, this method is protected.

include/rtc/mediahandler.hpp
```diff
@@ -35,15 +35,15 @@ class RTC_CPP_EXPORT MediaHandler : public std::enable_shared_from_this<MediaHan
 	/// @param send Send callback to send messages back to the peer
 	virtual void outgoing([[maybe_unused]] message_vector &messages, [[maybe_unused]] const message_callback &send) {}
 
-	virtual bool requestKeyframe(const message_callback &send);
-	virtual bool requestBitrate(unsigned int bitrate, const message_callback &send);
+	[[nodiscard]] virtual bool requestKeyframe(const message_callback &send);
+	[[nodiscard]] virtual bool requestBitrate(unsigned int bitrate, const message_callback &send);
```
paullouisageneau, line 39: Discarding is perfectly fine too.

src/impl/channel.cpp
```diff
@@ -14,7 +14,7 @@ namespace rtc::impl {
 void Channel::triggerOpen() {
 	mOpenTriggered = true;
 	try {
-		openCallback();
+		std::ignore = openCallback();
```
paullouisageneau, line 17: I don't see the point of adding `nodiscard` to then introduce `std::ignore` on most callback invocations. It looks extremely verbose for nothing.

paullouisageneau:
Thanks for the PR. Sadly, the version header simplification would break version defines in a lot of cases. The way it is currently implemented makes version defines available outside of CMake.

I'm fine with adding a few `[[nodiscard]]` but it's currently done in a blanket way without considering if return values could be discarded or not.

Sum1cares (approved):

PR #1511, opened 2026-01-27
Google Video Layers Allocation extension support

kmansoft:
Adds support for the Google "[Video Layers Allocation](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/video-layers-allocation00)" extension.

Assumptions:
- Separate from RFC 8851 even though it's similar info - gives more flexibility to applications (may want one but not the other), easier for the library to implement (don't have to reach into the SDP Description from RTP code)
- The application will configure the SDP negotiation for the VLA extension if it wants it
- The application will indicate to the library when a frame is a key frame
- Because the library will send the VLA (if configured) for the first 100 packets of a video stream + on each packet of each key frame - those are the recommendations from our SFU people, personally I think they are reasonable
- The MID and RID extensions are also required, those are already supported by the library and also need to be configured by the application (i.e. not picked up from the SDP automatically) - so I'm just following the existing pattern.

Most changes should be self-explanatory, let me just explain the `GoogleVideoLayerAllocation` shared pointer - this is _shared_ data about _all_ the layers that the application is sending. Since there is no guarantee for which layer's packets will be encoded / generated / reach the receiver first, the Google VLA works by including information about all layers in the extension's payload for each layer (== RID, SSRC).

There are tests.

### DIFF
--- a/src/rtppacketizer.cpp
+++ b/src/rtppacketizer.cpp
@@ -9,7 +9,7 @@
 #if RTC_ENABLE_MEDIA
 
 #include "rtppacketizer.hpp"
-#include "video_layers_allocation_ext.hpp"
+#include "video_layers_allocation.hpp"
 
 #include <cmath>
 #include <cstring>
@@ -41,11 +41,10 @@ message_ptr RtpPacketizer::packetize(const binary &payload, bool mark) {
 	binary videoLayersAllocationBuf;
 	if (rtpConfig->videoLayersAllocationId > 0 &&
 		rtpConfig->videoLayersAllocationStreams &&
-		checkVideoLayersAllocation()) {
+		shouldEmitVideoLayersAllocation()) {
 		// Yes, generate it now so we can calculate total size
-		videoLayersAllocationBuf = generateVideoLayersAllocation(
-		    rtpConfig->videoLayersAllocationStreams,
-		    rtpConfig->videoLayersAllocationStreamIndex);
+		videoLayersAllocationBuf = rtpConfig->videoLayersAllocationStreams->
+				generate(rtpConfig->videoLayersAllocationStreamIndex);
 	}
 
 	// Determine if a two-byte header is necessary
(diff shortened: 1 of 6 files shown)


src/rtppacketizer.cpp
```diff
@@ -151,7 +174,7 @@ message_ptr RtpPacketizer::packetize(const binary &payload, bool mark) {
 			               byte(rtpConfig->colorMatrix), byte(range_chr)};
 
 			offset += extHeader->writeHeader(
-			    twoByteHeader, offset, rtpConfig->playoutDelayId, data, 4);
+			    twoByteHeader, offset, rtpConfig->colorSpaceId, data, 4);
```
kmansoft: This is a pre-existing bug

kmansoft:
Hi,

I've updated the PR to:

- Remove mention of Google from the names of data structures etc. - current code doesn't call out the companies that designed various technology pieces, so this makes the PR consistent with that
- Fix an edge case in the generation of VLA extension payload

include/rtc/video_layers_allocation_ext.hpp
```diff
@@ -0,0 +1,53 @@
...
+	};
+
+	std::vector<RtpStream> rtpStreams;  // up to 4 streams
+};
+
+/// Generate the wire format for Google Video Layers Allocation RTP header extension
+/// @param allocation The layer allocation data
+/// @param streamIndex The RTP stream index (0-3) for this packet's stream
+/// @return Binary payload for the RTP header extension, empty if allocation is invalid
+binary RTC_CPP_EXPORT generateVideoLayersAllocation(
```
paullouisageneau: This could be a method `binary VideoLayersAllocation::generate(streamIndex)`.

paullouisageneau (commented):
It looks good, thank you. I only have a few nitpicks, mostly about naming.

kmansoft:
@paullouisageneau Thank you for reviewing.

Believe I addressed all your comments in the latest commit.

kmansoft:
Thank you!

PR #1536, opened 2026-04-06
Parse simulcast layers (rids) including RFC 8851 attributes

kmansoft:
As of now, `a=rid:foo` lines are parsed a generic attributes.

But `Description` now has a class to represent rid's, so we can use that.

This PR adds:

- Parsing simulcast rid's into `Rid` objects
- With RFC 8851 attributes (if present)

There are tests too.

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -57,30 +57,8 @@ inline std::pair<string_view, string_view> parse_pair(string_view attr) {
 	return std::make_pair(std::move(key), std::move(value));
 }
 
-inline std::vector<string_view> split_into_list(string_view value, char ch) {
-	std::vector<string_view> result;
-
-	size_t pos = 0;
-	size_t end = value.size();
-	while (pos < end) {
-		auto i = value.find(ch, pos);
-		if (i == string_view::npos) {
-			i = end;
-		}
-
-		const auto substr = value.substr(pos, i - pos);
-		if (!substr.empty()) {
-			result.push_back(substr);
-		}
-
-		pos = i + 1;
-	}
-
-	return result;
-}
-
-inline std::vector<std::pair<string_view, string_view>> parse_attr_list(string_view attr_list) {
-	std::vector<std::pair<string_view, string_view>> attrs;
+inline std::vector<std::pair<string, string>> parse_attr_list(const string& attr_list) {
+	std::vector<std::pair<string, string>> attrs;
 
 	size_t pos = 0;
 	size_t end = attr_list.size();
@@ -96,7 +74,7 @@ inline std::vector<std::pair<string_view, string_view>> parse_attr_list(string_v
 			const auto key = key_value.substr(0, j);
 			const auto value = key_value.substr(j + 1);
 			if (!key.empty() && !value.empty()) {
-				attrs.push_back(std::make_pair(key, value));
+				attrs.emplace_back(key, value);
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


src/description.cpp
```diff
@@ -57,6 +57,55 @@ inline std::pair<string_view, string_view> parse_pair(string_view attr) {
 	return std::make_pair(std::move(key), std::move(value));
 }
 
+inline std::vector<string_view> split_into_list(string_view value, char ch) {
```
paullouisageneau: This is already implemented by `utils::explode()`, please use the existing function.

src/description.cpp
```diff
@@ -57,6 +57,55 @@ inline std::pair<string_view, string_view> parse_pair(string_view attr) {
...
+
+inline std::vector<std::pair<string_view, string_view>> parse_attr_list(string_view attr_list) {
+	std::vector<std::pair<string_view, string_view>> attrs;
+
+	size_t pos = 0;
+	size_t end = attr_list.size();
+	while (pos < end) {
+		auto i = attr_list.find(';', pos);
+		if (i == string_view::npos) {
+			i = end;
```
paullouisageneau, line 68:
```suggestion
			break;
```

src/description.cpp
```diff
@@ -810,6 +863,39 @@ void Description::Entry::parseSdpLine(string_view line) {
 			// a bundled "m=" section from a BUNDLE group, the offerer [...] MUST NOT assign an SDP
 			// 'bundle-only' attribute to the "m=" section.
 			mIsRemoved = false;
+		} else if (key == "rid") {
+			// "a:rid=foo send" or optionally "a:rid=foo send abc=def;x=y;m=n"
+			const auto list = split_into_list(value, ' ');
+			if (list.size() < 2 || list.size() > 3) {
```
paullouisageneau: Maybe for the sake of compatibility it would be good to ignore anything after.

```suggestion
			if (list.size() < 2) {
```

src/description.cpp
```diff
@@ -810,6 +863,39 @@ void Description::Entry::parseSdpLine(string_view line) {
...
+				throw std::invalid_argument("Invalid rid line \"" + string(line) + "\" in description");
+			}
+
+			const auto rid = list[0];
+			const auto direction = list[1];
+
+			auto builder = RidBuilder(string(rid));
+
+			if (list.size() == 3) {
```
paullouisageneau:
```suggestion
			if (list.size() >= 3) {
```

paullouisageneau (commented):
Thank you for implementing!

src/description.cpp
kmansoft: Actually no, there can be more than one attribute (key/value pair) and we need to iterate them all. The way the loop is written is it does that.

src/description.cpp
kmansoft: Sure, I don't see a problem with that

src/description.cpp
kmansoft: Sure

src/description.cpp
kmansoft: Early break would prevent the subsequent code block from executing - and that's the one that actually extracts the key/value pair

paullouisageneau: Oh sorry, my bad, the last value has no separator, obviously.

paullouisageneau (approved):
Looking good, thanks!

PR #1542, opened 2026-04-14
Export sync timestamps in C API.

eerimoq:


src/capi.cpp
```diff
@@ -1552,6 +1571,18 @@ int rtcGetLastTrackSenderReportTimestamp(int id, uint32_t *timestamp) {
 	});
 }
 
+int rtcGetTrackSyncTimestamps(int id, uint64_t *rtpTimestamp, uint64_t *ntpTimestamp) {
```
paullouisageneau: I would call it `rtcGetTrackRtcpSyncTimestamps` to make it clear that it requires a track id but the timestamps are from the `RtcpReceivingSession`. Also, `id` should be called `tr`.

src/capi.cpp
```diff
@@ -263,6 +267,20 @@ void emplaceRtpConfig(shared_ptr<RtpPacketizationConfig> ptr, int tr) {
 	rtpConfigMap.emplace(std::make_pair(tr, ptr));
 }
 
+shared_ptr<RtcpReceivingSession> getRtcpReceivingSession(int id) {
```
paullouisageneau: `id` should be called `tr` instead.

src/capi.cpp
eerimoq: `rtcGetLastTrackSenderReportTimestamp()` also takes an `id` and does not have RTCP in the function name.

paullouisageneau: Indeed, it should be `tr`. But it has `SenderReport` in the name which makes it clear enough that it requires an SR reporter.

src/capi.cpp
```diff
@@ -263,6 +267,20 @@ void emplaceRtpConfig(shared_ptr<RtpPacketizationConfig> ptr, int tr) {
...
+	std::lock_guard lock(mutex);
+	if (auto it = rtcpReceivingSessionMap.find(id); it != rtcpReceivingSessionMap.end()) {
+		return it->second;
+	} else {
+		throw std::invalid_argument("RTCP receiving session ID does not exist");
+	}
+}
+
+void emblaceRtcpReceivingSession(shared_ptr<RtcpReceivingSession> ptr, int tr) {
```
paullouisageneau:
```suggestion
void emplaceRtcpReceivingSession(shared_ptr<RtcpReceivingSession> ptr, int tr) {
```

src/capi.cpp
eerimoq: Fixed.

src/capi.cpp
eerimoq: Fixed.

src/capi.cpp
eerimoq: Fixed.

paullouisageneau (approved):
Thanks!

PR #1550, opened 2026-04-22
Implement RFC 5104 RTCP FIR support

aronrosenberg:
Implement support for RTCP FIR requests as defined in RFC 5104 Section 4.3.1. Payload type 206, FMT 4

* Hookup to the existing onPli() callback handler.
* In PeerConnection::dispatchMedia parse and look for these FIRs since we need to extract one or more target SSRCs from the FCI payload portion for it to match to a media line correctly and get passed to the handler.
* Add `ccm fir` to the rtcp-fb section of each video Media description
* In the requestKeyframe() pathway, check if RFC5104 is supported by looking for the rtcp-fb `ccm fir` value and use this to send RFC 5104 compliant FIR's instead of PLI.
* Augment the existing requestKeyFrame pathway to take an optional SSRC that could be used in the future to allow for proper rid based targeting of the per rid SSRC in a spatial or temporal layer.
* Fix a bug in the existing RtcpFir implementation where the setMediaSourceSSRC was incorrectly set to the target SSRC instead of 0
* Add a helper to Description::Media::RtpMap called hasFeedback(std::string) to search if that string is found in the rtcp-fb lines.
* Update API and internals to allow for multiple SSRCs per FIR to be sent as the RFC allows that.
* Expose ability to send retransmits for FIR's that don't increment the sequence number


src/impl/peerconnection.cpp
```diff
@@ -584,7 +584,17 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
...
+        if rtcpfb->header.reportCount() == 4) {
+          // RFC 5104 FIR (PT=206, FMT=4): the target SSRC is inside the FCI payload
+					auto parsed = reinterpret_cast<const RtcpFir *>(header);
+					// FIRs have one FCI entry of fixed length
+					size_t pktLen = header->lengthInBytes();
+					if (pktLen == (sizeof(RtcpFbHeader) + sizeof(RtcpFirPart))) {
+						ssrcs.insert(parsed->messageSSRC());
+					}
+				} else if (header->reportCount() == 15 && header->lengthInBytes() >= sizeof(RtcpRemb)) {
+          // REMB
```
aronrosenberg:
```suggestion

				if rtcpfb->header.reportCount() == 4) {
					// RFC 5104 FIR (PT=206, FMT=4): the target SSRC is inside the FCI payload
					auto parsed = reinterpret_cast<const RtcpFir *>(header);
					// FIRs have one FCI entry of fixed length
					size_t pktLen = header->lengthInBytes();
					if (pktLen == (sizeof(RtcpFbHeader) + sizeof(RtcpFirPart))) {
						ssrcs.insert(parsed->messageSSRC());
					}
				} else if (header->reportCount() == 15 && header->lengthInBytes() >= sizeof(RtcpRemb)) {
					// REMB
```

src/impl/peerconnection.cpp
```diff
@@ -597,7 +607,8 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 						continue;
 					}
 				}
-				ssrcs.insert(rtcpfb->mediaSourceSSRC());
+        if (rtcpfb->mediaSourceSSRC() != 0)
```
aronrosenberg:
```suggestion
				if (rtcpfb->mediaSourceSSRC() != 0)
```

src/impl/peerconnection.cpp
```diff
@@ -597,7 +607,8 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 						continue;
 					}
 				}
-				ssrcs.insert(rtcpfb->mediaSourceSSRC());
+				if (rtcpfb->mediaSourceSSRC() != 0)
+				  ssrcs.insert(rtcpfb->mediaSourceSSRC());
```
aronrosenberg:
```suggestion
					ssrcs.insert(rtcpfb->mediaSourceSSRC());
```

src/impl/peerconnection.cpp
```diff
@@ -597,7 +607,8 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 						continue;
 					}
 				}
-				ssrcs.insert(rtcpfb->mediaSourceSSRC());
+				if (rtcpfb->mediaSourceSSRC() != 0)
```
aronrosenberg:
```suggestion
				if (rtcpfb->mediaSourceSSRC())
```

src/impl/peerconnection.cpp
```diff
@@ -625,7 +625,13 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 						if (remb->hasValidId())
 							for (int i = 0; i < remb->getSSRCCount(); ++i)
 								ssrcs.insert(remb->getSSRC(i));
-					}
+					}	else if (header->payloadType() == 206 && rtcpfb->header.reportCount() == 4 &&
```
aronrosenberg:
```suggestion
					} else if (header->payloadType() == 206 && rtcpfb->header.reportCount() == 4 &&
```

aronrosenberg:
@paullouisageneau This PR is ready for your review, we have completed all interop testing.

src/plihandler.cpp
```diff
@@ -28,11 +28,31 @@ void PliHandler::incoming(message_vector &messages, [[maybe_unused]] const messa
...
 				uint8_t feedback_message_type = header->reportCount();
 				if (feedback_message_type == 1) {
 					mOnPli();
 					break;
+				} else if (feedback_message_type == 4) {
+					auto fir = reinterpret_cast<const RtcpFir *>(message->data() + offset);
+					bool invokeHandler = false;
+					unsigned int fciCount = fir->getFciCount();
+					for (unsigned int i = 0; i < fciCount; i++) {
+						uint8_t firSeqNo = fir->parts[i].getSeqNo();
```
paullouisageneau: You must ensure `offset + header->lengthInBytes() <= message->size()` at the beginning of the loop, otherwise the message may be truncated and this may read past the end. 

You could also check that `length >= sizeof(RtcpFir)` before casting for the sake of consistency, even if it's not necessary here given `RtcpFir` is just the header and FCIs, otherwise it looks like an unchecked cast.

src/plihandler.cpp
```diff
@@ -28,11 +28,31 @@ void PliHandler::incoming(message_vector &messages, [[maybe_unused]] const messa
...
+						// RFC 5104 Section 4.3.1.1 says seqNo checking is done against the combo of the two values,
+						// so pack the sender SSRC and target SSRC into a single value for lookup in the map.
+						uint64_t comboSSRCs = (((uint64_t) fir->header.packetSenderSSRC()) << 32) + fir->parts[i].getSSRC();
```
paullouisageneau:
```suggestion
						uint64_t comboSSRCs = (uint64_t(fir->header.packetSenderSSRC()) << 32) + fir->parts[i].getSSRC();
```

src/track.cpp
```diff
@@ -66,15 +66,19 @@ void Track::chainMediaHandler(shared_ptr<MediaHandler> handler) {
...
+	// only request key frames for video
 	if (description().type() == "video")
 		if (auto handler = impl()->getMediaHandler())
-			return handler->requestKeyframe([this](message_ptr m) { impl()->transportSend(m); });
+			return handler->requestKeyframe(targetSSRCs, retransmit, [this](message_ptr m) { impl()->transportSend(m); });
 
 	return false;
 }
 
+bool Track::requestKeyframe(const SSRC ssrc, bool retransmit) {
```
paullouisageneau:
```suggestion
bool Track::requestKeyframe(SSRC ssrc, bool retransmit) {
```

src/impl/peerconnection.cpp
```diff
@@ -625,7 +625,15 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 						if (remb->hasValidId())
 							for (int i = 0; i < remb->getSSRCCount(); ++i)
 								ssrcs.insert(remb->getSSRC(i));
-					}
+					} else if (header->payloadType() == 206 && rtcpfb->header.reportCount() == 4 &&
+						length >= (sizeof(RtcpFbHeader) + sizeof(RtcpFirFci))) {
```
```suggestion
						length >= sizeof(RtcpFir)) {
```

include/rtc/rtp.hpp
```diff
@@ -300,20 +300,23 @@ struct RTC_CPP_EXPORT RtcpPli {
...
+	[[nodiscard]] SSRC getSSRC() const;
+	[[nodiscard]] uint8_t getSeqNo() const;
 };
 
 struct RTC_CPP_EXPORT RtcpFir {
 	RtcpFbHeader header;
-	RtcpFirPart parts[1];
-
-	static unsigned int Size();
+	RtcpFirFci parts[1];
```
paullouisageneau: I think this could be renamed `_fcis` and you could add a `const RtcpFirFci *getFci(int num) const` method to access FCIs with bound checking.

include/rtc/rtp.hpp
```diff
@@ -300,20 +300,23 @@ struct RTC_CPP_EXPORT RtcpPli {
...
 
-	void preparePacket(SSRC messageSSRC, uint8_t seqNo);
+	unsigned int getFciCount() const;
```
paullouisageneau: This could return `int` like `getSSRCCount`.

include/rtc/rtp.hpp
```diff
@@ -300,20 +300,23 @@ struct RTC_CPP_EXPORT RtcpPli {
...
+	[[nodiscard]] static size_t SizeWithFCIs(size_t count);
```
paullouisageneau: For naming consistency, this should be named `SizeWithFcis` and take an `int count`.

include/rtc/rtcpreceivingsession.hpp
```diff
@@ -82,6 +84,8 @@ class RTC_CPP_EXPORT RtcpReceivingSession : public MediaHandler {
 	SSRC mRtxPrimarySsrc = 0;
 	bool mRtxEnabled = false;
 
+	bool mSupportsRfc5104FIR = false;
+	std::unordered_map<SSRC, uint32_t> mRfc5104FIRCmdNums; // SSRC -> seq no mapping
```
paullouisageneau: Nitpick, but I think they should be named `mSupportsRfc5104Fir` and `mRfc5104FIRCmdNums` for consistency.

src/rtp.cpp
```diff
@@ -657,14 +657,25 @@ void RtcpPli::preparePacket(SSRC messageSSRC) {
...
+	return unsigned((header.header.lengthInBytes() - sizeof(RtcpFbHeader)) / sizeof(RtcpFirFci));
+}
+
+void RtcpFir::preparePacket(SSRC senderSSRC, std::vector<RtcpFirFci> firFcis) {
+	header.header.prepareHeader(206, 4, uint16_t((offsetof(RtcpFir, parts) / sizeof(uint32_t)) + (2 * firFcis.size()) - 1));
+	header.setPacketSenderSSRC(senderSSRC);
+	header.setMediaSourceSSRC(0);
+	unsigned int offset = 0;
+	for (std::vector<RtcpFirFci>::iterator it = firFcis.begin(); it != firFcis.end(); ++offset, ++it) {
+		parts[offset]._ssrc = htonl(it->_ssrc);
```
paullouisageneau: This looks weird as it means the field must be in host bit order here while it will be in network bit order when reading it. IMHO there should be a method to fill the fields, you should not have to access the fields directly.

Sum1cares (approved):

src/rtcpreceivingsession.cpp
```diff
@@ -276,19 +284,55 @@ void RtcpReceivingSession::pushRR(const message_callback &send, unsigned int las
...
+			SSRC targetSSRC = ssrc;
+			if (targetSSRC == 0)
+				targetSSRC = mSsrc;
+			if(!retransmit)
+				++mRfc5104FIRCmdNums[targetSSRC];
+
+			RtcpFirFci firFci;
+			firFci._seqNo = static_cast<uint8_t>(mRfc5104FIRCmdNums[targetSSRC] % 256);
+			firFci._ssrc = targetSSRC;
+			firFci._dummy1 = 0; firFci._dummy2 = 0;
```
paullouisageneau: There could be a `preparePacket` method to do this.

paullouisageneau (commented):
Thanks for implementing! I've a few comments about naming and API consistency. Also, I think proper checking is missing in `PliHandler::incoming`.

src/plihandler.cpp
aronrosenberg: I rewrote the logic so the size check is more explicit. I was inside the getFciCount() before and is now at the initial RTCP area.

include/rtc/rtp.hpp
aronrosenberg: Changed

include/rtc/rtp.hpp
aronrosenberg: Changed to your suggestion

include/rtc/rtcpreceivingsession.hpp
aronrosenberg: Changed

src/rtp.cpp
aronrosenberg: Changed the structure so that it builds like the other senders do.

src/rtcpreceivingsession.cpp
aronrosenberg: Added preparePacket and moved the logic there.

paullouisageneau (approved):
Thank you, it looks good to me.

PR #1551, opened 2026-04-22
Update code to work with OpenSSL 4.0 which deprecated some functions.

aronrosenberg:
OpenSSL 4.0 deprecates SSL_set1_host in favor of SSL_set1_dnsname, so put a conditional version check and use the new one if using 4.0 or newer

* CI on Windows switched to using OpenSSL 4.0 since it got released and this fails all builds.


aronrosenberg:
@paullouisageneau Can you approve and merge this ASAP. It fixes a CI failure on Windows that is causing all other PRs to fail their windows tests since the release of OpenSSL 4.0 this week.

paullouisageneau (approved):
Thank you for the fix !

PR #1554, opened 2026-04-25
Fix race between DTLS ClientHello and registerIncoming()

arvidn:
## Problem

When two peers connect, the DTLS client can send its ClientHello before the remote peer has registered its incoming callback, causing the packet to be silently dropped. The connection then stalls until the DTLS retransmit timer fires (~1 second later).

This was discovered while building a fuzzer for RTC connections in libtorrent. In the fuzzer the message is reliably dropped, so every connection attempt incurred a 1-second delay before the handshake could proceed.

## Root cause

Each transport layer is created and started only after the layer below it completes its own handshake. The sequence for a WebRTC connection is:

  1. ICE handshake completes on both peers simultaneously
  2. Each peer's ICE state-change callback fires -> initDtlsTransport() is called -> `DtlsTransport::start()` -> `registerIncoming()` sets `IceTransport::onRecv`

The race lives in (2) ICE connecting on both peers is a concurrent event. The DTLS client peer (the offerer) calls `SSL_do_handshake()` / `gnutls_handshake()` and sends a ClientHello the moment its ICE connects. That ClientHello is a UDP datagram that can arrive at the DTLS server peer's `IceTransport::recv()` before the server peer has had a chance to create its DtlsTransport and call `registerIncoming()`.

## Fix

Queue any message that arrives in `Transport::recv()` when no handler is registered. When `onRecv()` is called, drain the queue before returning. A `std::mutex` guards the queue and the callback pointer together, preventing the A-B/B-A interleaving where a message is queued after the drain has already run.

### DIFF
--- a/src/impl/transport.cpp
+++ b/src/impl/transport.cpp
@@ -38,7 +38,24 @@ void Transport::unregisterIncoming() {
 
 Transport::State Transport::state() const { return mState; }
 
-void Transport::onRecv(message_callback callback) { mRecvCallback = std::move(callback); }
+void Transport::onRecv(message_callback callback) {
+	std::vector<message_ptr> pending;
+	{
+		std::lock_guard lock(mPendingMutex);
+		mRecvCallback = std::move(callback);
+		if (mRecvCallback)
+			pending = std::move(mPendingRecv);
+		else
+			mPendingRecv.clear();
+	}
+	for (auto &msg : pending) {
+		try {
+			mRecvCallback(msg);
+		} catch (const std::exception &e) {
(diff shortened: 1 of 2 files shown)


arvidn:
this seems like a pre-existing failure:
```
src\impl\tlstransport.cpp(625): warning C4996: 'SSL_set1_host': Since OpenSSL 4.0
```

aronrosenberg:
Rebase your branch to latest master fix the OpenSSL 4.0 error.

paullouisageneau (approved):
Thank you for finding and fixing the issue, it looks good to me!

PR #1557, opened 2026-04-28
Add abs-capture-time RTP header extension

auscaster:
## Summary

Adds support for the `abs-capture-time` RTP header extension on the
sending side. Applications can stamp every outgoing frame with the
source's absolute capture wallclock in NTP format; receivers (Chrome,
Firefox, etc.) recover that timestamp via `RTCRtpReceiver` and
`requestVideoFrameCallback`'s `metadata.captureTime`, enabling true
glass-to-glass latency measurement.

Spec:
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/abs-capture-time

## Changes

* `RtpPacketizationConfig` gains `absCaptureTimeId` (the negotiated
  extension id from SDP `extmap`).
* `FrameInfo` gains `optional<uint64_t> absCaptureTimeNtp` carrying a
  per-frame 64-bit NTP timestamp (high 32 bits seconds since the NTP
  epoch, low 32 bits fractional).
* When both are set, `RtpPacketizer` writes the 8-byte (shortened)
  form of the extension on every packet of the frame, accounting for
  it in extension-block size and one-byte / two-byte header
  selection.

The receive side is intentionally out of scope for this PR.

## Wire format

8-byte (shortened) form: 64-bit NTP timestamp, network order. No
estimated capture clock offset. This matches the form Chrome emits
by default for outgoing media and is what Chrome's receiver expects
to surface `captureTime` to JS.

## Behaviour when not used

When `absCaptureTimeId` is `0` (default) the path is fully bypassed:
no extra bytes, no header changes. Existing applications are
unaffected.

## Diff size

49 insertions, 2 deletions across 4 files.

## Build

Verified locally on macOS (Apple clang, OpenSSL 3) with
`-DWARNINGS_AS_ERRORS=ON`. Clean build.

### DIFF
--- a/include/rtc/frameinfo.hpp
+++ b/include/rtc/frameinfo.hpp
@@ -27,6 +27,16 @@ struct RTC_CPP_EXPORT FrameInfo {
 	optional<std::chrono::duration<double>> timestampSeconds;
 
 	bool isKeyFrame = false;	// Set by the application
+
+	/// Absolute capture timestamp in NTP-format (high 32 bits seconds since
+	/// the NTP epoch 1900-01-01 UTC; low 32 bits fractional seconds). When
+	/// set alongside an RtpPacketizationConfig with absCaptureTimeId > 0,
+	/// the RtpPacketizer writes the 8-byte form of the abs-capture-time RTP
+	/// header extension on every packet of this frame. Receivers use this
+	/// to recover the source's capture wallclock and compute glass-to-glass
+	/// latency.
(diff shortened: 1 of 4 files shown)


paullouisageneau (approved):
Thank you for the implementation!

auscaster:
Thanks for merging! Do you have a timeline for the next release?

Sum1cares (approved):

PR #1561, opened 2026-05-07
Fix certificate generation with BoringSSL

soares-sergio:
### Problem
When building `libdatachannel` in an environment that uses **BoringSSL** instead of standard OpenSSL (such as within the Matter/Project CHIP ecosystem), `Certificate::Generate` fails with a `runtime_error: Unable to set certificate public key`.

This seems to happen because `X509_set_pubkey` is called very early in the generation process, before setting the certificate version, serial number, and validity times. BoringSSL seems to be stricter and rejects setting the public key on a certificate that is not sufficiently initialized.

### Solution
Move the call to `X509_set_pubkey` to just before `X509_sign`, after all other properties (validity, version, serial, subject/issuer names) have been set. This makes the sequence compatible with BoringSSL while remaining compatible with standard OpenSSL.

### Testing / Reproduction
1. Build `libdatachannel` linking against BoringSSL.
2. Call `rtc::impl::Certificate::Generate(rtc::CertificateType::Ecdsa, "test")`.
3. Without this fix, it throws an exception. With this fix, it succeeds and returns a valid certificate.

### DIFF
--- a/src/impl/certificate.cpp
+++ b/src/impl/certificate.cpp
@@ -510,9 +510,6 @@ Certificate Certificate::Generate(CertificateType type, const string &commonName
 	auto *commonNameBytes =
 	    reinterpret_cast<unsigned char *>(const_cast<char *>(commonName.c_str()));
 
-	if (!X509_set_pubkey(x509.get(), pkey.get()))
-		throw std::runtime_error("Unable to set certificate public key");
-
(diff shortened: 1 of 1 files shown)


paullouisageneau (approved):
Thank you for the fix!

PR #1562, opened 2026-05-09
Implement RTCP App packet sending and receiving

aronrosenberg:
Implementation for sending and receiving RTCP APP packets (RTCP Payload Type 204).

* Add RtcpAppHandler which allows for receiving inbound RTCP APP packets and passes up the name, subtype and any data payload
* Create Track::sendRtcpApp which allows for sending RTCP APP packets for any name, subtype and data payload.

Interop tested with an internal Apple implementation of RTCP APP packets.

### DIFF
--- a/include/rtc/apphandler.hpp
+++ b/include/rtc/apphandler.hpp
@@ -19,16 +19,16 @@ namespace rtc {
 
 /// Handles RTCP APP packets (payload type 204) as defined in RFC 3550 section 6.7.
 class RTC_CPP_EXPORT AppHandler final : public MediaHandler {
-	rtc::synchronized_callback<string, uint8_t, binary> mOnApp;
+	rtc::synchronized_callback<const RtcpAppName, uint8_t, binary> mOnApp;
 
 public:
 	/// Constructs the AppHandler to notify when RTCP APP packets are received.
 	/// @param onApp Callback invoked with (name, subtype, application-dependent data)
-	AppHandler(std::function<void(string name, uint8_t subtype, binary data)> onApp);
+	AppHandler(std::function<void(const RtcpAppName name, uint8_t subtype, binary data)> onApp);
 
 	void incoming(message_vector &messages, const message_callback &send) override;
 
-	bool sendRtcpApp(SSRC ssrc, const char name[4], uint8_t subtype, const binary &data,
+	bool sendRtcpApp(SSRC ssrc, const RtcpAppName &name, uint8_t subtype, const binary &data,
 	                 const message_callback &send) override;
 };
 
--- a/include/rtc/apphandler.hpp
+++ b/include/rtc/apphandler.hpp
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2026 Apple Inc
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef RTC_APP_HANDLER_H
+#define RTC_APP_HANDLER_H
+
+#if RTC_ENABLE_MEDIA
+
+#include "mediahandler.hpp"
+#include "rtp.hpp"
+#include "utils.hpp"
+
+namespace rtc {
+
+/// Handles RTCP APP packets (payload type 204) as defined in RFC 3550 section 6.7.
+class RTC_CPP_EXPORT AppHandler final : public MediaHandler {
+	rtc::synchronized_callback<string, uint8_t, binary> mOnApp;
+
+public:
+	/// Constructs the AppHandler to notify when RTCP APP packets are received.
+	/// @param onApp Callback invoked with (name, subtype, application-dependent data)
+	AppHandler(std::function<void(string name, uint8_t subtype, binary data)> onApp);
+
+	void incoming(message_vector &messages, const message_callback &send) override;
+
+	bool sendRtcpApp(SSRC ssrc, const char name[4], uint8_t subtype, const binary &data,
+	                 const message_callback &send) override;
+};
+
+} // namespace rtc
+
+#endif // RTC_ENABLE_MEDIA
+
+#endif // RTC_APP_HANDLER_H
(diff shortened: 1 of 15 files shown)


include/rtc/apphandler.hpp
```diff
@@ -0,0 +1,39 @@
...
+#if RTC_ENABLE_MEDIA
+
+#include "mediahandler.hpp"
+#include "rtp.hpp"
+#include "utils.hpp"
+
+namespace rtc {
+
+/// Handles RTCP APP packets (payload type 204) as defined in RFC 3550 section 6.7.
+class RTC_CPP_EXPORT AppHandler final : public MediaHandler {
```
paullouisageneau: IMHO it should be called `RtcpAppHandler` to make it more explicit and consistent with other handlers.

src/impl/peerconnection.cpp
```diff
@@ -629,11 +629,17 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
 				}
 				break;
 
+			case 204: // APP
+				if (length >= sizeof(RtcpHeader) + sizeof(SSRC) + 4) {
```
```suggestion
				if (length >= RtcpApp::SizeWithData(0)) {
```

src/track.cpp
```diff
@@ -83,6 +83,14 @@ bool Track::requestBitrate(unsigned int bitrate) {
 	return false;
 }
 
+bool Track::sendRtcpApp(uint32_t ssrc, const RtcpAppName &name, uint8_t subtype, const binary &data) {
+	if (auto handler = impl()->getMediaHandler())
+		return handler->sendRtcpApp(ssrc, name, subtype, data,
+		                            [this](message_ptr m) { impl()->transportSend(m); });
```
paullouisageneau: Do you really need to delegate sending the RTCP app packet to a handler in the chain? The only reason to do this for REMB and PLI is that the receiving session sets the correct SSRC (it would actually be better to get the SSRC from track metadata instead). Here, you don't need any context, so it looks like you could send the packet directly in `Track::sendRtcpApp`.

paullouisageneau (commented):
Thank you for adding this! I think adding `MediaHandler::sendRtcpApp` is probably not useful in this case, it could be simpler.

include/rtc/apphandler.hpp
aronrosenberg: Renamed class and file to `RtcpAppHandler`

src/track.cpp
aronrosenberg: Removed the path going through the Handler.

src/rtcpapphandler.cpp
```diff
@@ -0,0 +1,63 @@
...
+				size_t minSize = RtcpApp::SizeWithData(0); // header + SSRC + name
+				if (length < minSize)
+					break;
+
+				auto app = reinterpret_cast<RtcpApp *>(message->data() + offset);
+				uint8_t subtype = app->subtype();
+				RtcpAppName name = app->name();
+				size_t dataLen = app->dataSize();
+				binary data(message->data() + offset + minSize,
+				            message->data() + offset + minSize + dataLen);
```
paullouisageneau: Why not use `app->data()` here?

```suggestion
				binary data(app->data(), app->data() + dataLen);
```

src/track.cpp
```diff
@@ -83,6 +85,22 @@ bool Track::requestBitrate(unsigned int bitrate) {
...
+	// RTCP APP packets carry an explicit SSRC, so no handler context is needed: build and send
+	// the packet directly.
+	size_t packetSize = RtcpApp::SizeWithData(data.size());
+	auto message = make_message(packetSize, Message::Control);
+
+	auto app = reinterpret_cast<RtcpApp *>(message->data());
+	app->preparePacket(ssrc, name, subtype, data.size());
+
+	if (!data.empty())
+		std::memcpy(app->_data, data.data(), data.size());
```
paullouisageneau, line 99: You should add a `setData()` method for this, which could check the size against `dataSize()`.

src/rtp.cpp
```diff
@@ -725,6 +725,42 @@ bool RtcpNack::addMissingPacket(unsigned int *fciCount, uint16_t *fciPID, uint16
...
+	uint16_t lengthField =
+	    uint16_t((sizeof(SSRC) + 4 + dataLength) / 4); // in 32-bit words, minus 1 for header
+	header.prepareHeader(204, subtype, lengthField);
+	setSSRC(ssrc);
+	setName(name);
+}
+
+void RtcpApp::setSSRC(SSRC ssrc) { _ssrc = htonl(ssrc); }
+
+void RtcpApp::setName(const RtcpAppName &name) { std::memcpy(_name, name.data(), 4); }
```
paullouisageneau, line 756: This is unsafe, you should check `name.size()`. You could throw `invalid_argument` if it is not 4.

paullouisageneau (commented):
Thank you for addressing everything, I have a few more comments then I think it's ready to be merged.

src/rtp.cpp
aronrosenberg: I can add, but doesn't the strict typing of RtcpAppName being a `std::array<char, 4>` mean its already guaranteed to always have the space?

src/track.cpp
aronrosenberg: Not sure we want to expose a setData() method to overwrite the packet data here. The use of make_message already guarantees the correct amount of memory space is there since the struct type uses the char array byte pointer syntax. Exposing setData could lead to heap overflows if used outside this function context.

src/rtp.cpp
paullouisageneau: Indeed, I missed that. My bad, it's all good.

src/track.cpp
paullouisageneau: OK, fair enough!

src/rtcpapphandler.cpp
```diff
@@ -0,0 +1,62 @@
...
+			if (header->payloadType() == 204) {
+				size_t minSize = RtcpApp::SizeWithData(0); // header + SSRC + name
+				if (length < minSize)
+					break;
+
+				auto app = reinterpret_cast<RtcpApp *>(message->data() + offset);
+				uint8_t subtype = app->subtype();
+				RtcpAppName name = app->name();
+				size_t dataLen = app->dataSize();
+				binary data(app->data(), app->data() + dataLen);
```
paullouisageneau: Sorry, `app->data()` returns `const char *` so you need to cast:
```suggestion
				auto dataBytes = reinterpret_cast<const byte *>(app->data());
				binary data(dataBytes, dataBytes + dataLen);
```

paullouisageneau (approved):

Issue #545, opened 2022-01-25
ICE and DTLS restart support


paullouisageneau:
ICE and DTLS restart support should be implemented.

Requires https://github.com/paullouisageneau/libjuice/issues/130

baerrus:
+1

sinFunc:
hello, I total agree with that. i have been considering if there is a interface to set the timeout of heart beat between tow peers on any time.

paullouisageneau:
> hello, I total agree with that. i have been considering if there is a interface to set the timeout of heart beat between tow peers on any time.
This is mostly unrelated to timeouts. If there are datachannels, you can set the SCTP hearbeat interval with `heartbeatInterval` in `SetSctpSettings()` before creating the peer connection. You can also implement your own applicative timeouts.

paullouisageneau:
> hello, I total agree with that. i have been considering if there is a interface to set the timeout of heart beat between tow peers on any time.

ICE restart support is unrelated to exposing timeout settings.

Note if there are datachannels, you can set the SCTP hearbeat interval with `heartbeatInterval` in `SetSctpSettings()` before creating the peer connection. You can also implement your own applicative timeouts.

PR #1568, opened 2026-05-22
Expose ICE restart through the C API

pvizeli:
## Motivation

Long-running PeerConnections (federation transports, P2P signalling channels) need to recover from `ICEState::FAILED` without tearing down the DTLS/SCTP layers above. The W3C-spec way is an ICE restart — re-run ICE on the existing PC while leaving DTLS and SCTP intact — but today neither path through libdatachannel's C API can trigger one.

Two structural gaps in the current code:

1. **No way to pass fresh ICE credentials.** `rtcSetLocalDescription(int pc, const char *type)` (`src/capi.cpp:564`) doesn't plumb `LocalDescriptionInit::iceUfrag` / `icePwd`. The second call regenerates the SDP from the *existing* `iceTransport`, so the new offer carries the **same** ufrag/pwd as the first one — that isn't an ICE restart, just re-signalling.
2. **No re-gather, so `GatheringState::Complete` never re-fires.** `setLocalDescription` only calls `gatherLocalCandidates` when `gatheringState == GatheringState::New` (`src/peerconnection.cpp:165`). After the first connect, that's `Complete`, so no new gather cycle. `changeGatheringState` early-returns when the state is unchanged, so the gathering-state callback never re-fires either. Any caller waiting on a fresh gather-complete hangs forever.

## Changes

* **New C API `rtcSetLocalDescriptionEx(pc, type, iceUfrag, icePwd)`** — same shape as `rtcSetLocalDescription` plus two optional credential parameters. Non-NULL `iceUfrag` + `icePwd` plumb through `LocalDescriptionInit` to the existing C++ `setLocalDescription(type, init)` path. NULL credentials retain the original `rtcSetLocalDescription` semantics.
* **`setLocalDescription` now re-runs ICE gathering when the caller supplies fresh credentials**, even if the previous gathering cycle reached `Complete`. Rolls the gathering state back to `InProgress` (so the `Complete` callback fires again at the end of the new cycle) and calls `gatherLocalCandidates`.

No changes to existing C API signatures. `rtcSetLocalDescription` callers see no behaviour change; ICE-restart support is opt-in via the new `Ex` function.

## Test plan

- [ ] Build with the new function compiled in (`-DUSE_SYSTEM_OPENSSL`, default cmake config).
- [ ] Loopback PC pair: handshake → CONNECTED → call `rtcSetLocalDescriptionEx(pc, "offer", fresh_ufrag, fresh_pwd)` → the new SDP carries `fresh_ufrag` / `fresh_pwd` (different from the original).
- [ ] `GatheringState::InProgress` callback fires on restart; `Complete` callback fires again at the end of the new cycle.
- [ ] DataChannel established before the restart continues to carry traffic afterwards (SCTP layer survives).

## Downstream usage

[social-home-io/aiolibdatachannel](https://github.com/social-home-io/aiolibdatachannel) is the immediate consumer — async Python wrapper that exposes `PeerConnection.restart_ice()` on top of this new C API.

### DIFF
--- a/include/rtc/rtc.h
+++ b/include/rtc/rtc.h
@@ -213,6 +213,14 @@ RTC_C_EXPORT int rtcSetGatheringStateChangeCallback(int pc, rtcGatheringStateCal
 RTC_C_EXPORT int rtcSetSignalingStateChangeCallback(int pc, rtcSignalingStateCallbackFunc cb);
 
 RTC_C_EXPORT int rtcSetLocalDescription(int pc, const char *type); // type may be NULL
+// Same as rtcSetLocalDescription, but allows supplying fresh ICE
+// credentials. When both iceUfrag and icePwd are non-NULL on a
+// previously-connected PC, the call performs an ICE restart: the new
+// credentials replace the existing ones and a fresh gathering cycle
+// is kicked off. Passing NULL for either of iceUfrag/icePwd is
+// equivalent to rtcSetLocalDescription. type may be NULL.
+RTC_C_EXPORT int rtcSetLocalDescriptionEx(int pc, const char *type, const char *iceUfrag,
+                                          const char *icePwd);
 RTC_C_EXPORT int rtcSetRemoteDescription(int pc, const char *sdp, const char *type);
 RTC_C_EXPORT int rtcAddRemoteCandidate(int pc, const char *cand, const char *mid);
 
--- a/src/capi.cpp
+++ b/src/capi.cpp
@@ -570,6 +570,22 @@ int rtcSetLocalDescription(int pc, const char *type) {
 	});
 }
 
+int rtcSetLocalDescriptionEx(int pc, const char *type, const char *iceUfrag,
+                             const char *icePwd) {
+	return wrap([&] {
+		auto peerConnection = getPeerConnection(pc);
+		LocalDescriptionInit init;
+		if (iceUfrag)
+			init.iceUfrag = std::string(iceUfrag);
+		if (icePwd)
+			init.icePwd = std::string(icePwd);
+		peerConnection->setLocalDescription(type ? Description::stringToType(type)
+		                                         : Description::Type::Unspec,
+		                                    std::move(init));
+		return RTC_ERR_SUCCESS;
+	});
+}
+
 int rtcSetRemoteDescription(int pc, const char *sdp, const char *type) {
 	return wrap([&] {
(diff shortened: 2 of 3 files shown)


pvizeli:
Closing without merging — see comment below.

After landing this patch downstream and trying it end-to-end, I discovered the C API exposure is only half of what ICE restart needs. libjuice itself refuses to update ICE attributes once gathering has started:

\`\`\`c
// deps/libjuice/src/agent.c
int agent_set_local_ice_attributes(juice_agent_t *agent, const char *ufrag, const char *pwd) {
    if (agent->conn_impl) {
        JLOG_WARN(\"Unable to set ICE attributes, candidates gathering already started\");
        return JUICE_ERR_FAILED;
    }
    ...
}
\`\`\`

\`conn_impl\` is set once the agent starts gathering, so any subsequent \`juice_set_local_ice_attributes\` returns \`JUICE_ERR_FAILED\`. \`IceTransport::setIceAttributes\` (libdatachannel side) turns that into \`std::invalid_argument\` → \`RTC_ERR_INVALID\` at the C API boundary. Plus, \`initIceTransport()\` short-circuits with the existing transport (\`std::atomic_load(&mIceTransport)\` returns the same shared_ptr), so the juice agent is never recreated either.

So the patch in this PR is necessary but not sufficient: real ICE restart through the C API also needs either (a) a libjuice patch dropping the \`conn_impl\` guard + resetting the gather pipeline so a fresh round of candidates is emitted under the new credentials, or (b) a libdatachannel patch that destroys + recreates the \`IceTransport\` on restart.

Closing this PR until one of those is in scope. Happy to reopen / rework into a larger change if you have a preferred direction.

Context: downstream consumer is [social-home-io/aiolibdatachannel#14](https://github.com/social-home-io/aiolibdatachannel/issues/14) — async Python wrapper that wanted ICE restart for federation transport recovery from \`ICEState::FAILED\`.

paullouisageneau:
libjuice does not currently support ICE restart, see https://github.com/paullouisageneau/libdatachannel/issues/545

I'm not sure creating a new ICE transport is the correct approach here, I think it makes more sense for libjuice to support ICE restart and perform it automatically. There are multiple reasons for this, for instance it should be possible to restart ICE and gather candidates again while the current connection is still operating normally, while recreating the transport would force full re-establishment and introduce convoluted transport-switching logic in libdatachannel. Also, libnice, which can be used as alternative ICE backend, implements ICE restart. Therefore, your option (a) sounds better to me.

PR #1573, opened 2026-06-02
Fix conflict with external define None

### LINKED ISSUE #1571: Global.hpp fails to compile on certain linux machines
https://github.com/KhronosGroup/OpenCL-CLHPP/issues/169

```
/webrtc/include/rtc/global.hpp
    Line 21: In file included from upload/webrtc/src/main.cpp:11:
In file included from upload/webrtc/src/main.hpp:6:
In file included from upload/webrtc/include/rtc/rtc.hpp:14:
expected identifier
   21 |         None = 0,
      |         ^
    Line 21: expanded from macro 'None'
  115 | #define None                 0L /* universal null resource or null atom */
      |                              ^
```

it complains about None being an invalid identifier in global.hpp since that's the name of a macro defined in some system header (I think from X11 based on [this issue](https://github.com/KhronosGroup/OpenCL-CLHPP/issues/169) ). that's easily fixed by adding #undef None right before defining the LogLevel enum to remove that macro

paullouisageneau:
Fixes https://github.com/paullouisageneau/libdatachannel/issues/1571

### DIFF
--- a/include/rtc/global.hpp
+++ b/include/rtc/global.hpp
@@ -17,6 +17,10 @@
 
 namespace rtc {
 
+#ifdef None
+#undef None
+#endif
+
 enum class LogLevel { // Don't change, it must match plog severity
 	None = 0,
 	Fatal = 1,


Nemirtingas:
I don't think this should've been merged. Conflicts between libraries are unfortunatly common with defines. I think the user should have put libdatachannel to its own translation unit instead of adding an undef in libdatachannel itself. This library is not responsible for other libraries wrong coding habits.

paullouisageneau:
@Nemirtingas I would agree in general, but in this particular case `None` is a rather generic name and it is defined by X11 headers which are commonly used, so I think it doesn't hurt.

Nemirtingas:
Yep, I know, I also use X11 headers, thats why I've put all my X11 code to its own source file, so it doesn't conflict with any other symbols. None is not the only generic word that X11 defines x)

PR #1574, opened 2026-06-02
Draft: Added an Asio interface allowing to customize the task scheduler implementation.

Nemirtingas:


Nemirtingas:
Draft until I fix the tests, but I think this could be of some use to people out there. The example using boost::asio is working.

Issue #1575, opened 2026-06-03
Both Track::sendFrame() overrides return void when outgoing() returns bool, silently dropping errors


Rtificer:
This one is pretty simple.

Both sendFrame functions have the signature

```Cpp
// include/rtc/track.hpp
void sendFrame(binary data, FrameInfo info);
void sendFrame(const byte *data, size_t size, FrameInfo info);
```

But internally call `impl()->outgoing()`
```Cpp
// src/track.cpp
void Track::sendFrame(binary data, FrameInfo info) {
	impl()->outgoing(make_message(std::move(data), std::make_shared<FrameInfo>(std::move(info))));
}

void Track::sendFrame(const byte *data, size_t size, FrameInfo info) {
	sendFrame(binary(data, data + size), std::move(info));
}
```
But outgoing is fallible and thus returns a bool
```Cpp
// src/impl/track.hpp
bool outgoing(message_ptr message);
```

These should be changed to return a bool to avoid silently dropping the error.

paullouisageneau:
The returned `bool` on the send path does not indicate errors, it indicates if the send happened synchronously or not. `Channel::send()` returning `false` typically means [the message was buffered](https://github.com/paullouisageneau/libdatachannel/blob/ed2fb2c73f872c166d2266b11842448db3171eb1/include/rtc/channel.hpp#L28), which is useful for datachannels as it indicates you are sending too fast. Arguably, this is not useful and actually misleading when sending frames (for instance with a `PacingHandler`, every frame is buffered so it always returns `false`).

If there is an error, an exception is thrown, like for other methods.

paullouisageneau:
Closing this since it is a misconception, please reopen if you think there is a good reason for exposing the return value.

Rtificer:
Thanks for the clarification; apologies for that.

PR #1577, opened 2026-06-03
fixed sendFrame signature correctly report errors

Rtificer:
Simply changed the signatures of the Track::sendFrame() overrides to not silently drop errors

See [Issue #1575](https://github.com/paullouisageneau/libdatachannel/issues/1575)


paullouisageneau:
See https://github.com/paullouisageneau/libdatachannel/issues/1575#issuecomment-4617900688

PR #1580, opened 2026-06-06
Add AV1 RTP depacketizer

henryruhs:
# Prolog

# Description

**Add an AV1 RTP depacketizer.**

The library ships an AV1 RTP **packetizer** (`AV1RtpPacketizer`) but no depacketizer, so receiving AV1 over RTP is not currently possible. This adds `AV1RtpDepacketizer`, reassembling OBUs from RTP per the [AV1 RTP spec](https://aomediacodec.github.io/av1-rtp-spec/) — handling the aggregation header, OBU element counts, and fragments continued across packets.

### Added

- `include/rtc/av1rtpdepacketizer.hpp` and `src/av1rtpdepacketizer.cpp` — the `AV1RtpDepacketizer` class.
- Wired into `CMakeLists.txt` and exported via `rtc.hpp`.
- C-API setter: `rtcSetAV1Depacketizer(int tr, rtcObuPacketization obuPacketization)`.
- `DOC.md` update.

### Output mode

The `rtcObuPacketization` parameter mirrors `rtcSetAV1Packetizer` for API symmetry:

- `TemporalUnit` — re-emits each OBU **with** a reconstructed `obu_has_size_field`, producing a standard decodable bitstream.
- `Obu` — emits OBUs as received.

I'm happy to take direction here: an alternative is to always reconstruct the size field (as libwebrtc does) and drop the parameter. I kept it explicit for symmetry with the existing packetizer setter.

### Files

- `include/rtc/av1rtpdepacketizer.hpp` (new)
- `src/av1rtpdepacketizer.cpp` (new)
- `CMakeLists.txt`
- `include/rtc/rtc.hpp`
- `include/rtc/rtc.h`
- `src/capi.cpp`
- `DOC.md`


include/rtc/av1rtpdepacketizer.hpp
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright (c) 2024 Paul-Louis Ageneau
```
paullouisageneau: You should put your name here.

src/av1rtpdepacketizer.cpp
```diff
@@ -0,0 +1,213 @@
+/**
+ * Copyright (c) 2024 Paul-Louis Ageneau
```
paullouisageneau: Same here.

src/av1rtpdepacketizer.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+namespace {
+
+void appendObuWithSize(binary &frame, const std::byte *data, size_t size) {
+	if (size < 1)
+		return;
+
+	uint8_t header = std::to_integer<uint8_t>(data[0]);
+	bool hasExtension = (header >> 2) & 1;
+	bool hasSizeField = (header >> 1) & 1;
```
paullouisageneau: You could add mask constants for bits like in the AV1 packetizer. Additionally, there is no need to convert to `uint8_t`, you can perform the bit operations on bytes directly.

src/av1rtpdepacketizer.cpp
```diff
@@ -0,0 +1,213 @@
...
+		payloadData = data + offset;
+		payloadSize = size - offset;
+		if (existingSize < payloadSize)
+			payloadSize = existingSize;
+	} else {
+		payloadData = data + headerSize;
+		payloadSize = size - headerSize;
+	}
+
+	frame.push_back(std::byte(header | 0x02));
```
paullouisageneau: This sets the size field bit but it's not obvious, it should use a constant.

src/av1rtpdepacketizer.cpp
```diff
@@ -0,0 +1,213 @@
...
+		if (packet->size() <= rtpHeaderSize + paddingSize + 1)
+			continue;
+
+		const std::byte *payload = packet->data() + rtpHeaderSize;
+		size_t payloadSize = packet->size() - rtpHeaderSize - paddingSize;
+
+		if (payloadSize < 1)
+			continue;
+
+		uint8_t aggHeader = std::to_integer<uint8_t>(payload[0]);
```
paullouisageneau: Same, conversion to `uint8_t` is not necessary, you can perform bit operations on `byte`.

paullouisageneau (commented):
It looks OK to me, but hardcoded bits could definitely use constants for readability.

henryruhs:
@paullouisageneau code has been updated

paullouisageneau (approved):
Thanks for the fixes!

PR #1581, opened 2026-06-06
Fix AV1 RTP packetizer handling of OBUs that already carry a size field

henryruhs:
# Prolog

# Description

**Fix AV1 RTP packetizer handling of OBUs that already carry a size field.**

In `AV1RtpPacketizer::extractTemporalUnitObus`, when an input OBU has `obu_has_size_field` set, the previous code copied the OBU range as-is and computed its offsets incorrectly — the OBU extension byte was not accounted for — producing malformed OBU elements.

The fix rebuilds each element and clears `obu_has_size_field`, as the [AV1 RTP spec §4.5](https://aomediacodec.github.io/av1-rtp-spec/) recommends: *"To minimize overhead, the obu_has_size_field flag SHOULD be set to zero in all OBUs."* The element length is instead conveyed by the aggregation header (§4.4) — a LEB128 length prefix per element, with the last element's size implicit.

This change:

- Detects the extension bit and computes the correct header size.
- Clears `obu_has_size_field` in the emitted OBU header (`& ~obuHasSizeMask`).
- Rebuilds the OBU as header (+ optional extension byte) + payload, **dropping the LEB128 size field** (so the OBU carries no length of its own), with a bounds check against the input buffer.

### Files

- `src/av1rtppacketizer.cpp`

### Note

Best paired with a round-trip test (packetize → depacketize) to lock in the behavior.

### DIFF
--- a/src/av1rtppacketizer.cpp
+++ b/src/av1rtppacketizer.cpp
@@ -63,6 +63,10 @@ std::vector<binary> AV1RtpPacketizer::extractTemporalUnitObus(const binary &data
 			return obus;
 		}
 
+		size_t startIndex = index;
+		bool hasExtension = (data.at(index) & obuHasExtensionMask) != byte(0);
+		size_t headerSize = obuHeaderSize + (hasExtension ? 1 : 0);
+
 		if ((data.at(index) & obuHasExtensionMask) != byte(0)) {
 			index++;
 		}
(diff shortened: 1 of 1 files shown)


src/av1rtppacketizer.cpp
```diff
@@ -63,6 +63,10 @@ std::vector<binary> AV1RtpPacketizer::extractTemporalUnitObus(const binary &data
 			return obus;
 		}
 
+		size_t startIndex = index;
+		bool hasExtension = (data.at(index) & obuHasExtensionMask) != byte(0);
+		size_t headerSize = obuHeaderSize + (hasExtension ? 1 : 0);
+
 		if ((data.at(index) & obuHasExtensionMask) != byte(0)) {
```
paullouisageneau, line 70:
```suggestion
		if (hasExtension) {
```

paullouisageneau (approved):

PR #1583, opened 2026-06-06
Add rtcGetPayloadTypesForCodec and fix wildcard rtcp-fb parsing

henryruhs:
# Prolog

# Description

**Add `rtcGetPayloadTypesForCodec` and fix the wildcard `a=rtcp-fb:*` SDP parsing it exposes.**

These two changes belong together: the new helper parses an SDP description, which is exactly what surfaces the wildcard `rtcp-fb` parser bug.

Usecase in our Python implementation:

```python
def get_payload_type(sdp_offer : SdpOffer, codec : AudioCodec | VideoCodec) -> int:
	datachannel_library = datachannel_module.create_static_library()
	payload_type_buffer = (ctypes.c_int * 16)()
	payload_type_total = datachannel_library.rtcGetPayloadTypesForCodec(sdp_offer.encode(), codec.lower().encode(), payload_type_buffer, 16)

	if payload_type_total:
		return payload_type_buffer[0]

	return 0
```

### Added

- `rtcGetPayloadTypesForCodec(const char *sdp, const char *ccodec, int *buffer, int size)` — returns the RTP payload types for a codec parsed from a **raw SDP string**, without requiring a track. It complements the existing `rtcGetTrackPayloadTypesForCodec`, which needs a live track handle.

### Fixed

- Wildcard payload type in `a=rtcp-fb:*` lines. When parsing `a=rtcp-fb:<pt> <feedback>`, `Description::Media::parseSdpLine` runs the payload-type token through `to_integer<int>(...)`. RFC 4585 allows `*` to apply the feedback to **all** formats, but `to_integer<int>("*")` misparses it (and can crash the parser). This handles the wildcard: when the token is `*`, the feedback is appended to every entry in the rtp map; otherwise the per-payload-type behavior is unchanged.

Without this fix, calling `rtcGetPayloadTypesForCodec` on SDP that contains `a=rtcp-fb:*` (common from browsers/SFUs) hits the bug — hence shipping both together.

### Notes

- A small `const` overload of `Description::Entry::extIds()` is included (used by the const SDP-parsing paths); it can be dropped if you'd prefer to keep the PR strictly to the above.

### Files

- `src/capi.cpp`
- `include/rtc/rtc.h`
- `src/description.cpp`
- `include/rtc/description.hpp`

### DIFF
--- a/include/rtc/description.hpp
+++ b/include/rtc/description.hpp
@@ -173,7 +173,7 @@ public:
 			Direction direction = Direction::Unknown;
 		};
 
-		std::vector<int> extIds();
+		std::vector<int> extIds() const;
 		ExtMap *extMap(int id);
 		const ExtMap *extMap(int id) const;
 		void addExtMap(ExtMap map);
@@ -287,6 +287,11 @@ public:
 		void removeRtpMap(int payloadType);
 		void removeFormat(const string &format);
 
+		// Media-level RTCP feedback, serialized with the wildcard payload type ("*")
+		void addFeedback(string fb);
+		void removeFeedback(const string &str);
+		bool hasFeedback(const string &str) const;
+
 		void addRtxCodec(int payloadType, int origPayloadType, unsigned int clockRate);
 
 		optional<int> getRtxPayloadType(int primaryPayloadType) const;
@@ -303,6 +308,7 @@ public:
 
 		std::vector<int> mOrderedPayloadTypes;
 		std::map<int, RtpMap> mRtpMaps;
+		std::vector<string> mRtcpFbs; // media-level feedback (wildcard payload type)
 		std::vector<uint32_t> mSsrcs;
 		std::map<uint32_t, string> mCNameMap;
 		std::map<SSRC, SSRC> mSsrcToRtxSsrc;  // primary_ssrc -> rtx_ssrc
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -752,7 +752,7 @@ void Description::removeAttribute(const string &attr) {
 	    mAttributes.end());
 }
 
-std::vector<int> Description::Entry::extIds() {
+std::vector<int> Description::Entry::extIds() const {
 	std::vector<int> result;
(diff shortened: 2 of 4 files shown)


src/description.cpp
```diff
@@ -1374,12 +1377,20 @@ void Description::Media::parseSdpLine(string_view line) {
...
-			auto it = mRtpMaps.find(pt);
-			if (it == mRtpMaps.end())
-				it = mRtpMaps.insert(std::make_pair(pt, Description::Media::RtpMap(pt))).first;
+			auto ptStr = value.substr(0, p);
+			auto fbValue = value.substr(p + 1);
 
-			it->second.rtcpFbs.emplace_back(value.substr(p + 1));
+			if (ptStr == "*") {
+				for (auto &[_, rtp] : mRtpMaps)
+					rtp.rtcpFbs.emplace_back(fbValue);
```
paullouisageneau: I agree this needs to be fixed but I think it should not be done like that. This transforms the wildcard feedback into non-wildcard feedback for each payload type, meaning that parsing the description and serializing it again would change it, but also that an answer wouldn't reflect the offer, which is incorrect and in violation of the spec for some feedback formats. For instance, RFC8888 says: `The payload type used with "ccfb" feedback MUST be the wildcard type ("*")`

paullouisageneau: A possible approach would be adding a field `rtcpFbs` at media level, with similar methods `addFeedback`, `removeFeedback`, and `hasFeedback`.

include/rtc/description.hpp
```diff
@@ -174,6 +174,7 @@ class RTC_CPP_EXPORT Description {
 		};
 
 		std::vector<int> extIds();
```
paullouisageneau: You may remove the overload and simply add `const` to the existing method.

src/description.cpp
henryruhs: has been integrated (by claude)

include/rtc/description.hpp
henryruhs: has been fixed (by claude)

paullouisageneau (approved):

PR #1593, opened 2026-06-16
perf: reduce internal message handoff overhead

mertushka:
## Summary

Reduce avoidable internal message handoff overhead across data-channel and transport paths.

Changes:

- move `message_ptr` values through one-way internal handoff paths instead of forwarding lvalues
- move `message_variant` into the channel message callback instead of copying it
- make read-only helpers take `const message_ptr&`
- extend the same ownership handoff cleanup to WebSocket, TCP, TLS, HTTP proxy, DTLS, and DTLS-SRTP paths where the message is no longer used afterward

## Rationale

Several internal paths accept a `shared_ptr<Message>` by value and then forward it as an lvalue even though the caller does not use it afterward. Moving ownership through those paths avoids unnecessary `shared_ptr` refcount churn while preserving the existing ownership model.

`Channel::flushPendingMessages()` also copied the received `message_variant` into the callback path. Moving the variant avoids an extra binary payload copy before user callback delivery.

The follow-up transport changes intentionally avoid media fan-out paths where the same message can be delivered to multiple receivers.

### DIFF
--- a/src/impl/channel.cpp
+++ b/src/impl/channel.cpp
@@ -71,7 +71,7 @@ void Channel::flushPendingMessages() {
 			break;
 
 		try {
-			messageCallback(*next);
+			messageCallback(std::move(*next));
 		} catch (const std::exception &e) {
 			PLOG_WARNING << "Uncaught exception in callback: " << e.what();
 		}
--- a/src/impl/datachannel.cpp
+++ b/src/impl/datachannel.cpp
@@ -66,7 +66,7 @@ struct AckMessage {
 
 #pragma pack(pop)
 
-bool DataChannel::IsOpenMessage(message_ptr message) {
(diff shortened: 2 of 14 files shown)


mertushka:
@paullouisageneau Can you review this PR? Thank you.

mertushka:
I noticed similar lvalue-forwarding patterns in a few non-data-channel paths such as WebSocket/TCP/TLS while preparing this patch. I kept this PR scoped to the measured data-channel path to keep the diff small. If this approach looks good, I can follow up separately for those paths.

paullouisageneau (approved):
It looks good, thanks. Feel free to update non-datachannels paths too.

mertushka:
@paullouisageneau This is ready to merge.

paullouisageneau (approved):
Looking good, thank you!

PR #1596, opened 2026-06-20
Implement concurrent WebSocket connection attempts

### LINKED ISSUE #1556: WebSocket client does not attempt all IPs with short timeouts
Hi, I migrated to libdatachannel and since then a few of my application's users experience consistent timeouts when their IPv6 connection has problems, e.g. when `ping -6` always times out but `ping -4` works. I used the [QWebSocket](https://doc.qt.io/qt-6/qwebsocket.html) implementation before and it never had that issue, so I think it's circumventable.

Some ideas:
1. Prefer IPv4 (perhaps make this configurable)
2. Fall back to IPv4 when IPv6 fails (probably difficult because timeouts will take forever)
3. Connect to both at the same time (probably undesirable)

I don't really see a better option than 1. I think that's what Qt is doing too, but I am not sure. In my eyes that's the only way to explain why the WebSocket connection never times out with an older version that uses Qt WebSockets.

Would it be possible to add a switch to prefer IPv4 over IPv6?

Any other ideas are welcome too. Thank you!

Related: https://github.com/ungive/discord-music-presence/issues/623

ungive:
This PR implements concurrent connection attempts for WebSockets and is a direct result of #1556. I implemented this mostly to serve my own needs and I didn't read the full Happy Eyeballs RFCs ([RFC 6555](https://www.rfc-editor.org/info/rfc6555/) and [RFC 8305](https://www.rfc-editor.org/info/rfc8305)), but I did my best to address all comments made in #1556, except "For WebSockets the IP address family should also be cached for later connections.", which I left as a TODO.

Summary (commit message): Connects to addresses concurrently, so that IPv4 and IPv6 addresses are attempted concurrently. This reduces connection latency in cases where the first address is not reachable. Reorders addresses in such a way that IPv4 and IPv6 addresses are interleaved and addresses from both families are attempted as early as possible. Closes all created sockets that are unused, once a connection succeeds or all connections fail. Adds an option to configure the connection attempt delay or to disable concurrent connection attempts altogether. Also adds verbose logging to help with debugging.

Fixes #1556.

So far I've only tested this on Linux. I will add this to my app where it will get some battle testing and I will also test this by hand on Windows and Mac. I'll report back once I got the results of that.

I'm certain there are things that need to be adapted or change, I'd appreciate any feedback!

### DIFF
--- a/src/impl/tcptransport.cpp
+++ b/src/impl/tcptransport.cpp
@@ -248,9 +248,21 @@ void TcpTransport::attempt() {
 		if (state() != State::Connecting)
 			return; // Cancelled
 
+		// Handle the rare case where the poll service emits two events in quick succession, while
+		// concurrent connection attempts are enabled: A delayed connection attempt timeout event
+		// that triggers another connection attempt asynchronously on the thread pool, followed by
+		// an out event for a successful connection. The latter acquires the lock and closes all
+		// sockets, then releases the lock before it changes the state to connected (see the
+		// processConnect() method for reasoning). If this happens before the socket for the queued
+		// attempt is created here, which is likely because the thread pool takes some time to
+		// execute the next attempt, we would create a new socket while we have already established
+		// a connection with a socket (race condition). This check prevents the creation of another
+		// connection in this case, since access to mSock is synchronized via mSendMutex.
+		if (mSock != INVALID_SOCKET)
+			return;
+
 		if (mResolved.empty()) {
 			PLOG_WARNING << "Connection to " << mHostname << ":" << mService << " failed";
-			assert(mSock == INVALID_SOCKET);
 			closeUnusedSockets();
 			changeState(State::Failed);
 			return;
@@ -586,6 +598,11 @@ void TcpTransport::processConnect(PollService::Event event, socket_t sock, bool 
 				throw std::runtime_error(msg.str());
 			}
 
+			if (mSock != INVALID_SOCKET) {
+				assert(false);
+				throw std::logic_error("Already have a valid socket");
+			}
+
 			// Use this socket and close all other opened sockets
 			mSock = sock;
 			closeUnusedSockets();
@@ -594,6 +611,10 @@ void TcpTransport::processConnect(PollService::Event event, socket_t sock, bool 
 		// Success
 		PLOG_INFO << "TCP connected";
 		PLOG_VERBOSE << "Using socket with descriptor " << mSock;
+
+		// We must change the connected state after releasing the mSendMutex lock, since the
+		// connected state change handler might e.g. initialize TLS, which would cause a resource
+		// deadlock, if we were still holding the lock to mSendMutex.
 		changeState(State::Connected);
 		setPoll(PollService::Direction::In);
 
--- a/src/impl/tcptransport.cpp
+++ b/src/impl/tcptransport.cpp
@@ -225,8 +225,6 @@ void TcpTransport::resolve() {
 
 		freeaddrinfo(result);
 
-		mPendingAddresses = static_cast<int>(mResolved.size());
-
 	} catch (const std::exception &e) {
 		PLOG_WARNING << e.what();
 		changeState(State::Failed);
@@ -269,6 +267,7 @@ void TcpTransport::attempt() {
 		// reuse closed file descriptors
 		assert(mSocks.find(sock) == mSocks.end());
 		mSocks.insert(sock);
+		mPendingSocks += 1;
 
 	} catch (const std::runtime_error &e) {
 		PLOG_DEBUG << e.what();
@@ -521,7 +520,7 @@ void TcpTransport::closeUnusedSockets() {
 	}
 
 	mSocks.clear();
-	mPendingAddresses = 0;
+	mPendingSocks = 0;
 }
 
 void TcpTransport::processConnect(PollService::Event event, socket_t sock, bool isDelayTimeout) {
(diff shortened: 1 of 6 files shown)


ungive:
Some comments about how I tested: I used `test/websocket.cpp` and `test/websocketserver.cpp` for testing and when `wss://echo.websocket.org` didn't work, I used `wss://ws.ifelse.io`, which also comes with 2 IPv4 and 2 IPv6 addresses, which is good for testing with more than just 1 IPv4 and IPv6 address. I played around with different permutations of `connectionAttemptDelay` and `tcpConnectionTimeout`, to trigger various cases.

I also tried blocking certain addresses to check how the implementation behaves in a real-world scenario: `sudo ip6tables -A OUTPUT -d <IPv6 from the verbose logs> -j DROP`

ungive:
The last commit I pushed (29d59cf5c2f469ab3f02ab404fcd9777b6d22e2c) fixes a race condition that was present because I didn't protect the state change `changeState(State::Connected)` in `processConnect()` with the `mSendMutex`. Holding a lock to `mSendMutex` while calling `changeState(State::Connected)` is not feasible because the connected state change callback registered in `WebSocket::setTcpTransport` via the `transport->onStateChange(...)` call would cause a deadlock since e.g. TLS initialization locks the `mSendMutex` again.

Example for the race condition that affected all commits up to and including 88db65008c8fc4a95623df0440dfad142f078f05, with e.g. 4 addresses being attempted:
1. "timeout" poll interrupt on address 3/4
2. "timeout" processing triggers another attempt on address 4/4. this attempt is executed asynchronously on the thread pool, which takes some time to actually happen. meanwhile the next steps happen
3. "out" poll interrupt directly after due to successful connect on address 1/4. "directly after" means the poll was exited in pollservice.cpp and processConnect() is called twice before the poll is started again (which is rare to happen)
4. "out" processing acquires the lock, sets mSock and closes all other unused sockets. it releases the lock to mSendMutex, but does not yet change the state to "connected"
5. the triggered attempt on address 4/4 acquires the lock and adds another pending socket
6. "out" processing now continues to update the state to "connected" and changes the poll service direction for the socket to "in"
8. after a short time, address 4/4 causes "out" poll interrupt due to successfully connecting after some delay
9. the state is not "connecting" anymore, yet mPendingSocks is greater than zero
10. mSock will be overwritten and the previous socket that is being used for communication will be closed

Let me know, if the solution in 29d59cf5c2f469ab3f02ab404fcd9777b6d22e2c is good in your eyes. Holding the lock to `mSendMutex` while calling `changeState(State::Connected)` would require some bigger refactors, since the callback would have to be called after the lock is released again.

src/impl/tcptransport.cpp
NotAFour, line 394: When the transport is torn down while concurrent attempts are still in flight, the in-flight sockets in `mSocks` are never closed, only `mSock` is. `closeUnusedSockets()` is only called on a successful connect and on the "all addresses failed" path in `attempt()`, and the teardown path doesn't compensate: `~TcpTransport()` just calls `close()`, and `Transport::stop()` is only `unregisterIncoming()`. Neither drains `mSocks`. (The poll service doesn't invoke callbacks when stopped, so `processConnect` can't be relied on to clean these up, each owned socket has to be closed by the transport itself, like `mSock` is.)

Adding `closeUnusedSockets()` to `close()` fixes it: no-op after a successful connect, closes the whole set on a mid-connect teardown.

ungive: Thank you! The latest commit fixes this.

src/impl/tcptransport.cpp
```diff
@@ -439,38 +523,117 @@ void TcpTransport::process(PollService::Event event) {
...
+		// 1. We connect synchronously (no connection attempt delay is configured)
+		// 2. We connect concurrently and this socket failed before the delay timeout was reached
+		// 3. We connect concurrently and this was the last pending socket that timed out. This
+		//    means that all sockets that were created so far failed and we must proceed either with
+		//    queueing a connection attempt for the next address or with handling connection
+		//    failure, for the case that there are no more addresses left
+		if (!mConnectAttemptDelay.has_value() || event != PollService::Event::Timeout ||
+		    isLastPendingSock) {
+			ThreadPool::Instance().enqueue(weak_bind(&TcpTransport::attempt, this));
+		}
```
ungive: The code is running in production with my application and is working well so far, though someone discovered a bug in this code today that happens under the following conditions:

- The first address takes longer than the configured `connectAttemptDelay` to connect
- The last address immediately fails or fails quickly enough that the first address does not have enough time to connect

The result is that all ongoing connection attempts are cancelled once the last address fails. This is because `event != PollService::Event::Timeout` evaluates to `true` and calls `attempt()`, which closes all "unused" connections, even though there are still connections that didn't fail.

ungive: This is reproducible with a domain that has the following DNS records:
- `A 104.21.29.32` (ws.ifelse.io)
- `AAAA 2606:4700:3333::ac13:ab44` (ws.ifelse.io but scrambled so it immediately fails)

Then try the websocket.cpp test with `config.connectionAttemptDelay = 1ms;`. The test will fail.

Fixed in the latest commits.

src/impl/tcptransport.cpp
```diff
@@ -158,9 +161,32 @@ void TcpTransport::connect() {
 	ThreadPool::Instance().enqueue(weak_bind(&TcpTransport::resolve, this));
 }
 
+// Reorders the list of addresses to alternate IPv4 and IPv6 addresses towards the beginning of the
+// list whenever possible, while always keeping the first address in its original place.
+static void interleave_address_families(addrinfo *&head) {
```
paullouisageneau: You should move this function to `utils.cpp`.

PR #1599, opened 2026-07-04
Add support for mbedTLS 4.1, use PSA Crypto API

Faless:
mbedTLS released the new LTS version ([4.1.0](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-4.1.0)) in March, and support for the old LTS branch ([3.6](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.6)) is guaranteed only until March 2027.

This PR makes the mbedTLS backend of libdatachannel compatible with mbedTLS 4, while retaining compatibility with mbedTLS 3.6.6 (latest old LTS at the time of writing).

In practice, this moves all the cryptographic functions to the new [PSA Crypto API](https://arm-software.github.io/psa-api/crypto/1.1/) (supported in mbedTLS 3.6, and what mbedTLS 4+ implements).

Plus, it adds a few `ifdef`s for the functions that (annoyingly) changed signature. I can move the clutter inside the `init.hpp`/`init.cpp` as functions in the mbedtls namepsace if you prefer, just let me know (I'm not sure which way is best).

Please note that this only builds with `NO_MEDIA=1`, since libsrtp does not currently support mbedTLS 4 (there is a PR opened for that: https://github.com/cisco/libsrtp/pull/813)

You can test building mbedTLS 4.1.0 + libdatachannel on linux with the following cmake options:
```
# from /path/to/mbedtls
cmake -B build --install-prefix=$(pwd)/build/install -DCMAKE_C_FLAGS='-DMBEDTLS_SSL_DTLS_SRTP' -DCMAKE_POSITION_INDEPENDENT_CODE=1 .
cmake --build build
cmake --install build
```

```
# from /path/to/libdatachannel
PKG_CONFIG_PATH="/path/to/mbedtls/build/install/lib/pkgconfig:$PKG_CONFIG_PATH" \
  cmake -B build -DNO_MEDIA=1 -DUSE_MBEDTLS=1 -DCMAKE_CXX_FLAGS='-DMBEDTLS_SSL_DTLS_SRTP' .
cmake --build build
```

### DIFF
--- a/.github/workflows/build-mbedtls.yml
+++ b/.github/workflows/build-mbedtls.yml
@@ -30,6 +30,30 @@ jobs:
     - name: test
       run: ./build/tests
 
+  build-linux-v4:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+      with:
+        submodules: recursive
+        fetch-depth: 1
+
+    - name: Set up Homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - name: Install Mbed TLS
+      run: brew update && brew install mbedtls@4
+
+    - name: cmake
+      run: cmake -B build -DUSE_MBEDTLS=1 -DNO_MEDIA=1 -DWARNINGS_AS_ERRORS=1  -DCMAKE_PREFIX_PATH=$(brew --prefix mbedtls@4)
+
+    - name: make
+      run: (cd build; make -j2)
+
+    - name: test
+      run: ./build/tests
+
   build-macos:
     runs-on: macos-latest
     steps:
--- a/.github/workflows/build-mbedtls.yml
+++ b/.github/workflows/build-mbedtls.yml
@@ -30,6 +30,30 @@ jobs:
     - name: test
       run: ./build/tests
 
+  build-linux-v4:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+      with:
+        submodules: recursive
+        fetch-depth: 1
+
+    - name: Set up Homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - name: Install Mbed TLS
+      run: brew update && brew install mbedtls@4
+
+    - name: cmake
+      run: cmake -B build -DUSE_MBEDTLS=1 -DWARNINGS_AS_ERRORS=1 -DCMAKE_PREFIX_PATH=$(brew --prefix mbedtls@4)
+
+    - name: make
+      run: (cd build; make -j2)
+
+    - name: test
+      run: ./build/tests
+
   build-macos:
     runs-on: macos-latest
     steps:
--- a/src/impl/certificate.cpp
+++ b/src/impl/certificate.cpp
@@ -203,9 +203,11 @@ string make_fingerprint(mbedtls_x509_crt *crt,
 	}
 
 	size_t hash_size = 0;
-	mbedtls::check(psa_hash_compute(alg, crt->raw.p, crt->raw.len,
-	        buffer.data(), buffer.size(), &hash_size),
-	        "Failed to generate certificate fingerprint");
+	int ret = mbedtls::safe_psa([&] {
+		return psa_hash_compute(alg, crt->raw.p, crt->raw.len,
+		        buffer.data(), buffer.size(), &hash_size);
+	});
+	mbedtls::check(ret, "Failed to generate certificate fingerprint");
 	if (hash_size != buffer.size()) {
 		throw std::runtime_error("Unexpected hash size");
 	}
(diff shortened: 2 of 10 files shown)


src/impl/certificate.cpp
```diff
@@ -180,32 +180,38 @@ string make_fingerprint(mbedtls_x509_crt *crt,
...
 
+	psa_algorithm_t alg = PSA_ALG_NONE;
+	size_t hash_size = 0;
+
 	switch (fingerprintAlgorithm) {
 	case CertificateFingerprint::Algorithm::Sha1:
-		mbedtls::check(mbedtls_sha1(crt->raw.p, crt->raw.len, buffer.data()),
-		               "Failed to generate certificate fingerprint");
+		alg = PSA_ALG_SHA_1;
+		hash_size = 20;
```
paullouisageneau: It seems there is a macro `PSA_HASH_LENGTH(alg)`, so you don't need to set `hash_size` manually.

.github/workflows/build-mbedtls.yml
```diff
@@ -50,3 +74,24 @@ jobs:
 
     - name: test
       run: ./build/tests
+
+  build-macos-v4:
```
paullouisageneau: Is there a reason to add a MacOS build? I think the Linux one would be sufficient.

paullouisageneau (commented):
Thank you for updating to MbedTLS 4. It looks good to me, only I think you should use `PSA_HASH_LENGTH` so you don't have to hardcode lengths.

The ifdefs are fine to me, the OpenSSL code is already full of them. The `mbedtls` namepsace is only for utility functions.

src/impl/certificate.cpp
Faless: Ah, indeed, that was a bit messy, in fact, we already had hard-coded `size` (coming from `CertificateFingerprint::AlgorithmSize(fingerprintAlgorithm)`) set as buffer size.

So I'm now using `buffer.size()` (which is the best way to never go out of bound), and confirm the output size is what we expect (should always be if the function succeeds, but it doesn't hurt checking).

.github/workflows/build-mbedtls.yml
Faless: Ah, no, not specifically, I've removed it and just left the linux one.

Faless:
I'm converting this to draft, since I realized that when `MBEDTLS_THREADING_C` is not enabled (e.g. currently on windows), PSA calls are not thread safe (they rely on a shared state).

I think we need to check that and use a global lock in case.

I'll work on it a bit more tomorrow :pray:

Faless:
Alright, I've finished investigating this. I've added 2 commits to the PR for ease of discussion, I'll squash them when there's a decision.

TL;DR;
--

The first commit is a fix to the current implementation which always require mbedtls threading support to be enabled.

The second commit is a proposed solution to maintain compatibility with mbedTLS 3 in scenarios when threading support is not needed by the TLS functions themselves.

Details
--

In short, PSA Crypto API as implemented by mbedTLS relies on a shared state.

For this reason calling any `psa_*` function, from multiple threads might result race condition on the shared state unless the library is built with `MBEDTLS_THREADING_C`.

In mbedTLS 3, non-PSA functions are safe to call from multiple threads even when `MBEDTLS_THREADING_C` is not enabled unless `MBEDTLS_USE_PSA_CRYPTO` is also set (with the exception of [TLSv1.3](https://github.com/Mbed-TLS/mbedtls/issues/9991) contexts).

In mbedTLS 4+ non-PSA functions might still use the PSA Crypto API internally so calling any function from multiple threads is never safe unless the library is built `MBEDTLS_THREADING_C`.

Currently, there is no upstream threading implementation on windows (on linux/mac it can use pthreads, and the brew build enables it).

A custom thread implementation can be provided by platform implementation (we do so in Godot, and can be done when building a fully static library, but it doesn't make sense to do so in this library).

Making all TLS functions be called from a single thread doesn't seem reasonable with the current library architecture (but I haven't dug into it too much), so for mbedTLS 4 one really needs threading.

At least mbedTLS 3 keep working without threading support like the previous behavior even if with a small hack (not that brew builds from CI have threading enabled).

Let me know what you think @paullouisageneau .

paullouisageneau (approved):
Thank you for the investigation and fixes. I think your solution looks good the way it is. Indeed serializing all TLS functions on a single thread doesn't seem like a workable solution so mbedTLS would need threading support. Could you please merge the two commits and address the conflicts?

Also https://github.com/cisco/libsrtp/pull/813 has recently been merged, so it should be possible to use mbedTLS 4 with media by simply bumping the submodule.

Faless:
Thanks! 

> Could you please merge the two commits and address the conflicts?

I've rebased on master and squashed the commits.

> Also [cisco/libsrtp#813](https://github.com/cisco/libsrtp/pull/813) has recently been merged, so it should be possible to use mbedTLS 4 with media by simply bumping the submodule.

I've also updated the submodule to latest `2_x_dev` and re-enabled media in CI for the mbedtls 4 build (so now it has the same options of the mbedtls 3 one outside the cmake prefix for mbedtls itself).

paullouisageneau:
Great, thanks!

PR #1601, opened 2026-07-05
pass -A to cmake, based on b2's address-model (64 or 32 bits)

arvidn:
The `Jamfile` in this project invokes `cmake` directly, but it only forwards the `debug`/`release` build variant, no other build options.

When passing `address-model=32` (for 32 bit builds) to the `Jamfile`, `cmake` is still invoked making a 64 bit build. This patch looks at the `address-model` build feature and passes it along to `cmake` via `-A`, to build a compatible library.

This patch is demonstrated to work in this PR: https://github.com/arvidn/libtorrent/pull/8532 Specifically, this commit: https://github.com/arvidn/libtorrent/pull/8532/changes/82cd97f710a3ff533f81f906fd910789f4f36c8d

Building 32bit python wheels for libtorrent (with cibuildwheel) failed prior to this patch because of a 32/64 bit incompatibility.

### DIFF
--- a/Jamfile
+++ b/Jamfile
@@ -110,7 +110,14 @@ rule make_libusrsctp_msvc ( targets * : sources * : properties * )
 {
 	local VARIANT = [ feature.get-values <variant> : $(properties) ] ;
 	VARIANT on $(targets) = $(VARIANT) ;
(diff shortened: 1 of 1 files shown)


paullouisageneau (approved):
Thank you for the fix!

PR #1602, opened 2026-07-06
feat: enable SCTP I-DATA interleaving

### LINKED ISSUE #1595: Consider enabling SCTP ndata / I-DATA message interleaving
libdatachannel currently disables SCTP fragment interleaving in `SctpTransport`:

```cpp
int level = 0;
usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level, sizeof(level));
```

The send path also has a TODO for SCTP ndata:

```cpp
// TODO: Implement SCTP ndata specification draft when supported everywhere
```

Vendored usrsctp appears to support I-DATA / ndata via:

- `SCTP_FRAGMENT_INTERLEAVE = 2`
- `SCTP_INTERLEAVING_SUPPORTED = 1`

This could be useful for WebRTC data channels because it can reduce SCTP association-level head-of-line blocking when a large fragmented message is sent while small messages on another stream should remain low-latency.

Questions:

1. Is enabling SCTP ndata still considered unsafe for WebRTC interop today?
2. Would you prefer this as an opt-in `rtc::Configuration` flag first?
3. Are there known Chrome/Firefox/libdatachannel interop issues with usrsctp I-DATA?
4. Would a PR adding guarded opt-in support plus interop tests be acceptable?

A conservative first implementation could keep the default unchanged and only enable it when explicitly configured, then validate:

- libdatachannel <> libdatachannel
- libdatachannel <> Chrome
- ordered reliable large messages plus small unordered unreliable messages on another stream
- `maxRetransmits: 0`
- large messages over 64 KiB

mertushka:
Enable RFC 8260 I-DATA when the linked public usrsctp header exposes `SCTP_INTERLEAVING_SUPPORTED`.

- request fragmentation interleave level 2 and I-DATA, with the existing DATA path as fallback
- assemble partial delivery per stream and clear aborted partial messages
- add a cross-stream regression and a `DISABLE_SCTP_INTERLEAVING` opt-out

Depends on paullouisageneau/usrsctp#1; the general usrsctp change is sctplab/usrsctp#746. This remains draft until the pinned submodule is updated.

Closes #1595.

### DIFF
--- a/src/impl/sctptransport.cpp
+++ b/src/impl/sctptransport.cpp
@@ -59,6 +59,18 @@ using utils::to_uint32;
 static LogCounter COUNTER_UNKNOWN_PPID(plog::warning,
                                        "Number of SCTP packets received with an unknown PPID");
 
+#ifdef SCTP_INTERLEAVING_SUPPORTED
+constexpr int SCTP_INTERLEAVING_SUPPORTED_OPTION = SCTP_INTERLEAVING_SUPPORTED;
+#else
+constexpr int SCTP_INTERLEAVING_SUPPORTED_OPTION = 0x00001206;
+#endif
+
+#ifdef SCTP_FRAG_LEVEL_2
+constexpr int SCTP_FRAGMENT_INTERLEAVE_LEVEL_2 = SCTP_FRAG_LEVEL_2;
+#else
+constexpr int SCTP_FRAGMENT_INTERLEAVE_LEVEL_2 = 2;
+#endif
+
 class SctpTransport::InstancesSet {
 public:
 	void insert(SctpTransport *instance) {
@@ -267,13 +279,45 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
 		throw std::runtime_error("Could not set socket option SCTP_INITMSG, errno=" +
 		                         std::to_string(errno));
 
-	// Prevent fragmented interleave of messages (i.e. level 0), see RFC 6458 section 8.1.20.
-	// Unless the user has set the fragmentation interleave level to 0, notifications
-	// may also be interleaved with partially delivered messages.
-	int level = 0;
-	if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level, sizeof(level)))
-		throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
-		                         std::to_string(errno));
+	const auto disableFragmentedInterleave = [&]() {
+		// Prevent fragmented interleave of messages (i.e. level 0), see RFC 6458 section 8.1.20.
+		// Unless the user has set the fragmentation interleave level to 0, notifications
+		// may also be interleaved with partially delivered messages.
+		int level = 0;
+		if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level,
+		                       sizeof(level)))
+			throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
+			                         std::to_string(errno));
+	};
+
+#if !defined(DISABLE_SCTP_INTERLEAVING)
+	// Enable RFC 8260 I-DATA support when usrsctp exposes it. The association negotiates the
+	// extension and falls back to DATA chunks when the remote peer does not support interleaving.
+	int level = SCTP_FRAGMENT_INTERLEAVE_LEVEL_2;
+	if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level, sizeof(level))) {
+		const int error = errno;
+		PLOG_WARNING << "Could not enable SCTP fragmented interleave, falling back to level 0, "
+		                "errno="
+		             << error;
+		disableFragmentedInterleave();
+	} else {
+		struct sctp_assoc_value interleaving = {};
+		interleaving.assoc_id = SCTP_FUTURE_ASSOC;
+		interleaving.assoc_value = 1;
+		if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_INTERLEAVING_SUPPORTED_OPTION,
+		                       &interleaving, sizeof(interleaving))) {
+			const int error = errno;
+			PLOG_WARNING << "Could not enable SCTP I-DATA interleaving, falling back to level 0, "
+			                "errno="
+			             << error;
+			disableFragmentedInterleave();
+		} else {
+			PLOG_DEBUG << "SCTP I-DATA interleaving enabled for negotiation";
+		}
+	}
+#else
+	disableFragmentedInterleave();
+#endif
 
 #ifdef SCTP_ACCEPT_ZERO_CHECKSUM // not available in usrsctp v0.9.5.0
 	// When using SCTP over DTLS, the data integrity is ensured by DTLS. Therefore, there's no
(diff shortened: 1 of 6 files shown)


mertushka:
**Benchmark:** loopback test, one `bulk` channel streaming 256 KB messages continuously to saturate the SCTP association, one `ping` channel sending an 8-byte timestamped message every 200 ms (echoed back), measuring RTT. 50 pings/run, 3 runs.

| Metric | Baseline | + Interleaving | Improvement |
|---|---|---|---|
| Mean RTT | 583.9 ms | 96.6 ms | 6.0x lower |
| Median RTT | 420.6 ms | 103.2 ms | 4.1x lower |
| p95 RTT | 1560.0 ms | 124.6 ms | 12.5x lower |
| Max RTT | 1760.1 ms | 138.2 ms | 12.7x lower |
| Ping delivery rate | ~24% (10-15/50) | ~48% (24/50) | ~2x |

@paullouisageneau This is amazing! Really hope this lands soon :)

src/impl/sctptransport.cpp
```diff
@@ -59,6 +59,18 @@ using utils::to_uint32;
 static LogCounter COUNTER_UNKNOWN_PPID(plog::warning,
                                        "Number of SCTP packets received with an unknown PPID");
 
+#ifdef SCTP_INTERLEAVING_SUPPORTED
+constexpr int SCTP_INTERLEAVING_SUPPORTED_OPTION = SCTP_INTERLEAVING_SUPPORTED;
+#else
+constexpr int SCTP_INTERLEAVING_SUPPORTED_OPTION = 0x00001206;
```
paullouisageneau: IMHO you shouldn't set SCTP constants like that. `SCTP_INTERLEAVING_SUPPORTED` should be defined by usrsctp, if it is not then interleaving should be disabled at compile time.

src/impl/sctptransport.cpp
```diff
@@ -59,6 +59,18 @@ using utils::to_uint32;
...
+#endif
+
+#ifdef SCTP_FRAG_LEVEL_2
+constexpr int SCTP_FRAGMENT_INTERLEAVE_LEVEL_2 = SCTP_FRAG_LEVEL_2;
+#else
+constexpr int SCTP_FRAGMENT_INTERLEAVE_LEVEL_2 = 2;
+#endif
```
paullouisageneau: The value for level 2 is 2 by definition. No need for a define.

src/impl/sctptransport.cpp
```diff
@@ -267,13 +279,45 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
...
 		                         std::to_string(errno));
 
-	// Prevent fragmented interleave of messages (i.e. level 0), see RFC 6458 section 8.1.20.
-	// Unless the user has set the fragmentation interleave level to 0, notifications
-	// may also be interleaved with partially delivered messages.
-	int level = 0;
-	if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level, sizeof(level)))
-		throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
-		                         std::to_string(errno));
+	const auto disableFragmentedInterleave = [&]() {
```
paullouisageneau: You shouldn't use local lambdas to manage the program flow, please use actual methods or refactor the logic.

src/impl/sctptransport.cpp
```diff
@@ -267,13 +279,45 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
...
+		if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level,
+		                       sizeof(level)))
+			throw std::runtime_error("Could not disable SCTP fragmented interleave, errno=" +
+			                         std::to_string(errno));
+	};
+
+#if !defined(DISABLE_SCTP_INTERLEAVING)
+	// Enable RFC 8260 I-DATA support when usrsctp exposes it. The association negotiates the
+	// extension and falls back to DATA chunks when the remote peer does not support interleaving.
+	int level = SCTP_FRAGMENT_INTERLEAVE_LEVEL_2;
```
paullouisageneau: No need for a define.
```suggestion
	int level = 2;
```

src/impl/sctptransport.cpp
```diff
@@ -267,13 +279,45 @@ SctpTransport::SctpTransport(shared_ptr<Transport> lower, const Configuration &c
...
+	if (usrsctp_setsockopt(mSock, IPPROTO_SCTP, SCTP_FRAGMENT_INTERLEAVE, &level, sizeof(level))) {
+		const int error = errno;
```
paullouisageneau: I don't think it is necessary to introduce `error` here.

src/impl/sctptransport.cpp
paullouisageneau: This comment does not make sense anymore now that interleaving is enabled.

src/impl/sctptransport.cpp
```diff
@@ -517,19 +561,23 @@ void SctpTransport::doRecv() {
 
 			} else {
 				// SCTP message
-				mPartialMessage.insert(mPartialMessage.end(), buffer, buffer + len);
-				if (mPartialMessage.size() > mMaxMessageSize) {
+				if (infotype != SCTP_RECVV_RCVINFO)
+					throw std::runtime_error("Missing SCTP recv info");
+
+				const PartialMessageKey key(info.rcv_sid, info.rcv_ssn,
+				                            (info.rcv_flags & SCTP_UNORDERED) != 0);
```
paullouisageneau: Why does the key contains SSN and unordered flag? [RFC 6458](https://www.rfc-editor.org/rfc/rfc6458.html#section-8.1.20) says the application should only look at the stream id. Am I missing something?

src/impl/sctptransport.cpp
mertushka: Agreed. I removed the private constant entirely. I-DATA now compiles only when the public usrsctp header defines `SCTP_INTERLEAVING_SUPPORTED`; otherwise the transport keeps level-0 behavior. I opened paullouisageneau/usrsctp#1 for the fork pinned by libdatachannel, alongside the general upstream proposal sctplab/usrsctp#746, and converted this PR to draft pending that dependency.

src/impl/sctptransport.cpp
mertushka: Agreed. The private level constant is gone; `configureInterleaving()` now assigns `int level = 2` directly.

src/impl/sctptransport.cpp
mertushka: Refactored. The constructor now calls the `SctpTransport::configureInterleaving()` member instead of using a local flow-control lambda.

src/impl/sctptransport.cpp
mertushka: Applied. The revised method uses `int level = 2` directly when the public usrsctp capability is available.

src/impl/sctptransport.cpp
mertushka: Agreed. The local error copy was removed and the fallback warning logs `errno` directly.

src/impl/sctptransport.cpp
mertushka: Updated. The stale TODO is replaced with a statement that usrsctp emits DATA or I-DATA according to association negotiation.

src/impl/sctptransport.cpp
mertushka: You are correct. Partial messages are now keyed only by `rcv_sid`. I also subscribed to `SCTP_PARTIAL_DELIVERY_EVENT` and erase that stream entry on `SCTP_PARTIAL_DELIVERY_ABORTED`, so an abandoned partial delivery cannot contaminate the next message.

mertushka:
@paullouisageneau Thanks for the review! I’ve addressed the comments. The remaining dependency is [paullouisageneau/usrsctp#1](https://github.com/paullouisageneau/usrsctp/pull/1); once that lands, I’ll update the submodule here and mark this ready.

PR #1604, opened 2026-07-08
perf: reduce transport packet handoff copies

mertushka:
Reduces internal packet handoff overhead on the DataChannel send path by adding a borrowed-buffer transport send path between SCTP, DTLS, and ICE.

It avoids allocating and copying intermediate `Message` objects for synchronous lower-layer sends from usrsctp write callbacks, DTLS write callbacks, and ICE UDP sends, while preserving the existing `message_ptr` fallback for other transports.

### DIFF
--- a/src/impl/dtlstransport.cpp
+++ b/src/impl/dtlstransport.cpp
@@ -134,16 +134,23 @@ void DtlsTransport::stop() {
 }
 
 bool DtlsTransport::send(message_ptr message) {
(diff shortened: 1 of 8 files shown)


mertushka:
Closing this experiment. It does not yet have benchmark or profile evidence demonstrating a material benefit, and the borrowed-send path needs stronger buffer-lifetime and DSCP-state guarantees before it could be justified. No upstream change should proceed without that evidence.

PR #1605, opened 2026-07-08
perf: reduce datachannel send allocation overhead

mertushka:
Reduces DataChannel send allocation overhead by trying immediate SCTP sends from caller-provided buffers before creating a queued Message, while preserving owned-message fallback behavior when SCTP cannot send synchronously.


mertushka:
Closing this experiment. It does not yet have benchmark evidence demonstrating a material benefit, and the direct-send path duplicates SCTP send semantics without sufficient concurrent-ordering, fallback, accounting, and lifecycle coverage. No upstream change should proceed without that evidence.

PR #1610, opened 2026-07-20
Fix return type of RtcpReportBlock::getNTPOfSR

### LINKED ISSUE #1607: Confusing last SR timestamp returning by RtcpReportBlock::getNTPOfSR
I'm developing an application that processes data from RTCP Receiver Report packets.
In the process, I calculate RTT as follows:
```c++
    NTPTimestamp now = getNtpNow();
    std::uint32_t nowRtcp = ntpToRtcpFormat(now);
    if (lsr + dlsr > nowRtcp) {
        return -1;
    }
    std::uint32_t rttUnits = nowRtcp - lsr - dlsr;
    double rttMs = static_cast<double>(rttUnits) / RTT_UNITS_PER_SEC * MS_PER_SEC;
    return rttMs;
```
I was getting strange values, RTT was on the order of tens of millions.
Then I noticed how the last SR timestamp value is set and returned:
```c++
void RtcpReportBlock::setNTPOfSR(uint64_t ntp) { _lastReport = htonl((uint32_t)(ntp >> 16)); }
uint32_t RtcpReportBlock::getNTPOfSR() const { return ntohl(_lastReport) << 16u; }
```
It seems that RtcpReportBlock::getNTPOfSR returns the upper 16 bits of the fraction part of the NTP timestamp.
Perhaps I'm wrong, but it seems to me that there shouldn't be a shift by 16 bits in RtcpReportBlock::getNTPOfSR.
If I'm mistaken, please explain where my error is?

bnchworm:
Closes #1607

### DIFF
--- a/include/rtc/rtp.hpp
+++ b/include/rtc/rtp.hpp
@@ -99,7 +99,7 @@ struct RTC_CPP_EXPORT RtcpReportBlock {
 	[[nodiscard]] uint32_t delaySinceSR() const;
 
 	[[nodiscard]] SSRC getSSRC() const;
-	[[nodiscard]] uint32_t getNTPOfSR() const;
+	[[nodiscard]] uint64_t getNTPOfSR() const;
 	[[nodiscard]] uint8_t getFractionLost() const;
 	[[nodiscard]] unsigned int getPacketsLostCount() const;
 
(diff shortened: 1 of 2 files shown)


bnchworm:
Check [this](https://github.com/paullouisageneau/libdatachannel/issues/1607#issuecomment-5032282415) before accept

paullouisageneau (approved):

PR #1612, opened 2026-07-22
Fix DTLS startup ordering

mertushka:
Serialize DTLS initialization with remote description processing so ICE connectivity cannot start authentication before the remote fingerprint is committed. Require GnuTLS servers to request the peer certificate so fingerprint verification runs in both roles. Add delayed full-answer coverage for valid and invalid fingerprints across DTLS and DTLS-SRTP builds.

### DIFF
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -171,7 +171,14 @@ shared_ptr<IceTransport> PeerConnection::initIceTransport() {
 						    break;
 					    case IceTransport::State::Connected:
 						    changeIceState(IceState::Connected);
-						    initDtlsTransport();
+						    // Serialize with setRemoteDescription so DTLS sees the committed fingerprint.
+						    mProcessor.enqueue([weak_this] {
+							    if (auto locked = weak_this.lock()) {
+								    std::lock_guard signalingLock(locked->signalingMutex);
+								    if (locked->remoteDescription())
+									    locked->initDtlsTransport();
+							    }
+						    });
 						    break;
 					    case IceTransport::State::Completed:
 						    changeIceState(IceState::Completed);
@@ -1186,6 +1193,14 @@ void PeerConnection::processRemoteDescription(Description description) {
 	}
 
 	auto dtlsTransport = std::atomic_load(&mDtlsTransport);
+	if (!dtlsTransport) {
+		// ICE might have connected before the remote fingerprint was committed.
+		auto iceTransport = std::atomic_load(&mIceTransport);
+		if (iceTransport && (iceTransport->state() == Transport::State::Connected ||
+		                     iceTransport->state() == Transport::State::Completed))
+			dtlsTransport = initDtlsTransport();
+	}
+
 	if (description.hasApplication()) {
 		auto sctpTransport = std::atomic_load(&mSctpTransport);
 		if (!sctpTransport && dtlsTransport &&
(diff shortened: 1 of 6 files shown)


src/impl/peerconnection.cpp
```diff
@@ -171,7 +171,14 @@ shared_ptr<IceTransport> PeerConnection::initIceTransport() {
 						    break;
 					    case IceTransport::State::Connected:
 						    changeIceState(IceState::Connected);
-						    initDtlsTransport();
+						    // Serialize with setRemoteDescription so DTLS sees the committed fingerprint.
+						    mProcessor.enqueue([weak_this] {
```
paullouisageneau: You shouldn't use the processor here, I think you should call directly `initDtlsTransport()` if there is a remote description. 
Besides, tasks queued in `mProcessor` should use a shared pointer, otherwise you could end up in a situation where the previous task attempts to delete the object and waits indefinitely on a non-empty processor.

src/impl/peerconnection.cpp
```diff
@@ -171,7 +171,14 @@ shared_ptr<IceTransport> PeerConnection::initIceTransport() {
...
+							    if (auto locked = weak_this.lock()) {
+								    std::lock_guard signalingLock(locked->signalingMutex);
```
paullouisageneau: Why do you lock `signalingMutex` here? It is used to synchronize the signaling state, which is unrelated.

test/dtls.cpp
```diff
@@ -0,0 +1,191 @@
+/**
+ * Copyright (c) 2026
```
paullouisageneau: You should put your name here.

src/impl/peerconnection.cpp
mertushka: Updated. The ICE-connected callback now calls initDtlsTransport() directly when a remote description is present, without queueing on mProcessor. I also serialize DTLS initialization with a narrow mutex because the callback and the post-description catch-up can otherwise race to construct two transports.

src/impl/peerconnection.cpp
mertushka: Removed. DTLS startup no longer takes signalingMutex.

test/dtls.cpp
mertushka: Added the contributor name to the test header.

mertushka:
@paullouisageneau Thanks for the review! This should be ready to merge now.

src/impl/peerconnection.cpp
```diff
@@ -221,6 +222,7 @@ shared_ptr<IceTransport> PeerConnection::initIceTransport() {
 
 shared_ptr<DtlsTransport> PeerConnection::initDtlsTransport() {
 	try {
+		std::lock_guard lock(mDtlsTransportMutex);
```
```suggestion
		std::lock_guard lock(mDtlsTransportInitMutex);
```

src/impl/peerconnection.hpp
```diff
@@ -153,6 +153,7 @@ struct PeerConnection : std::enable_shared_from_this<PeerConnection> {
 
 	shared_ptr<IceTransport> mIceTransport;
 	shared_ptr<DtlsTransport> mDtlsTransport;
+	std::mutex mDtlsTransportMutex;
```
paullouisageneau:
```suggestion
	std::mutex mDtlsTransportInitMutex;
```

paullouisageneau (approved):
It looks good, thank you.

PR #1613, opened 2026-07-22
Preserve codec order in SDP attributes

mertushka:
Fixes SDP codec attributes to follow m-line preference order. Adds regression coverage for parsed, constructed, RTX, and remove/re-add cases.


paullouisageneau:
Is this necessary for interoperability? I might be wrong, but to my knowledge codec attributes are not required to follow the m-line order. If this is just for better readability, adding tests seems really overkill.

mertushka:
Thanks, you're right. The payload type order on the m-line is the normative codec preference; the ordering of the corresponding rtpmap attributes is not required for SDP interoperability. 

Closing this PR.

Issue #1253, opened 2024-09-12
how to handle 2-way video and identify incoming tracks


fusionfoxy:
I am streaming video between a browser and libdatachannel in a scenario where multiple video feeds may exist on both sides and I am having trouble identifying incoming video tracks.

**Scenario:**
 * Libdatachannel side calls `addTrack()` with a video track with SSRC: 42, MSID: "foo"
 * Negotiation happens and browser gets `onTrack` event
 * Browser sees MSID: "foo" and can know which of the feeds available on the libdatachannel side this is.
 * Some time later, the browser calls `addTrack()` with a video track with SSRC: 24, MSID: "bar"
 * Negotiation happens, the browser reuses the `m=` section from the existing track
 * onTrack is not triggered on libdatachannel side, existing `rtc::Track` simply starts receiving RTP data (in addition to the RTCP from its own track). The existing `rtc::Track` does not seem to provide me any insights into the SSRC added by the browser or its msid. This means the libdatachannel side cannot identify that it is the "bar" feed the browser is sending.

**Questions***

`rtc::Track` seems to act like Transceivers do in the browser in that it seems to represent an entire `m=` section in the SDP. The `onMessage` callback receives data from both video tracks. It also kind of acts like a browser `MediaStreamTrack` in that you can use it with an existing mid with `addTrack()` to add an SSRC to an existing `m=` section.

1. Is there no clear distinction between tracks and transceivers in libdatachannel?
2. It seems like `onTrack` should be  triggered when the browser adds its track or am I missing something?
3. If `onTrack` should be triggered, would it be invoked with another `rtc::Track` instance representing the received feed(meaning the contained SSRCs would be those of the browser)?
4. When adding the browser track, how can i get the msid of the browser track on the libdatachannel side?
5. Calling `addSSRC(42, "foo", "bar", "baz")` will produce an offer with ssrc: 42, cname: "foo", msid: "bar baz" as I would expect. This makes me believe cname and msid is handled properly by libdatachannel. Is this only true when sending tracks, not when receiving?

**Side notes**
When browser feeds are created, the browser picks the msid randomly. I am conveying its meaning separately allowing me to assign meaning to the MSID "bar" value from the scenario.

PR #1614, opened 2026-07-22
Expose remote descriptions on tracks

mertushka:
Add C++ and C APIs for reading a Track's applied remote media description and observing later changes on the same m-line.

Notifications run after remote description and signaling state commit, are serialized with `Track::close()`, and are suppressed if the Track closes before queued delivery. Existing `Track::description()` behavior remains unchanged.

Refs #1253.

### DIFF
--- a/src/impl/peerconnection.cpp
+++ b/src/impl/peerconnection.cpp
@@ -1125,7 +1125,10 @@ void PeerConnection::processLocalCandidate(Candidate candidate) {
 	                   &localCandidateCallback, std::move(candidate));
 }
 
-void PeerConnection::processRemoteDescription(Description description) {
+std::vector<PeerConnection::TrackRemoteDescriptionUpdate>
+PeerConnection::processRemoteDescription(Description description) {
+	std::vector<TrackRemoteDescriptionUpdate> trackUpdates;
+
 	// Create tracks from remote description
 	for (int i = 0; i < description.mediaCount(); ++i) {
 		auto media = description.media(i);
@@ -1140,6 +1143,10 @@ void PeerConnection::processRemoteDescription(Description description) {
 						desc.disableRtx();
 						track->setDescription(std::move(desc));
 					}
+
+					auto remoteDescription = *remoteMedia;
+					if (track->setRemoteDescription(remoteDescription))
+						trackUpdates.emplace_back(track, std::move(remoteDescription));
 				}
 				continue;
 			}
@@ -1157,6 +1164,7 @@ void PeerConnection::processRemoteDescription(Description description) {
 
 			// Create incoming track
 			auto track = std::make_shared<Track>(weak_from_this(), std::move(reciprocated));
+			track->setRemoteDescription(*remoteMedia);
 			mTracks.emplace(std::make_pair(track->mid(), track));
 			mTrackLines.emplace_back(track);
 			triggerTrack(track); // The user may modify the track description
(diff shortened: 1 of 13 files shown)


src/impl/track.hpp
```diff
@@ -70,9 +75,12 @@ class Track final : public std::enable_shared_from_this<Track>, public Channel {
 #endif
 
 	Description::Media mMediaDescription;
+	optional<Description::Media> mRemoteDescription;
 	shared_ptr<MediaHandler> mMediaHandler;
 
 	mutable std::shared_mutex mMutex;
+	std::recursive_mutex mRemoteDescriptionCallbackMutex;
```
paullouisageneau, line 82: A `synchronized_callback` doesn't need a mutex, let alone a recursive one. You should remove the mutex.

src/impl/track.cpp
```diff
@@ -65,13 +70,48 @@ void Track::setDescription(Description::Media desc) {
...
 void Track::close() {
 	PLOG_VERBOSE << "Closing Track";
 
 	if (!mIsClosed.exchange(true))
 	{
 		triggerClosed();
 		setMediaHandler(nullptr);
+		{
+			std::lock_guard lock(mRemoteDescriptionCallbackMutex);
+			remoteDescriptionCallback = nullptr;
```
paullouisageneau, line 113: This should be done in `resetCallbacks()`.

src/impl/track.hpp
```diff
@@ -50,7 +51,11 @@ class Track final : public std::enable_shared_from_this<Track>, public Channel {
 	string mid() const;
 	Description::Direction direction() const;
 	Description::Media description() const;
+	optional<Description::Media> remoteDescription() const;
 	void setDescription(Description::Media desc);
+	bool setRemoteDescription(Description::Media desc);
```
paullouisageneau, line 56: `setRemoteDescription()` should be not return a boolean as it makes semantics inconsistent with `setDescription()`. It should return `void` for consistency.

src/impl/peerconnection.cpp
```diff
@@ -1140,6 +1143,10 @@ void PeerConnection::processRemoteDescription(Description description) {
 						desc.disableRtx();
 						track->setDescription(std::move(desc));
 					}
+
+					auto remoteDescription = *remoteMedia;
+					if (track->setRemoteDescription(remoteDescription))
+						trackUpdates.emplace_back(track, std::move(remoteDescription));
```
paullouisageneau, line 1149: You should call `triggerTrackRemoteDescription` synchronously here so you don't need to pass `trackUpdates` around and you can remove the dispatch function. This would also allow the user to configure the track immediately. `triggerTrack` is already called synchronously from here.

src/impl/peerconnection.cpp
```diff
@@ -1157,6 +1164,7 @@ void PeerConnection::processRemoteDescription(Description description) {
 
 			// Create incoming track
 			auto track = std::make_shared<Track>(weak_from_this(), std::move(reciprocated));
+			track->setRemoteDescription(*remoteMedia);
 			mTracks.emplace(track->mid(), track);
 			mTrackLines.emplace_back(track);
 			triggerTrack(track); // The user may modify the track description
```
paullouisageneau, line 1170: I think you should call `triggerTrackRemoteDescription` for new tracks too for the sake of consistency.

PR #1615, opened 2026-07-26
Add typed media stream associations

mertushka:
Add a typed C++ API for RFC 8830 media-level `a=msid` associations.

The getter returns valid remote associations in SDP order. The setter validates and atomically replaces a same-track association set. The change also avoids duplicate associations from repeated SSRCs and prevents an answer from copying the offerer's `msid` values.

No C ABI is added; the existing raw-SDP C path remains available.

Specification: https://www.rfc-editor.org/rfc/rfc8830.html

### DIFF
--- a/src/description.cpp
+++ b/src/description.cpp
@@ -984,6 +984,98 @@ void Description::Entry::ExtMap::setDescription(string_view description) {
 	}
 }
 
+namespace {
+
+bool isValidMsidToken(string_view value) {
+	if (value.empty() || value.size() > 64)
+		return false;
+
+	return std::all_of(value.begin(), value.end(), [](char value) {
+		const auto c = static_cast<unsigned char>(value);
+		return c == 0x21 || (c >= 0x23 && c <= 0x27) || (c >= 0x2A && c <= 0x2B) ||
+		       (c >= 0x2D && c <= 0x2E) || (c >= 0x30 && c <= 0x39) ||
+		       (c >= 0x41 && c <= 0x5A) || (c >= 0x5E && c <= 0x7E);
+	});
+}
+
+optional<Description::Media::MediaStreamAssociation>
+parseMediaStreamAssociation(string_view value) {
+	const size_t separator = value.find(' ');
+	const string_view streamId = value.substr(0, separator);
+	if (!isValidMsidToken(streamId))
+		return nullopt;
+
+	optional<string> trackId;
+	if (separator != string_view::npos) {
+		const string_view appData = value.substr(separator + 1);
+		if (!isValidMsidToken(appData))
+			return nullopt;
+
+		trackId = string(appData);
+	}
+
+	return Description::Media::MediaStreamAssociation{string(streamId), std::move(trackId)};
+}
+
+} // namespace
+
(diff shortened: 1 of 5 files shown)


src/description.cpp
```diff
@@ -992,6 +984,98 @@ void Description::Entry::ExtMap::setDescription(string_view description) {
...
+bool isValidMsidToken(string_view value) {
+	if (value.empty() || value.size() > 64)
+		return false;
+
+	return std::all_of(value.begin(), value.end(), [](char value) {
+		const auto c = static_cast<unsigned char>(value);
+		return c == 0x21 || (c >= 0x23 && c <= 0x27) || (c >= 0x2A && c <= 0x2B) ||
+		       (c >= 0x2D && c <= 0x2E) || (c >= 0x30 && c <= 0x39) ||
+		       (c >= 0x41 && c <= 0x5A) || (c >= 0x5E && c <= 0x7E);
+	});
```
paullouisageneau, line 998: It looks like this checks that the string is made of `token-char` in the sense of RFC 4566. Note this formulation has been deprecated by RFC 8866. It seems overkill, and also a bit inconsistent. Why would only MSIDs be strictly checked like that as it applies to all SDP tokens? IMHO I think you should check the length only.

src/description.cpp
```diff
@@ -992,6 +984,98 @@ void Description::Entry::ExtMap::setDescription(string_view description) {
...
+void Description::Media::setMediaStreamAssociations(
+    std::vector<MediaStreamAssociation> associations) {
+	std::vector<MediaStreamAssociation> normalized;
+	for (auto &association : associations) {
+		if (!isValidMsidToken(association.streamId) ||
+		    (association.trackId && !isValidMsidToken(*association.trackId)))
+			throw std::invalid_argument("Invalid media stream association");
+
+		if (!normalized.empty() && association.trackId != normalized.front().trackId)
+			throw std::invalid_argument("Media stream associations have different track IDs");
```
paullouisageneau, line 1057: If you need to do this, the API is probably ill-suited, as the method should take a single track id.

src/description.cpp
```diff
@@ -1004,7 +1088,7 @@ void Description::Media::addSSRC(uint32_t ssrc, optional<string> name, optional<
 	if (msid) {
 		mAttributes.emplace_back("ssrc:" + std::to_string(ssrc) + " msid:" + *msid + " " +
 		                         trackId.value_or(*msid));
-		mAttributes.emplace_back("msid:" + *msid + " " + trackId.value_or(*msid));
+		addAttribute("msid:" + *msid + " " + trackId.value_or(*msid));
```
paullouisageneau, line 1091: If the user calls `setMediaStreamAssociations()` then `addSSRC()`, no sanity check is done about the MSID or track id. I think it should at least call the media stream association method for consistency. Why not adding a `addMediaStreamAssociation(string msid, string trackId)` method instead? You could also add `removeMediaStreamAssociation(string msid)` and `hasMediaStreamAssociation(string msid, string trackId)` to make it consistent with the other methods of `Description`.

paullouisageneau (commented):
I'm not sure about the API for this. Other

PR #1624, opened 2026-08-06
Implement sending RTCP BYE when a media track is closed

aronrosenberg:
Sends an RTCP BYE when a media track is closed, so remote peers learn a stream has ended rather than having to time it out.

* Create new RtcpBye packet type
* Add a `MediaHandler::close()` hook with a `closeChain()` walk, so handlers are told the media is closing and can send final packets while the transports are still up
* Send the BYE from `RtcpSrReporter` (its sending SSRC plus the RTX SSRC) and from `RtcpReceivingSession` (receive-only media)
* Detach the transports after the queued close tasks so the BYE is not raced by transport teardown, without delaying the Closed state or the callback reset
* Extract SSRCs from inbound BYE when demultiplexing compound RTCP
* Deliver compound RTCP packets to each track only once, not once per SSRC per track in a payload
* Add unit and integration tests for the packet, both handlers, every media direction, multiple tracks, RTX, and the close ordering

AI Assisted


include/rtc/configuration.hpp
```diff
@@ -78,6 +78,7 @@ struct RTC_CPP_EXPORT Configuration {
 	bool disableAutoGathering = false;
 	bool forceMediaTransport = false;
 	bool disableFingerprintVerification = false;
+	bool sendRtcpByeOnTrackClose = false;
```
paullouisageneau, line 81: I think there should not be an opt-in per-connection config flag for such a behavior. It should be enabled by default, with optionally a compile flag to disable if you think regressions are possible. If you really want to make it optional at runtime it should be implemented in a media handler.

src/impl/track.cpp
```diff
@@ -128,6 +133,72 @@ void Track::open(shared_ptr<DtlsSrtpTransport> transport) {
...
+	std::vector<SSRC> ssrcs;
+	std::vector<string> cnames;
+	{
+		std::shared_lock lock(mMutex);
+
+		// Only audio and video tracks carry RTP/RTCP traffic — application (data channel) and
+		// other non-media m= sections never have an SSRC stream and must not emit BYE.
+		auto mediaType = mMediaDescription.type();
+		if (mediaType != "audio" && mediaType != "video")
+			return;
```
paullouisageneau, line 153: Tracks are guaranteed to be media, there is no need for this redundant test, and it's misleading because it hints that media type could be application or something else here. You should remove it.

src/impl/track.cpp
```diff
@@ -128,6 +133,72 @@ void Track::open(shared_ptr<DtlsSrtpTransport> transport) {
...
+		ssrcs = mMediaDescription.getSSRCs();
+		cnames.reserve(ssrcs.size());
+		for (auto ssrc : ssrcs)
+			cnames.push_back(mMediaDescription.getCNameForSsrc(ssrc).value());
+	}
+
+	// RFC 3550 section 6.3.7: a participant which never sent RTP/RTCP MUST NOT send BYE.
+	// Only send for tracks we were transmitting on.
+	if (dir != Description::Direction::SendOnly && dir != Description::Direction::SendRecv)
+		return;
```
paullouisageneau: This check doesn't do what the comment says. First, the direction doesn't say if an RTP/RTCP packet was actually sent on the track. Then, RTCP packets are also sent on `RecvOnly` tracks.

src/impl/track.cpp
```diff
@@ -128,6 +133,72 @@ void Track::open(shared_ptr<DtlsSrtpTransport> transport) {
...
+		auto bye = reinterpret_cast<RtcpBye *>(msg->data() + rrSize + sdesSize);
+		bye->preparePacket(1);
+		bye->setSsrc(0, ssrc);
+
+		try {
+			transport->sendMedia(msg);
+		} catch (const std::exception &e) {
+			PLOG_WARNING << "Failed to send RTCP BYE for SSRC " << ssrc << ": " << e.what();
+		}
+	}
```
paullouisageneau, line 200: This should definitely be done in `RtcpReceivingSession` instead of generating a dummy RR. It would also make it easier to arm the sending of RTCP BYE as soon as an RTP/RTCP packet has been sent out.

include/rtc/configuration.hpp
aronrosenberg: I redid the design and removed the opt-in config option and made it always be sent. I left out a compile time removal option.

src/impl/track.cpp
aronrosenberg: Removed and used a new integration model inside a MediaHandler. I did this by creating a new close() virtual that a handler can implement and is called sync when the track is torn down before the transports might get closed.

src/impl/track.cpp
aronrosenberg: Fixed and updated the logic in the new handler implementation.
The Track class now keeps a mSentPacket which is set by the transport send if any outbound packet has been sent. This is used in the new close() model, along with the direction in the handler so it can make an intelligent choice about if the send side SR handler needs to send the BYE under the send side SSRC, or in a recv-only case if the recv side SSRC is used to send the BYE.

src/impl/track.cpp
aronrosenberg: This has all been moved into both RtcpReceivingSession and RtcpSrReporter and goes into the appropriate SR or RR report depending on the Direction of the track.

src/impl/peerconnection.cpp
```diff
@@ -97,11 +97,12 @@ void PeerConnection::close() {
...
 	if (state.load() != State::Closed) {
-		// Close data channels and tracks asynchronously
+		// Close data channels, tracks and transports asynchronously and in order, so the track
+		// media handlers get a chance to send their final packets, e.g. RTCP BYE, while the
+		// transports are still up
 		mProcessor.enqueue(&PeerConnection::closeDataChannels, shared_from_this());
 		mProcessor.enqueue(&PeerConnection::closeTracks, shared_from_this());
-
-		closeTransports();
+		mProcessor.enqueue(&PeerConnection::closeTransports, shared_from_this());
```
aronrosenberg: @paullouisageneau I want to call out this change very closely. I had to switch the closeTransports() call to be async since otherwise the closeTracks would often lose the race to being able to put packets on the wire from the close() handler as the transport itself would be closed by the time the async invoke occurred. I added unit tests which proved this behavior, but it is a change.

Note that despite the function name being remoteClose(), a regular PeerConnection.close() funnels through here when there isn't a mSctpTransport active.

src/impl/peerconnection.cpp
```diff
@@ -658,10 +665,15 @@ void PeerConnection::dispatchMedia([[maybe_unused]] message_ptr message) {
...
+			// A compound packet may name several SSRCs belonging to the same media: report blocks,
+			// SDES chunks, REMB and FIR source lists, or a BYE listing a primary and its RTX SSRC.
+			// Deliver it once per track, otherwise a handler sees the same packet several times.
+			std::set<Track *> delivered;
 			for (uint32_t ssrc : ssrcs) {
 				if (auto it = mTracksBySsrc.find(ssrc); it != mTracksBySsrc.end()) {
 					if (auto track = it->second.lock())
-						track->incoming(message);
+						if (delivered.insert(track.get()).second)
+							track->incoming(message);
```
aronrosenberg, line 681: @paullouisageneau I want to call out this change very closely. This was another pre-existing bug I found and fixed. If a compound RTCP packet had multiple SSRCs belonging to the same track, it was previously being delivered N times, once per unique SSRC. The new BYE's when RTX is active were triggering this because the SSRC for the main media and the SSRC for RTX were in the same BYE payload.

It was also possible if multiple SSRCs were used for main media that this would spawn duplicate RR's being sent as when a SR with multiple SSRC's was received, it would deliver the SR twice to the handler which would then cause two RR's to be sent skewing the transmission states. I don't think would actually happen since the library doesn't implement simulcast support using multiple SSRC's, only RID based.

src/impl/peerconnection.cpp
```diff
@@ -387,6 +387,12 @@ void PeerConnection::closeTransports() {
 	setMediaHandler(nullptr);
 	resetCallbacks();
 
+	// Detach the transports after the tasks queued by remoteClose(), so a track being closed on
+	// the processor can still send its final packets, e.g. RTCP BYE, while they are up
+	mProcessor.enqueue(&PeerConnection::detachTransports, shared_from_this());
+}
+
+void PeerConnection::detachTransports() {
```
aronrosenberg, line 395: @paullouisageneau I want to call out this change very closely. I had to switch the closeTransports() call to be a mix of sync and async since otherwise the closeTracks would often lose the race to being able to put packets on the wire from the close() handler as the transport itself would be closed by the time the async invoke occurred if invoked from the destructor. I added unit tests which proved this behavior, but it is a structural change.

If you have other ideas on a better to solve this, let me know and I can change it.

Issue #1538, opened 2026-04-06
Deadlock: DTLS incoming queue full while poll thread holds conn_lock


gasperov:
Co-Authored-By: Claude

# Deadlock: DTLS incoming queue full while poll thread holds conn_lock

## TL;DR

Under high bidirectional TURN relay traffic, the libjuice poll thread blocks inside
`DtlsTransport::incoming` waiting to push to a full queue, while the ThreadPool thread
that would drain that queue is waiting to acquire `registry->mutex` — which the poll
thread is already holding. Neither thread can proceed. **Only TURN relay connections
are affected.** The correct fix is to unlock `registry->mutex` in libjuice before
invoking `cb_recv`.

## Summary

Under sustained high-throughput bidirectional traffic, libdatachannel can deadlock permanently.
The main thread pool stops making progress and the connection hangs indefinitely.

### Call stacks at freeze

**Poll thread (blocked in push):**
```
Queue<Message>::push          src/impl/queue.hpp:92
DtlsTransport::incoming       src/impl/dtlstransport.cpp:504
IceTransport::RecvCallback    src/impl/icetransport.cpp
agent_input                   deps/libjuice/src/agent.c:885
agent_conn_recv               deps/libjuice/src/agent.c:821
conn_poll_process             deps/libjuice/src/conn_poll.c
  (holds registry->mutex since line 495)
```

**ThreadPool thread (blocked waiting for registry->mutex):**
```
mutex_lock                    (waiting for registry->mutex)
conn_poll_lock                deps/libjuice/src/conn_poll.c:635
conn_lock → juice_send_diffserv
IceTransport::outgoing        src/impl/icetransport.cpp
DtlsTransport::WriteCallback  src/impl/dtlstransport.cpp
  (triggered by SCTP retransmit or application send)
```

### Option A — Command-queue registry

The root problem is that `registry->mutex` is held for the entire poll iteration, including
recv callbacks. The fix is to restrict the mutex to protecting only the agents list membership,
and make the **poll thread the sole writer** of `agents[]`/`agents_size` and other registry members.

**Design:**

- Add a `pending_commands` queue to the registry (guarded by `registry->mutex`).
- API callers (`juice_create`, `juice_destroy`) append add/remove commands to the queue and return immediately. The mutex is held only for the queue append — never during polling.
[...]

PR #1630, opened 2026-08-26
Fix DTLS handshake deadlock when start() runs on the juice state callback thread (TURN relay)

ggodool-nx:
_I received help from claude-fable-5 with debugging and writing pull requests._
## Summary

`DtlsTransport::start()` calls `handleTimeout()` directly at the end. When `start()` is invoked from the libjuice state-change callback (via `IceTransport::processStateChange`), that callback runs while libjuice holds the agent/registry lock. If another thread already holds `mSslMutex` and is blocked inside `juice_send()` waiting for that same lock, the two threads deadlock and the DTLS handshake never completes.

This is the same lock cycle as #1538, but with a different trigger: there it was the incoming queue being full (fixed by #1546); here it is the direct `handleTimeout()` call at the end of `start()`.

## Deadlock

Thread A — juice poll thread (holds agent/registry lock):

```
conn_poll (conn_poll.c, holds registry->mutex)
agent_change_state -> state_changed callback (agent.c, called with lock held)
IceTransport::processStateChange
DtlsTransport::start()
DtlsTransport::handleTimeout()        <- blocks on mSslMutex
```

Thread B — thread pool (holds mSslMutex):

```
DtlsTransport::doRecv() (locks mSslMutex, processing queued ClientHello)
SSL_do_handshake -> BIO write -> IceTransport::outgoing
juice_send -> agent_send -> conn_lock(agent)   <- blocks on registry->mutex
```

`agent_send` only takes the agent lock on the relayed path (`agent_channel_send`); direct sends are lock-free. So the deadlock only occurs when the selected pair is a TURN relay.

## Reproduction conditions

- DTLS server (passive) role
- A ClientHello is already queued before/while `start()` runs (low-RTT relay makes this common; the remote sends its first flight as soon as its own pair is selected)
- Selected candidate pair is TURN relay

With an ice-transport-policy of relay-only and a low-latency TURN server, this reproduced on nearly every connection attempt for us (v0.24.5). Log instrumentation confirmed both threads blocked exactly as above: `handleTimeout()` entered but never acquired `mSslMutex`, and `juice_send()` never returned.

## Fix

Do not call `handleTimeout()` synchronously from `DtlsTransport::start()`; defer it to the thread pool so the juice callback thread never waits on `mSslMutex`. This uses `enqueueRecv()`, the same call the GnuTLS and Mbed TLS backends already make at the end of their `start()` implementations. The scheduled `doRecv()` processes any pre-queued messages first (calling `handleTimeout()` as needed) and routes the potential "Handshake timeout" exception through its existing try/catch:

```cpp
-	handleTimeout();
+	// start() can run on the libjuice poll thread via the ICE state-change
+	// callback, while juice still holds its registry mutex. Calling
+	// handleTimeout() inline then waits on mSslMutex, which a thread pool
+	// recv thread may hold while blocked in juice_send (conn_lock) on the
+	// TURN relay path, forming an ABBA deadlock. Defer the initial timeout
+	// handling to the thread pool, as the other TLS backends do.
+	enqueueRecv(); // to initiate the handshake
```

Only the OpenSSL backend is affected: the GnuTLS and Mbed TLS `start()` implementations already end with `enqueueRecv(); // to initiate the handshake` instead of a synchronous timeout call, so no equivalent change is needed there — this change also makes the OpenSSL backend consistent with the other two.


paullouisageneau:
Thank you for the fix, but there is really no need to add an example for this. Also please read [CONTRIBUTING.md](https://github.com/paullouisageneau/libdatachannel/blob/master/CONTRIBUTING.md) about AI tooling.

Please remove the example, otherwise the fix looks good.

ggodool-nx:
Thanks for the review! I removed the example.

Issue #1632, opened 2026-09-07
Intermittent DTLS handshake failure on loopback: SSL_do_handshake returns SSL_ERROR_SYSCALL with errno=0 and an empty error queue, before any BIO write (~1.75%)


johnhenry:
About 1.75% of loopback peer-to-peer connections fail during DTLS startup. ICE succeeds on both sides — the nominated candidate pair reports `state: succeeded` — and then `DtlsTransport::start()` fails immediately with `Handshake failed: fatal I/O error`, no data having been sent.

Reproduced with **libdatachannel alone**, no bindings, two `PeerConnection`s in one process on loopback with no ICE servers.

### Environment

| | |
|---|---|
| libdatachannel | v0.24.3 (`c6696d1`), submodules as pinned |
| TLS backend | OpenSSL 3 (`/opt/homebrew/opt/openssl@3`) |
| Build | Release, `-DNO_MEDIA=ON -DNO_WEBSOCKET=ON` |
| Platform | macOS 26.5.2, arm64 (Apple Silicon) |

### Result

```
RESULT 7/400 failed
```

with a perfect 1:1 correspondence — **7 failures, 7 handshake errors, 0 BIO write callbacks**.

### The failure

`src/impl/dtlstransport.cpp:840`, `DtlsTransport::start()`:

```cpp
ret = SSL_do_handshake(mSsl);
err = SSL_get_error(mSsl, ret);
...
openssl::check_error(err, "Handshake failed");   // SSL_ERROR_SYSCALL -> throw
```

Instrumented, every failure reports exactly:

```
ret=-1  err=5 (SSL_ERROR_SYSCALL)  errno=0  ERR_peek_last_error()=0
```

so `src/impl/tls.cpp:198` maps it to `"Handshake failed: fatal I/O error"`, `PeerConnection` goes to `Failed`, and the DTLS transport is destroyed. Incoming DTLS records that arrive milliseconds later are then dropped with `Transport::recv@79: dropping incoming message, no receive callback`.

Two things that seem worth noting:

- **`errno == 0` and an empty OpenSSL error queue.** Nothing actually failed at the syscall level.
- **`BioMethodWrite` is never entered.** I instrumented it for all three of its outcomes (null transport, `outgoing()` returning false, `outgoing()` throwing) and none fires — so the handshake aborts before producing any output, and the write path is not implicated.

[...]

PR #1633, opened 2026-09-09
Report errno and SSL_want for SSL_ERROR_SYSCALL

johnhenry:
Diagnostics only — no control flow changes. Every branch throws or returns exactly as before.

## Why

`SSL_get_error()` returns `SSL_ERROR_SYSCALL` for a real I/O failure **and** as its fall-through: when the SSL object wants I/O but the corresponding BIO's retry flags do not agree, it drops past the `SSL_want_read` / `SSL_want_write` branches and lands there with an empty error queue.

Both transports feed OpenSSL from memory BIOs, so the fall-through is a likely cause, and `"fatal I/O error"` sends the reader looking for an I/O failure that may not have happened. I lost several days to that message in #1632 before working out it wasn't reporting I/O at all.

## What

`check_error()` gains an optional `const SSL *` (defaulting to `nullptr`) and, when it has one, appends `SSL_want()` to the `SSL_ERROR_SYSCALL` message. The four handshake call sites pass it. Every other caller is untouched.

`errno` is deliberately **not** reported: the syscalls behind an I/O failure can happen far down the stack, possibly in another library, and it may not be set at all on Windows. `SSL_want()` is the SSL object's own state, so it is meaningful regardless of where any I/O happened.

## Measured

On a loopback reproducer against **v0.24.3**, which predates #1584 and so still has the BIO race, 800 trials, 12 failures:

```
before   Handshake failed: fatal I/O error                  ×12

after    Handshake failed: fatal I/O error (want=reading)   ×11
         Handshake failed: fatal I/O error (want=nothing)   ×1
```

The 11 are #1584's race — the SSL wants to read, but a concurrent `BIO_write` into the memory BIO cleared the retry flag `SSL_get_error()` was about to read. `want=reading` says that directly.

The remaining one is **something else**, and that is the argument for this patch more than the other eleven: one message was covering two distinct failures with no way to tell them apart in a log.

(These lines come from the original measurement run, which also printed `errno`; the `want=` values above are exactly what that run reported, with the `errno` field removed from the message.)

## Notes

- `SSL_RETRY_VERIFY` is `#ifdef`-guarded — it is OpenSSL 3.0+ and this still builds against 1.1.1.
- GnuTLS and MbedTLS builds are untouched; this is inside the `#else // OPENSSL` section of `tls.cpp`.
- Built and run clean against current master (OpenSSL 3.6.3, `NO_MEDIA=ON NO_WEBSOCKET=ON`), 0/100 trials failing, as expected since #1584 landed.


paullouisageneau:
You should not rely on errno in this library as actual syscalls for I/O operations may happen far away down the stack, possibly in other libraries, so there is absolutely no guarantee errno is properly set. Also, it may not be set on Windows.

johnhenry:
Thanks — agreed. I've dropped `errno` entirely (new commit) and kept only `SSL_want()`, which is the SSL object's own state and doesn't depend on where, or whether, a syscall happened. The message is now e.g. `Handshake failed: fatal I/O error (want=reading)`, and unchanged (`fatal I/O error`) at call sites that don't pass the `SSL *`.

It still separates the two cases from the original measurement: the eleven BIO-retry races report `want=reading`, and the other failure reports `want=nothing`. I've updated the description to match.

paullouisageneau (approved):

PR #1634, opened 2026-09-10
Reset mInitialized when doInit fails

clauadv:
### Problem

`Init::doInit` sets `mInitialized = true` before running the subsystems and never rolls it back if one throws. `PollService::start()` throws whenever `PollInterrupter`'s ctor can't set up its localhost UDP (cold Winsock, security software hooking socket creation). After that, `mInitialized` stays true, `mInterrupter` stays null, and nothing on `Init`'s public surface resets the flag.

The next `rtc::WebSocket` skips init because `mInitialized == true`, and `PollService::add` calls `mInterrupter->interrupt()` behind only an `assert`. In release the assert is compiled out, so it derefs a null `unique_ptr` and segfaults in `PollInterrupter::interrupt()`.

### Reproduction

Force the throw:

```diff
--- a/src/impl/pollinterrupter.cpp
+++ b/src/impl/pollinterrupter.cpp
@@
 PollInterrupter::PollInterrupter() {
+	throw std::runtime_error("failure");
 #ifdef _WIN32
```

Then open two WebSockets:

```cpp
for (int32_t idx = 1; idx <= 2; idx++) {
    try {
        auto ws = std::make_shared<rtc::WebSocket>(rtc::WebSocketConfiguration{});
        ws->open("ws://127.0.0.1:8787/");
    } catch (const std::exception &e) {
        printf("attempt %d threw: %s\n", idx, e.what());
    }
}
```

Without the fix the second attempt crashes instead of throwing cleanly

### Fix

`doInit` now catches failures with a function-try-block, reclaims the spawned thread pool workers, and resets `mInitialized = false` before rethrowing, so a failed initialization is recoverable: the next attempt re-runs it instead of skipping over a half-dead state.


paullouisageneau (approved):
It looks good, thanks for the fix!

